PR #670, opened 2014-09-25
Tests: Allowed nested suites/modules

### LINKED ISSUE #543: Allow nested suites (modules)?

JamesMGreene:
Fixes #543
Closes #670

Debatably a work-in-progress... but... it works pretty well so far! :+1:

For the sake of comparison (and so I didn't go crazy), I chose to implement this functionality in a separate API method, `QUnit.suite`, rather than trying to make `QUnit.module` work in 2 very different ways.  They can conceivably be combined, though.

Another note of interest in how I chose to make suites work differently than modules: every test inside of a suite gets its own atomic context object that can be modified along the way by every level's `beforeEach`/`afterEach` calls. Conversely, using modules, every test within that module shares a shallow copy of the same context object.

### DIFF
--- a/test/suitesOrder.js
+++ b/test/suitesOrder.js
@@ -1,0 +1,157 @@
+var execHistoryCopy,
+	execHistory = [];
+
+(function( suite, test, beforeEach, afterEach ) {
+
+	beforeEach(function( assert ) {
+		execHistory.push( assert.suite.fullName + ".beforeEach1" );
+	});
+	beforeEach(function( assert ) {
+		execHistory.push( assert.suite.fullName + ".beforeEach2" );
+	});
+
+	test( "test1", function( assert ) {
+		assert.expect( 0 );
+		execHistory.push( assert.suite.fullName + " -> test1.callback" );
+	});
+
+	suite( "level1", function() {
+		beforeEach(function( assert ) {
+			execHistory.push( assert.suite.fullName + ".beforeEach1" );
+		});
+		beforeEach(function( assert ) {
+			execHistory.push( assert.suite.fullName + ".beforeEach2" );
+		});
+
+		test( "test2", function( assert ) {
+			assert.expect( 0 );
+			execHistory.push( assert.suite.fullName + " -> test2.callback" );
+		});
+
+		suite( "level2", function() {
+			beforeEach(function( assert ) {
+				execHistory.push( assert.suite.fullName + ".beforeEach1" );
+			});
+			beforeEach(function( assert ) {
+				execHistory.push( assert.suite.fullName + ".beforeEach2" );
+			});
+
+			test( "test3", function( assert ) {
+				assert.expect( 0 );
+				execHistory.push( assert.suite.fullName + " -> test3.callback" );
+			});
+			test( "test4", function( assert ) {
+				assert.expect( 0 );
+				execHistory.push( assert.suite.fullName + " -> test4.callback" );
+			});
+
+			afterEach(function( assert ) {
+				execHistory.push( assert.suite.fullName + ".afterEach1" );
+			});
+			afterEach(function( assert ) {
+				execHistory.push( assert.suite.fullName + ".afterEach2" );
+			});
+		});
+
+		test( "test5", function( assert ) {
+			assert.expect( 0 );
+			execHistory.push( assert.suite.fullName + " -> test5.callback" );
+		});
+
+		afterEach(function( assert ) {
+			execHistory.push( assert.suite.fullName + ".afterEach1" );
+		});
+		afterEach(function( assert ) {
+			execHistory.push( assert.suite.fullName + ".afterEach2" );
+		});
+	});
+
+	test( "test6", function( assert ) {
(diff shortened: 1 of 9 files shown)


baloneysandwiches:
+1 for infinite-depth nesting. Excited to use this. Thanks guys.

JamesMGreene:
@gibson042 

> Specifically, the module picker, but generally, the fact that only the outermost groups are known before execution starts. And in that sense, I'm even more inclined to point out that what's being offered is not so much "nested suites" as "nested tests".

Actually, that's not true here. The `callback` param passed to the `QUnit.suite` method is invoked immediately (and synchronously), so we could build a whole hierarchy of "modules"/"suites" to put in the module/suite filter dropdown before the test run even starts.

JamesMGreene:
> I would like to have this one landed for the next version. Even if it's still a 1.x, we could have our first impressions on using the nested suites.

I agree with @leobalter on this, plus adding this into a late `v1.x` release gives us an easy SemVer "out" to change/remove the API if needed in `v2.x`. Should probably mark this API as "experimental" or "unstable" in the `v1.x` docs, though.

gibson042:
> The `callback` param passed to the `QUnit.suite` method is invoked immediately (and synchronously), so we could build a whole hierarchy of "modules"/"suites" to put in the module/suite filter dropdown before the test run even starts.

Oh, of _course_! It's not that code passed to `test` is _necessarily_ limited to assertions, but it **is** necessarily asynchronously invoked and optional. And coming from the other side, code passed to `module`/`suite` need not _necessarily_ lack assertions, but any assertions would necessarily execute synchronously and precede any `setup`/`beforeEach` code. So they **are** conceptually distinct, and with that in mind I withdraw my support for unification at this time—but would still like to preserve it as a future option, which will be important when evaluating invocation context and arguments.

So on that note...

> every test inside of a suite gets its own atomic context object that can be modified along the way by every level's `beforeEach`/`afterEach` calls. Conversely, using modules, every test within that module shares a shallow copy of the same context object.

That's the `this` object for `test`, `beforeEach` and `afterEach` code, right? I like the sound of this, but can you elaborate on the differences?

> This also provides basically the same non-verbose API as all of the others, when coupled with the simple IIFE I typically use to make a few important QUnit top-level properties (in this case: `QUnit.suite`, `QUnit.test`, `QUnit.beforeEach`, `QUnit.afterEach`) look globally accessible within it.

I think it's out of scope for this PR, but I'd really favor explicitly contextual versions of those, for reasons similar to those that motivated explicitly contextual assertions. There's a hint of it in the last paragraph of https://github.com/jquery/qunit/pull/670#issuecomment-57106566 .

And a bump on this topic, which I think was lost in the noise: https://github.com/jquery/qunit/pull/670#issuecomment-57346894
> > I prefer to arrive at the same end without the explicit/verbose nature of having to pass/expect/chain more arguments/parameters than truly necessary (which is really just `assert`, IMHO).

> Can you elaborate? I'd like `assert` to bundle only assertions, and in fact the `assert.suite` introduced here seems... misplaced.

My (very late) epiphany has definitely warmed me to this approach, though. Thanks for the clarification, @JamesMGreene.

test/suitesOrder.js
```diff
@@ -0,0 +1,157 @@
...
+		beforeEach(function( assert ) {
+			execHistory.push( assert.suite.fullName + ".beforeEach2" );
+		});
+
+		test( "test2", function( assert ) {
+			assert.expect( 0 );
+			execHistory.push( assert.suite.fullName + " -> test2.callback" );
+		});
+
+		suite( "level2", function() {
```
gibson042: Tying invocation of global `beforeEach`/`afterEach`/`test`/`suite` functions to suite context by virtue of being executed during a suite callback feels dirty (and _exactly_ like tying invocation of global assertion functions to test context by virtue of being executed during a test callback, which is deservedly on the way out). I know there are backcompat reasons for that behavior with `module`, but I think we can do better with this `suite` green field.

I'd like them to instead be accessible from either context or arguments, which directly relates to (and partly refutes) the end of https://github.com/jquery/qunit/pull/670#issuecomment-78513676 . _How_ to do so may be a matter of great bikeshedding, though. Some of my own goals on that point are as follows, in no particular order:
* Intuitively-readable invocations, possibly paradigm-dependent (e.g., use of BDD names)
* Maximum similarity between `suite` and `test` callbacks, for future iterative convergence
* Separate concerns (e.g., no properties on `assert` that do not pertain to assertions, since we might want to allow integrating external assertion libraries)

test/suitesOrder.js
```diff
@@ -0,0 +1,157 @@
+var execHistoryCopy,
+	execHistory = [];
+
+(function( suite, test, beforeEach, afterEach ) {
```
Krinkle: Can we use these directly instead of aliasing? Our API should not warrant this. If we don't even want to use it ourselves, we shouldn't ship it. I haven't found it to be an issue though. But I'm happy to revisit it. In our own tests should serve as good example though. I'd rather not encourage people copy this.

Issue #521, opened 2014-02-01
Missing CommonJS exports


jdalton:
QUnit 1.12.0-1.14.0 fail to work in CommonJS environments because of its faulty UMD.

It checks for `modules.exports` but some CommonJS environments may not have it, only having `exports`, so then it extends what it thinks is the global object but its just the module scope, so QUnit is never exposed.

See [here](https://github.com/lodash/lodash/blob/538491d773b43517f7d954f033d65e1f0905e173/dist/lodash.js#L179-L195) and [here](https://github.com/lodash/lodash/blob/538491d773b43517f7d954f033d65e1f0905e173/dist/lodash.js#L7333-L7359) for examples of working UMD.

PR #709, opened 2014-12-04
Build: Support Node.js export parity with CommonJS

JamesMGreene:
Ref #521
More info: https://github.com/jquery/qunit/issues/521#issuecomment-65656852

Basically, we want to ensure that `require('qunitjs').QUnit` works in all CommonJS environments, even if it is achieved by `QUnit.QUnit = QUnit`. Node.js was the only outlier at this point.

Also added and updated a few related tests.

### DIFF
--- a/src/export.js
+++ b/src/export.js
@@ -49,11 +49,14 @@ if ( typeof window !== "undefined" ) {
 }
 
 // For nodejs
-if ( typeof module !== "undefined" && module.exports ) {
+if ( typeof module !== "undefined" && module && module.exports ) {
 	module.exports = QUnit;
+
+	// For consistency with CommonJS environments' exports
+	module.exports.QUnit = QUnit;
 }
 
 // For CommonJS with exports, but without module.exports, like Rhino
-if ( typeof exports !== "undefined" ) {
+if ( typeof exports !== "undefined" && exports ) {
 	exports.QUnit = QUnit;
 }
--- a/test/globals-node.js
+++ b/test/globals-node.js
@@ -1,0 +1,20 @@
+/*jshint node:true */
+(function() {
+
+QUnit.module( "globals for Node.js only" );
+
+QUnit.test( "QUnit exports", function( assert ) {
+	var qunit = require( "../dist/qunit" );
+
+	assert.ok( qunit, "Required module QUnit truthy" );
+	assert.strictEqual( qunit, QUnit, "Required module QUnit matches global QUnit" );
+
+	assert.ok( qunit.hasOwnProperty( "QUnit" ), "Required module QUnit has property QUnit" );
+	assert.strictEqual(
+		qunit.QUnit,
+		qunit,
+		"Required module QUnit's property QUnit is self-referencing"
+	);
+});
+
+})();
--- a/test/globals.js
+++ b/test/globals.js
@@ -27,10 +27,6 @@ function checkExported( assert, methods, isAssertion ) {
 	}
 }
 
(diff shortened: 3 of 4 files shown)


jzaefferer:
@JamesMGreene could you update this?

test/globals.js
```diff
@@ -28,7 +28,7 @@ function checkExported( assert, methods, isAssertion ) {
 }
 
 QUnit.test( "QUnit object", function( assert ) {
-	assert.ok( QUnit instanceof QUnit.constructor, "Global QUnit built from it's own constructor" );
+	assert.ok( QUnit instanceof QUnit.constructor, "Global QUnit built from its own constructor" );
```
JamesMGreene: This used to be required because of the old-school logging setup (e.g. `QUnit.begin = fn;`). It can be dropped now that we no longer support that setup style.

src/export.js
```diff
@@ -49,11 +49,14 @@ if ( typeof window !== "undefined" ) {
...
+if ( typeof module !== "undefined" && module && module.exports ) {
 	module.exports = QUnit;
+
+	// For consistency with CommonJS environments' exports
+	module.exports.QUnit = QUnit;
 }
 
 // For CommonJS with exports, but without module.exports, like Rhino
-if ( typeof exports !== "undefined" ) {
+if ( typeof exports !== "undefined" && exports ) {
```
JamesMGreene: Double-checking what?  The first check (`typeof`) verifies that the variable has at least been declared, the second check verifies that its value is also truthy.

If I just did `if (exports)`, it would throw `ReferenceError`s... at least in oldIE. (e.g. http://stackoverflow.com/a/3390635/471696)

test/globals.js
leobalter: nice

test/globals-node.js
```diff
@@ -0,0 +1,25 @@
+/*jshint node:true */
+(function() {
+
+QUnit.module( "globals for Node.js only" );
+
+QUnit.test( "QUnit exports", function( assert ) {
+	var qunit = require( "../dist/qunit" );
+
+	assert.ok( qunit, "Required module QUnit truthy" );
```
JamesMGreene: I removed one of the following assertions but I intend to keep this one simple assertion.

JamesMGreene:
Added another commit to address a few of the PR feedback comments. I have no rebased it yet, just for the sake of seeing the changes to the 1st commit and the comment history. I'll happily rebase it whenever we think it's ready to be merged.

test/globals.js
JamesMGreene: Dropped.

test/globals-node.js
```diff
@@ -0,0 +1,25 @@
...
+	assert.ok(
+		qunit instanceof qunit.constructor,
+		"Required module QUnit built from its own constructor"
+	);
+
+	assert.strictEqual( qunit, QUnit, "Required module QUnit matches global QUnit" );
```
JamesMGreene: Dropped the constructor check.

JamesMGreene:
Eh, the commits are small enough. I've rebased against `master` now. My 2nd commit (now just rebased into the primary commit) was just removing 2 assertions (and 1 test) related to the now-irrelevant `QUnit.constructor`.

jzaefferer:
Looks good to me.

leobalter:
merged

JamesMGreene:
P.S. @jdalton Sorry, forgot to tag you on this earlier.

Issue #637, opened 2014-08-26
Add ability to mark a Test as skipped or incomplete


es:
Past discussion: #434

---

Would a PR adding the ability to skip tests be a welcome change? On a custom fork of TinyMCE we're modifying the functionality and as such certain tests are failing. We want to keep the tests in the codebase and are currently forced to comment them out. Having the ability to skip tests would be the best solution in this case.

The syntax would be `test.skip(func...)`.

I'll gladly make a PR for this, just looking for approval before I start working.

PR #714, opened 2014-12-09
Core: Skip tests with no/falsy callbacks

gibson042:
_I'm opening this as a pull request instead of an issue to make rejecting it easier if that is the final determination. If it is accepted, it should be coupled with a documentation change at [jquery/api.qunitjs.com](https://github.com/jquery/api.qunitjs.com), which I am more than happy to submit._

In https://github.com/jquery/qunit/issues/637#issuecomment-53467155 and https://github.com/jquery/qunit/pull/434#issuecomment-15652708, @JamesMGreene advocated for a syntax like `QUnit.test( "pending test" )`. 3f08a1aa1e03ec99eed7ebe31b36132ccc245a56 then added `QUnit.skip`, but not the above Mocha-like implicit skip. Although I was not a fan of that style at the time, thinking about how to isolate Sizzle dependencies in the jQuery core suite has me reconsidering. In particular, I'd like to take advantage of something very similar in order to inline test preconditions rather than wrap tests in if blocks: `QUnit.test( "name", preconditionsMet() && fnTest )`.

Implementation broke no existing unit tests, and in fact I found an opportunity to use the new functionality within the logs tests as I was updating them: ae84fc3008dab4c0860ef1d38c6273b303584d5f.

### DIFF
--- a/test/logs.js
+++ b/test/logs.js
@@ -8,6 +8,7 @@ var totalTests, moduleContext, moduleDoneContext, testContext, testDoneContext, 
 	testStart = 0,
 	testDone = 0,
 	log = 0,
+	isBrowser = typeof window !== "undefined" && window.document,
 	module1Context = {
 		name: "logs1",
 		tests: [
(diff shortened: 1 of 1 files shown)


binodpanta:
Did we come to a conclusion on this? I am looking for a feature such as JUnit's assumeTrue where the test will not run if a condition is false

platinumazure:
@binodpanta I don't know if this is something we want to reconsider for QUnit core, but this is something you can already do. You can just augment the QUnit object with a new method, as @jzaefferer [demonstrated above](https://github.com/qunitjs/qunit/pull/714#issuecomment-67961321).

PR #727, opened 2015-01-14
HTML Reporter: Fix regression for old markup

### LINKED ISSUE #716: Uncaught TypeError: Cannot set property 'innerHTML' of null at qunit:2638
I've just updated from 1.15 to 1.16 and one of my tests fail with:

`Uncaught TypeError: Cannot set property 'innerHTML' of null` at qunit.js:2638

At line 2638 in qunit.js I see:
```javascript
if ( tests ) {
  id( "qunit-testresult" ).innerHTML = html;
}
```

Until now I've used markup template like:
```html
<body>
	<h1 id="qunit-header">Wicket Ajax tests</h1>
	<h2 id="qunit-banner"></h2>
	<div id="qunit-testrunner-toolbar"></div>
	<h2 id="qunit-userAgent"></h2>
	<ol id="qunit-tests"></ol>
       <div id="qunit-fixture"></div>
```

Looking at http://qunitjs.com/ I see that now the template should be just:
```html
<body>
  <div id="qunit"></div>
  <div id="qunit-fixture"></div>
```

This fixed my test !

Problem 1) I didn't see a HEADS UP about this change in a minor release
Problem 2) Now the header says "All tests". Is there a way to use "Wicket Ajax tests" as before ? 
Problem 3) The `qunit-header` is no more a link. `qunit-banner` contains a link but it is not clickable because it doesn't have text child element:
```html
<h2 id="qunit-banner" class="qunit-pass"><a href="http://localhost/ajax-tests/test/js/all.html"></a> </h2>
```

JamesMGreene:
Also, fixed regression from `1.15.0` for adding a link to #qunit-header.

Fixes #716

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -405,14 +405,21 @@ function appendToolbar() {
 	}
 }
 
+function appendHeader() {
+	var header = id( "qunit-header" );
+
+	if ( header ) {
+		header.innerHTML = "<a href='" +
+			setUrl({ filter: undefined, module: undefined, testId: undefined }) +
+			"'>" + header.innerHTML + "</a> ";
+	}
+}
+
 function appendBanner() {
 	var banner = id( "qunit-banner" );
 
 	if ( banner ) {
 		banner.className = "";
-		banner.innerHTML = "<a href='" +
-			setUrl({ filter: undefined, module: undefined, testId: undefined }) +
-			"'>" + banner.innerHTML + "</a> ";
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jzaefferer:
Looks good. I wonder if we should add something to prevent further regressions. I'd rather not add more html pages, but can you try to add a `test()` to an existing page?

leobalter:
LGTM.

Imho, I think we should think the entire test suite for the reporter, at least something covering more details, in another PR, of course.

jzaefferer:
I've tested this manually by replacing the `#qunit` div in `test/index.html` wit this block:

```html
    <h1 id="qunit-header">QUnit Test Suite</h1>
    <h2 id="qunit-banner"></h2>
    <div id="qunit-testrunner-toolbar"></div>
    <h2 id="qunit-userAgent"></h2>
    <ol id="qunit-tests"></ol>
```

Also clicked on the header link after some rerun link, that also worked fine, with both markup options.

Let's keep in mind to add html reporter tests before making any other changes to it.

PR #730, opened 2015-01-16
Build: Ignore BrowserStack generated file

leobalter:
BrowserStack-runner is generating this binary file to the project's root

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ dist
 node_modules
 build/report
 browserstack-run.pid
+BrowserStackLocal


JamesMGreene:
Guessing it's from the local tunnel? Anyway, LGTM.

leobalter:
Yes, it is.

jzaefferer:
Let's report this on their end, instead of ignoring the issue. I've ignored it for too long myself.

jzaefferer:
Same for that pid file.

leobalter:
Makes sense. :+1:

jzaefferer:
Created a ticket: https://github.com/browserstack/browserstack-runner/issues/105

PR #731, opened 2015-01-19
HTML Reporter: handle case where unit root element is not present

jonbretman:
I suddenly started getting this error when using QUnit with Karma:

```
PhantomJS 1.9.8 (Mac OS X) ERROR
  TypeError: 'null' is not an object (evaluating 'toolbar.appendChild')
  at /Users/jonbretman/git/lyst/node_modules/qunitjs/qunit/qunit.js:2534
```

After some digging found that this bug was introduced with [this commit](https://github.com/jquery/qunit/commit/3fa5f6a4e4ead6b588376f6621fc3c4baab57444#diff-5b7e94e2a10b848c4ae4b539388f86ae). Before `appendToolbar` would check that the toolbar (`#qunit-testrunner-toolbar`) existed before trying to append to it, but the above mentioned commit changed it so that `toolbarModuleFilter` did the appending to the toolbar directly, but without checking first that it exists.

### DIFF
--- a/test/reporter-html-no-qunit-element.js
+++ b/test/reporter-html-no-qunit-element.js
@@ -1,0 +1,10 @@
+(function() {
+
+QUnit.module( "reporter-html-no-qunit-element");
+
+QUnit.test( "reporter/html when no qunit element present", function( assert ) {
+	assert.expect( 1 );
+    assert.ok(true);
+});
+
+})();
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -96,7 +96,8 @@ grunt.initConfig({
 			"test/autostart.html",
 			"test/startError.html",
 			"test/logs.html",
-			"test/setTimeout.html"
+			"test/setTimeout.html",
+			"test/reporter-html-no-qunit-element.html"
 		]
 	},
 	coveralls: {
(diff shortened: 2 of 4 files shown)


leobalter:
This fix #732. 

I'm just not sure about shipping this PR with the new test file, but that should be good until we land 2.0.0. 

Meanwhile, I'll just address some code styles inconsistencies, you can fix it or I'll do it here before merging.

test/reporter-html-no-qunit-element.js
```diff
@@ -0,0 +1,10 @@
+(function() {
+
+QUnit.module( "reporter-html-no-qunit-element");
```
leobalter: missing space before closing parenthesis.a

test/reporter-html-no-qunit-element.js
```diff
@@ -0,0 +1,10 @@
...
+
+QUnit.test( "reporter/html when no qunit element present", function( assert ) {
+	assert.expect( 1 );
+    assert.ok(true);
```
leobalter: use tabs instead of spaces for indentation and missing spaces between parenthesis.

leobalter:
@jzaefferer, @JamesMGreene: I did some changes to be squashed with this commit, please take a look at: https://github.com/leobalter/qunit/compare/jquery:master...fix-html-reporter-toolbar-bug

jonbretman:
@leobalter thanks for picking this up so quickly.

leobalter:
@jonbretman, thank you! 

Meanwhile, would you mind to sign the [CLA](http://contribute.jquery.org/CLA/)?

jonbretman:
@leobalter done.

PR #734, opened 2015-01-20
Fix html reporter toolbar bug

### LINKED ISSUE #732: 1.17.0 & phantomjs
Heads up, I'm seeing tests hang when using phantomjs as `phantomjs test.js`. It looks like things are getting exported correctly and the `QUnit.load()` is getting called. I'll know more this evening when I dig-in fully.

jzaefferer:
Creating a PR for @leobalter's branch, to have a proper reference when merging.

Fixes #732
Closes #731

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -96,7 +96,9 @@ grunt.initConfig({
 			"test/autostart.html",
 			"test/startError.html",
 			"test/logs.html",
-			"test/setTimeout.html"
+			"test/setTimeout.html",
+			"test/reporter-html-legacy-markup.html",
+			"test/reporter-html-no-qunit-element.html"
 		]
 	},
 	coveralls: {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -384,7 +384,7 @@ function toolbarModuleFilter() {
 		moduleFilter = document.createElement( "span" ),
 		moduleFilterHtml = toolbarModuleFilterHtml();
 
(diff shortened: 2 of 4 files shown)


Gruntfile.js
```diff
@@ -96,7 +96,9 @@ grunt.initConfig({
 			"test/autostart.html",
 			"test/startError.html",
 			"test/logs.html",
-			"test/setTimeout.html"
+			"test/setTimeout.html",
+			"test/reporter-html-legacy-markup.html",
+			"test/reporter-html-no-qunit-element.html"
```
jzaefferer: Without the reporter fix below, this test just times out. Let's try to improve this to produce a useful failure.

jzaefferer:
I think its fine to run those two new pages only through grunt/phantom. No need to blow up the runtime for browserstack-runner even further.

@leobalter can you look into producing a "proper failure", instead of the timeout?

Gruntfile.js
jonbretman: Yeh I noticed this too, tried to make it fail with an error but got lost :( A proper failure would be better than timing out. I assume it's because the error is being suppressed somewhere?

leobalter:
the better failure I could produce was this last commit, setting a timeout in order to print something at the PhantomJS reporter.

Throwing an error, etc, doesn't print anything there.

jzaefferer:
Good enough, I guess. Can you squash and merge?

PR #735, opened 2015-01-20
[WIP]: ES6 Modules

leobalter:
This is a WIP.

I've tried to set everything in ES6 modules, which showed to be more difficult than I thought with all the interdependencies in our code.

Currently, we need to make every module true independent, with that we can achieve the desired progress in this PR.

Also, the Browserify with 6to5 plugins aren't working as I expected. Bundling things still get then to use a require system, which I'm not aiming. 

I'll follow @jzaefferer's suggestion and try http://jspm.io/


leobalter:
Other problem: I still need to see how to handle JSCS to accept ES6. Probably easy, I just still need to go that further.

jzaefferer:
>  I still need to see how to handle JSCS to accept ES6. Probably easy, I just still need to go that further.

We're not running jshint and jscs against `src/` anyway, right? As long as the output in `dist/` is more or less the same, there shouldn't be much of a difference.

Although, with proper modules and dependencies actually defined, we should also be able to validate `src` files directly...

package.json
```diff
@@ -33,8 +33,10 @@
   ],
   "dependencies": {},
   "devDependencies": {
+    "6to5ify": "^3.1.2",
```
jzaefferer: Let's stick with fixed version numbers, drop the `^`.

src/assert.js
```diff
@@ -1,9 +1,14 @@
+import { pauseProcessing, resumeProcessing, objectValues } from "./core";
```
jzaefferer: Is the "./" always required here? I haven't seen that in ES6 module examples before.

src/core.js
```diff
@@ -1,8 +1,25 @@
+/*!
+ * QUnit @VERSION
```
jzaefferer: Replacing "@VERSION" was done by the concat task, via `process` option. Does browserify handle that?

leobalter:
> Although, with proper modules and dependencies actually defined, we should also be able to validate src files directly...

One of my main ideas on converting our modules into the es6 format is to allow running jshint and jscs against `src/` files, are they're the ones we should test, not a built file. 

Also, JSHint helped me a lot on fixing most of dependencies connections in this PR.

src/assert.js
leobalter: it's not required by ES6, it's required by browserify..

src/core.js
```diff
@@ -140,6 +95,27 @@ config = {
 // Push a loose unnamed module to the modules collection
 config.modules.push( config.currentModule );
 
+import Test from "./test";
```
jzaefferer: Why is this `import` down here?

package.json
leobalter: wip. I'll fix that anyway.

jzaefferer:
Well, now I see what you're talking about. Having modules depend on core, while those modules are also imported by core, is pretty bad. I think we need to, eventually, get rid of all these circular dependencies.

That said, overall it looks like we're on the right track. The import and export statements make sense and are easy enough to follow.

src/core.js
leobalter: interdependency issue. Test need QUnit.config properly set. :(

src/equiv.js
```diff
@@ -1,219 +1,214 @@
 // Test for equality any JavaScript type.
 // Author: Philippe Rathé <prathe@gmail.com>
-QUnit.equiv = (function() {
-
-	// Call the o related callback with the given arguments.
```
leobalter: this file got a huge diff because the indentation level reduced here. To ignore the whitespace change, see this diff including the [`?w=1`](https://github.com/jquery/qunit/pull/735/files?w=1) parameter.

src/assert.js
JamesMGreene: Can we just use [grunt-6to5](https://www.npmjs.com/package/grunt-6to5) instead?

leobalter: it will convert es6 to CommonJS modules, we would need to run Browserify after that in order to bundle them.

JamesMGreene: Ah, right, my bad.

leobalter: It's not a problem, I only know this because it's what I have tried before.

jzaefferer:
Instead of jspm we should try to use esperanto, like we now do (or started doing) in PEP: https://github.com/jquery/PEP/pull/157

leobalter:
This is getting more complex than expected and than I desired.

We have a lot of interdependencies that brings that complexity, I can make some inline comments to show it.

running `grunt build` will show some of the missing parts, but even if we solve the remaining, I don't like the interdependencies as they are right now. Solving them will look like a totally rewritten code, even if I don't try this.

At least, `equiv`, `jsdiff` and `dump` are easier to treat, as we still need to solve problems in `test`, `assert`, and the `html reporter` modules.

src/core.js
```diff
@@ -1,144 +1,73 @@
...
+
 var QUnit,
-	config,
 	onErrorFnPrev,
 	loggingCallbacks = {},
-	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//, "" ),
 	toString = Object.prototype.toString,
 	hasOwn = Object.prototype.hasOwnProperty,
 	// Keep a local reference to Date (GH-283)
 	Date = window.Date,
```
leobalter: we don't have the window as the global reference with these modules, it's necessary to fix it.

src/core.js
```diff
@@ -186,6 +115,8 @@ config.modules.push( config.currentModule );
 	QUnit.isLocal = location.protocol === "file:";
 }());
 
+export var urlParams = QUnit.urlParams;
```
leobalter: this shouldn't be exported this way, but the HTML Reporter needs it.

src/test.js
```diff
@@ -1,3 +1,13 @@
+import { extend, runLoggingCallbacks, now, saveGlobal, checkPollution, synchronize, defined, inArray } from "./core";
+import config from "config";
+import sourceFromStacktrace from "sourceFromStacktrace";
+import { extractStacktrace } from "sourceFromStacktrace";
+import Assert from "assert";
```
leobalter: these imports from core are wrong.

reporter/html.js
```diff
@@ -1,82 +1,11 @@
...
-
-	if ( tests ) {
-		result = document.createElement( "p" );
-		result.id = "qunit-testresult";
-		result.className = "result";
-		tests.parentNode.insertBefore( result, tests );
-		result.innerHTML = "Running...<br />&#160;";
-	}
-};
+import QUnit, { config, defined, urlParams } from "../src/core";
```
jzaefferer: Maybe we shouldn't define the reporter as a ES6 module? Its supposed to use the public/global API exported by QUnit, not the core module itself.

src/assert.js
jzaefferer: Can you change this, now that you're using esperanto?

src/build.js
```diff
@@ -0,0 +1,20 @@
+import QUnit from "core";
+import diff from "../external/jsdiff/jsdiff";
+import dump from "dump";
+import equiv from "equiv";
+import { begin, done, testStart, testDone, log } from "../reporter/html";
```
jzaefferer: This should disappear if we undo the conversion of the reporter to a ES6 module.

src/core.js
jzaefferer: Does that only apply to `window`? What about `setTimeout` and other globals? How do you access the global scope correctly?

src/dump.js
```diff
@@ -1,263 +1,260 @@
+import { inArray } from "./core";
```
jzaefferer: Can extract `inArray` from core into a standalone module to avoid the circular dependency here.

src/test.js
jzaefferer: Yeah, those need to be in their own modules, or whatever is necessary to get rid of the circular dependency.

src/test.js
```diff
@@ -1,3 +1,13 @@
+import { extend, runLoggingCallbacks, now, saveGlobal, checkPollution, synchronize, defined, inArray } from "./core";
+import config from "config";
+import sourceFromStacktrace from "sourceFromStacktrace";
+import { extractStacktrace } from "sourceFromStacktrace";
```
jzaefferer: Seems strange to have two imports statements from a single module

leobalter:
I believe it's better stop this PR for now, and treat some circular dependencies that will allow us to turn everything into ES6 Modules with something not such aggressive in the changes.

The code isn't designed for modules, but I've identified a lot of good and bad things in this PR which I'll will help in future work.

Some of the good parts:

- esperanto worked just fine
- jsdiff, equiv and dump will be easy to handle
- The grunt-esperanto module is getting some improvements to be able to used with UMD

Some parts still needed to improve:

- The html reporter has some circular dependencies on core's config and urlParams. That shouldn't happen. Maybe the build should come from it, as it requires QUnit and attach the logging methods.
- Test and Assert share a lot of cross dependencies with core, most of them can be identified to be detached from core.

Issue #693, opened 2014-11-19
Make it more obvious why test execution order changes


jzaefferer:
When `QUnit.config.reorder` is enabled and previously failed tests run first, we should show "Rerunning failed tests:" or something like that in the summary of the HTML reporter, replacing the default "Running tests:". This text will also be readable when async tests take long enough or the testrunner stops. In the context of confusion about non-atomic tests suddenly failing, this may still help.

PR #737, opened 2015-01-20
Retain failed test numbers when using hidepassed by moving passed tests off-screen instead of hiding them

insin:
When using the `hidepassed` option, since an `<ol>` is providing test numbers, the original numbers for any failed tests which follow passed tests are lost when `<li>`s for passed tests are hidden.

Moving `<li>`s for passed tests off the screen instead retains the failed test numbers.

Before:
qunit-before

After:
qunit-after

### DIFF
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -116,7 +116,11 @@
 
 #qunit-tests.hidepass li.running,
 #qunit-tests.hidepass li.pass {
-	display: none;
+	position: absolute;
+	left: -100px;
+	width: 1px;
+	height: 1px;
+	overflow: hidden;
 }
 
 #qunit-tests li strong {
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -116,11 +116,12 @@
 
 #qunit-tests.hidepass li.running,
 #qunit-tests.hidepass li.pass {
-	position: absolute;
-	left: -100px;
-	width: 1px;
-	height: 1px;
-	overflow: hidden;
+	visibility: hidden;
+	width:   0px;
+	height:  0px;
+	padding: 0;
+	border:  0;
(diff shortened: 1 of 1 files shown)


leobalter:
This is interesting and depends on the user PoV, because it might be strange - even with hidepassed on - to see a list of tests like:
```
2 - foo
5 - bar
8 - baz
```
In general, these numbers shouldn't be used as indexes. Even though, this might build a good reference for finding it when disabling the hidepassed option.

jzaefferer:
#693 is somewhat related. If we land this, we should probably try to apply the same when using the Rerun feature.

JamesMGreene:
If we want to achieve this effect, I strongly recommend we do not use this implementation (no offense, @insin) as it will force a horizontal scrollbar to appear for any consumers with RTL text direction.

Here's a relatively simple alternative that preserves the list numbering as well:

```css
#qunit-tests.hidepass li.running,
#qunit-tests.hidepass li.pass {
  visibility: hidden;
  width:   0px;
  height:  0px;
  margin:  0;
  padding: 0;
  border:  0;
}
```

You can see it live here:
&nbsp; &nbsp; http://jsbin.com/socim/1/edit?html,css,js,output

JamesMGreene:
FYI, @insin updated this PR with a 2nd commit to use the CSS styles I recommended.

leobalter:
Squashed and merged, thanks!

leobalter:
I'm sorry, @insin, I forgot to check if you signed the jQuery CLA, would you please sign it? 

http://contribute.jquery.org/CLA/

insin:
Signed the CLA.

leobalter:
Thank you

PR #746, opened 2015-02-05
[Core]: Remove JSDoc comments at the assertion methods

leobalter:
As the assertions are already documented within the QUnit public API, we can remove these JSDoc comments, avoiding duplication, less documented areas, etc.


jzaefferer:
Looks good to me.

JamesMGreene:
:+1:

PR #747, opened 2015-02-06
Build: use es6 modules from concatenated QUnit

leobalter:
Here's another not-working experiment. It was easy to make it work on the browsers, but it's conflicting on the node envs.

cc @jzaefferer

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,3 +1,4 @@
 {
-	"preset": "jquery"
+	"preset": "jquery",
+	"esnext": true
 }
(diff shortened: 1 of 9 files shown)


scottgonzalez:
I'm not sure what you're trying to accomplish here. If the goal is to actually use ES6 modules, then you should actually use ES6 modules, not just hack it into the outro. I don't see any benefit at all to attempting this.

leobalter:
If by doing this to at least remove that export code in `src/export.js` and having some progress to get the code ready for ES6, it would be something.

PR #752, opened 2015-02-12
Update ignoreGlobalErrors only if this.test is defined

mdartic:
I have some tests failing around these lines since 1.17.1 (before I was in 1.14.0).

Changing these lines made my tests work again.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -119,15 +119,17 @@ QUnit.assert = Assert.prototype = {
 			expected = null;
 		}
 
-		if (this.test)
+		if (this.test) {
 			this.test.ignoreGlobalErrors = true;
+		}
 		try {
 			block.call( this.test.testEnvironment );
 		} catch (e) {
 			actual = e;
 		}
-		if (this.test)
+		if (this.test) {
 			this.test.ignoreGlobalErrors = false;
+		}
 
 		if ( actual ) {
 			expectedType = QUnit.objectType( expected );
--- a/src/assert.js
+++ b/src/assert.js
@@ -119,13 +119,15 @@ QUnit.assert = Assert.prototype = {
 			expected = null;
 		}
 
-		this.test.ignoreGlobalErrors = true;
+		if (this.test)
+			this.test.ignoreGlobalErrors = true;
 		try {
 			block.call( this.test.testEnvironment );
 		} catch (e) {
 			actual = e;
 		}
-		this.test.ignoreGlobalErrors = false;
+		if (this.test)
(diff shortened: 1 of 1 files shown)


JamesMGreene:
Could you please show us an example test that is causing this issue?

Are you using global assertion references (e.g. `notEqual` and/or direct references on the `QUnit.assert` object (e.g. `QUnit.assert.notEqual`) rather than local instantiated references? e.g.

```js
QUnit.test("blah", function(localAssert) {
  localAssert.notEqual(true, false, "foobar");
});
```

Any which way should be fine and supported in the `1.x` series, so this sounds like a regression. However, you should be aware that only the loral assertion reference style will be supported in the upcoming `2.x` series.

mdartic:
Well, I've create this repo to test it : https://github.com/mdartic/qunit_752

And, indeed, I use direct references on the QUnit.assert...

We plan to integrate the local assertion in our tests, but for the moment we have a lot 'direct' style.

JamesMGreene:
How about if you use the global references? e.g.

```js
QUnit.test("If I use global assertions, it _____?", function() {
    expect(2);

    deepEqual(isNotNull('pouet'), true);
    
    throws(function () {
        isNotNull(null);
    });
});
```

mdartic:
I've update the test and... well, you're right !

I let you see if it's a good thing to integrate the check of 'this.test'.

Thanks for your help.

leobalter:
@mdartic, [QUnit.assert.push](https://github.com/jquery/qunit/blob/master/src/assert.js#L46) makes a test reference to the current running test when it's not present. It doesn't break the other assertions because they basically just call that method, but that doesn't happen on `QUnit.assert.throws`.

The fail is present at this [JSBin example](http://jsbin.com/quvupolaso/1/edit?js,output) where it fails when using the `QUnit.assert.throws` directly.

```js
function throwSomething() {
  throw 'something';
}

QUnit.test("Throw tests", function( assert ) {
    assert.throws(throwSomething);
    throws(throwSomething);
    QUnit.assert.throws(throwSomething);
});
```

Instead changing the method behavior, `throws` should get the same current test reference as `push` at this line:

```js
var assert = this,
    currentTest = ( assert instanceof Assert && assert.test ) || QUnit.config.current;
```

and use `currentTest` instead of `this.test`

jzaefferer:
Whatever the underlying issue, the proposed fix isn't acceptable. At best it hides an error, but doesn't actually solve it.

Please file a new issue if the problem persists.

PR #753, opened 2015-02-17
Fix #749 - Filter does not work correctly with upper case letter in test name

### LINKED ISSUE #749: 1.17.1 - Filter does not work correctly with upper case letter in test name
Steps to reproduce:
1. Open http://jsfiddle.net/nxrLcppq/
2. Type "Test Name" in "Filter" input and click "Go"
3. No any tests will be run

Compare results with "test name" filter string.

Seems, this issue was appeared after https://github.com/jquery/qunit/commit/1d44eb38052d25d97a631b9711e481303fc472b6 , where
    filter = config.filter && config.filter.toLowerCase(),
was replaced by
    filter = config.filter,

TheCloudlessSky:
As of 1.17.1, filters do not work when the module name contains upper case letters (#749) which was introduced by 1d44eb38052d25d97a631b9711e481303fc472b6.

This fixes it by changing `valid()` to lowercase the filter before using it further.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -345,7 +345,7 @@ Test.prototype = {
 
 	valid: function() {
 		var include,
-			filter = config.filter,
+			filter = config.filter && config.filter.toLowerCase(),
 			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
 			fullName = ( this.module.name + ": " + this.testName ).toLowerCase();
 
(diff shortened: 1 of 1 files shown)


jzaefferer:
Looks good to me.

leobalter:
:+1:

leobalter:
CLA checked.

Thanks, @TheCloudlessSky.

PR #756, opened 2015-03-01
HTML Reporter: Fix test filter without any module

### LINKED ISSUE #748: test filter input fails without any module
Steps to reproduce:
1. Open http://jsfiddle.net/gj6hudry/ sample.
2. Type 'testname' in 'Filter' input and click 'Go'.
3. Error "Cannot read property 'options' of null" will be occurred at:
    selection = decodeURIComponent( selectBox.options[ selectBox.selectedIndex ].value ),

leobalter:
Without any module at the test suite, the Reporter doesn't create
the select tag to filter the modules. The test filter was trying
to get the selected value to set the filter in order to keep it
at the url parameters.

Closes #748

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -289,12 +289,16 @@ function setUrl( params ) {
 }
 
 function applyUrlParams() {
-	var selectBox = id( "qunit-modulefilter" ),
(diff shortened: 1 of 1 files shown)


jzaefferer:
:+1:

Issue #706, opened 2014-12-01
Show source location of passed tests


jzaefferer:
Every now and then I'm looking at a green suite and want to know where the source for a specific test is located. We capture that information already, for all tests, but only output it when a test fails. We could always output for non-collapsed tests, or only for non-collapsed passing tests (since failing tests will output it already in some way).

leobalter:
By the passing tests you mean print the source location for each passed assertion, right?

With the global assertions, the test might not be enough for this. A assertion location is more precise.

jzaefferer:
No, I meant test. We already capture that information, so would just have to output it.

sammyshj:
I would like to work on this issue.

shivamdixit:
Hey Sambuddha, I think @gauravmittal1995 has already worked on the issue.
Refer #758

sammyshj:
Thanks for pointing that out @shivamdixit

jzaefferer:
This has been addressed in 9498355593a1284720e8accfbf6f37f72cc82a9b (didn't include a ticket reference)

PR #758, opened 2015-03-03
Test: Source Displayed Even for Passed Test

gauravmittal1995:
Related to : https://github.com/jquery/qunit/issues/706

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -668,7 +668,7 @@ QUnit.log(function( details ) {
 
 QUnit.testDone(function( details ) {
 	var testTitle, time, testItem, assertList,
-		good, bad, testCounts, skipped,
+		good, bad, testCounts, skipped, sourceName,
 		tests = id( "qunit-tests" );
 
 	if ( !tests ) {
@@ -691,8 +691,13 @@ QUnit.testDone(function( details ) {
 		}
 	}
 
+	sourceName = document.createElement( "p" );
+	sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
+	testItem.appendChild( sourceName );
+
 	if ( bad === 0 ) {
 		addClass( assertList, "qunit-collapsed" );
+		addClass( sourceName, "qunit-collapsed" );
 	}
 
 	// testItem.firstChild is the test name
@@ -711,9 +716,13 @@ QUnit.testDone(function( details ) {
 		skipped.className = "qunit-skipped-label";
 		skipped.innerHTML = "skipped";
 		testItem.insertBefore( skipped, testTitle );
+		addEvent( testTitle, "click", function() {
+			toggleClass( sourceName, "qunit-collapsed" );
+		});
 	} else {
 		addEvent( testTitle, "click", function() {
 			toggleClass( assertList, "qunit-collapsed" );
+			toggleClass( sourceName, "qunit-collapsed" );
 		});
 
 		testItem.className = bad ? "fail" : "pass";
@@ -723,6 +732,7 @@ QUnit.testDone(function( details ) {
 		time.innerHTML = details.runtime + " ms";
 		testItem.insertBefore( time, assertList );
 	}
+
 });
 
 if ( !defined.document || document.readyState === "complete" ) {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -724,13 +724,16 @@ QUnit.testDone(function( details ) {
 		testItem.insertBefore( time, assertList );
 	}
 
-	sourceName = document.createElement( "p" );
-	sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
-	addClass( sourceName, "qunit-collapsed qunit-source" );
-	addEvent( testTitle, "click", function() {
-		toggleClass( sourceName, "qunit-collapsed" );
-	});
-	testItem.appendChild( sourceName );
+	// Only add this information if stackTrace is supported
+	if ( Error().stack ) {
+		sourceName = document.createElement( "p" );
+		sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
+		addClass( sourceName, "qunit-collapsed qunit-source" );
+		addEvent( testTitle, "click", function() {
+			toggleClass( sourceName, "qunit-collapsed" );
+		});
+		testItem.appendChild( sourceName );
+	}
 });
 
 if ( !defined.document || document.readyState === "complete" ) {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -728,7 +728,10 @@ QUnit.testDone(function( details ) {
 	if ( Error().stack ) {
 		sourceName = document.createElement( "p" );
 		sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
-		addClass( sourceName, "qunit-collapsed qunit-source" );
+		addClass( sourceName, "qunit-source" );
+		if ( bad === 0 ) {
+			addClass( sourceName, "qunit-collapsed" );
+		}
 		addEvent( testTitle, "click", function() {
 			toggleClass( sourceName, "qunit-collapsed" );
 		});
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -691,13 +691,8 @@ QUnit.testDone(function( details ) {
 		}
 	}
 
-	sourceName = document.createElement( "p" );
-	sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
-	testItem.appendChild( sourceName );
-
 	if ( bad === 0 ) {
 		addClass( assertList, "qunit-collapsed" );
-		addClass( sourceName, "qunit-collapsed" );
 	}
 
 	// testItem.firstChild is the test name
@@ -716,13 +711,9 @@ QUnit.testDone(function( details ) {
 		skipped.className = "qunit-skipped-label";
 		skipped.innerHTML = "skipped";
 		testItem.insertBefore( skipped, testTitle );
-		addEvent( testTitle, "click", function() {
-			toggleClass( sourceName, "qunit-collapsed" );
-		});
 	} else {
 		addEvent( testTitle, "click", function() {
 			toggleClass( assertList, "qunit-collapsed" );
-			toggleClass( sourceName, "qunit-collapsed" );
 		});
 
 		testItem.className = bad ? "fail" : "pass";
@@ -733,6 +724,13 @@ QUnit.testDone(function( details ) {
 		testItem.insertBefore( time, assertList );
 	}
 
+	sourceName = document.createElement( "p" );
+	sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
+	addClass( sourceName, "qunit-collapsed qunit-source" );
+	addEvent( testTitle, "click", function() {
+		toggleClass( sourceName, "qunit-collapsed" );
+	});
+	testItem.appendChild( sourceName );
 });
 
 if ( !defined.document || document.readyState === "complete" ) {
--- a/src/test.js
+++ b/src/test.js
@@ -202,6 +202,9 @@ Test.prototype = {
 			assertions: this.assertions,
 			testId: this.testId,
 
+			// Source of Test
+			source: this.stack,
+
 			// DEPRECATED: this property will be removed in 2.0.0, use runtime instead
 			duration: this.runtime
 		});
--- a/test/logs.js
+++ b/test/logs.js
@@ -175,10 +175,16 @@ QUnit.test( module1Test2.name, function( assert ) {
 	// TODO: more tests for testDoneContext.assertions
 
 	delete testDoneContext.runtime;
+
 	// DEPRECATED: remove this delete when removing the duration property
 	delete testDoneContext.duration;
+
 	// Delete testDoneContext.assertions so we can easily jump to next assertion
 	delete testDoneContext.assertions;
+
+	// Delete testDoneContext.source
+	delete testDoneContext.source;
+
 	assert.deepEqual( testDoneContext, {
 		module: module1Context.name,
 		name: module1Test1.name,
@@ -259,6 +265,7 @@ QUnit.test( module2Test4.name, function( assert ) {
 
 	delete testDoneContext.runtime;
 	delete testDoneContext.duration;
+	delete testDoneContext.source;
 
 	assert.deepEqual( testDoneContext, {
 		assertions: [],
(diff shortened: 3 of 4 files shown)


src/test.js
```diff
@@ -272,12 +272,9 @@ Test.prototype = {
 				runtime: now() - this.started
 			};
 
-		if ( !result ) {
-			source = sourceFromStacktrace();
-
-			if ( source ) {
-				details.source = source;
-			}
+		source = sourceFromStacktrace();
```
leobalter: You can pass a parameter to `sourceFromStacktrace` to set the exact line you need to cap end send. The default (no parameter) works fine when the result is falsy, but it's not the same when it pass.

The example below - of the tests running with this current changes - seems we need to set the parameter to start with that second line (` at file:///Users/leobalter/dev/qunit/test/logs.js:120`).

Also, it's probably good to break this information to give just a single line location.

```
"    at temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9
     at file:///Users/leobalter/dev/qunit/test/logs.js:120
     at temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9
     at temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9
     at process (temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9)
     at begin (temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9)
     at temp//@1_Users@1_leobalter@1_dev@1_qunit@1_dist@1_qunit.js:9"
```

gauravmittal1995:
@leobalter Done. Please Review

src/test.js
```diff
@@ -272,12 +272,9 @@ Test.prototype = {
...
 
-		if ( !result ) {
-			source = sourceFromStacktrace();
-
-			if ( source ) {
-				details.source = source;
-			}
+		source = sourceFromStacktrace( 4 );
+		if ( source ) {
+			details.source = source;
```
leobalter: As I mentioned before, the sourceFromStacktrace parameter must be different when the test fail, otherwise you're just gonna solve the problem when it pass, before it was good when the tests failed.

test/logs.js
```diff
@@ -107,6 +107,7 @@ QUnit.test( module1Test1.name, function( assert ) {
 	);
 
 	delete logContext.runtime;
+	delete logContext.source;
```
leobalter: this is a harming flag to the tests, we still need to test that tests contains this source property, even if we than later remove it here to test exact values on other tests.

jzaefferer:
#706 is about showing the location of passed tests(!), not passed assertions.

gauravmittal1995:
@jzaefferer @leobalter Made Changes. Please Review

jzaefferer:
The output needs to be inside the collapsed element, so that its only shown when opening details of a test. Currently this source line is shown for every test, which looks pretty bad.

gauravmittal1995:
@jzaefferer Please Review

test/reporter-html.js
```diff
@@ -106,7 +106,7 @@ QUnit.test( "setup", function( assert ) {
 QUnit.test( "basics", function( assert ) {
 	assert.expect( 2 );
 	var previous = getPreviousTests( /^setup$/, /^timing$/ )[ 0 ],
-		runtime = previous.lastChild.previousSibling;
+		runtime = previous.lastChild.previousSibling.previousSibling;
```
leobalter: Here's a good opportunity to include tests for these changes.

Would you include a test with a regexp containing the static part of the source line?

test/logs.js
```diff
@@ -179,6 +179,8 @@ QUnit.test( module1Test2.name, function( assert ) {
 	delete testDoneContext.duration;
 	// Delete testDoneContext.assertions so we can easily jump to next assertion
 	delete testDoneContext.assertions;
+	// Delete testDoneContext.source
```
leobalter: The code style requires a blank space above comments, and you may also include the missing blank on the other comment above, if you don't mind.

leobalter:
Working good for me, now it's just about to add a test and fix the code style issues.

reporter/html.js
```diff
@@ -723,6 +727,10 @@ QUnit.testDone(function( details ) {
 		time.innerHTML = details.runtime + " ms";
 		testItem.insertBefore( time, assertList );
 	}
+
+	sourceName.innerHTML = "<b class='counts'>Source: </b>" + details.source;
+	assertList.parentNode.insertBefore(sourceName, assertList.nextSibling);
```
leobalter: missing space before and after arguments list.

```
assertList.parentNode.insertBefore( sourceName, assertList.nextSibling );
```

gauravmittal1995:
@leobalter Hey, What test is needed? Can u explain a bit more?

reporter/html.js
leobalter: I was playing with the code and we can do this easier:

```js
        sourceName = document.createElement( "p" );
	sourceName.innerHTML = "<strong>Source: </strong>" + details.source;
	addClass( sourceName, "qunit-collapsed qunit-source" );
	addEvent( testTitle, "click", function() {
		toggleClass( sourceName, "qunit-collapsed" );
	});
	testItem.appendChild( sourceName );
```

- The source can be inserted to any test
- it shouldn't use `counts` class, but just a strong tag on the `Source: ` text.
- the source element now uses a `qunit-source` class.
- the click event is made individually, this way you can still have the source information on skipped tests
- it is appended on the testItem element.

gauravmittal1995:
@leobalter Hey, Made the changes. Can u explain the test that is needed??

leobalter:
you need to add test for the changes you made, assert the previous test added a source line.

leobalter:
also: https://github.com/jquery/qunit/pull/758#discussion_r25803143

reporter/html.js
gauravmittal1995: @leobalter I tried the above. But gives error for addEvent because it calls addevent two time for testTitle click. 
Changed  innerhtml  to strong and used appendChild instead of addBefore.
Seems to be working fine now.

gauravmittal1995:
@leobalter How do i add test for the change??

jzaefferer:
HTML reporter output for passing tests looks good to me now. May want to hide that for failing tests, since the failing assertion already includes source lines.

leobalter:
I'm writing the tests and will land them as soon as I fix this PR on IE6 to 9, they're currently  saying "Source: undefined"

gauravmittal1995:
@jzaefferer Hey, IMHO, i think that we should keep the source even for failed test. Since failed test only show lines of fail assertion, a failed test can still have some passed assertions as well. Also the test can be big and the assertion called much later than where the test starts. 
Also this will also be consistent with the passed tests.

This is what i think. I could be wrong. Please let me know if i should hide it for failed assertions.

leobalter:
> i think that we should keep the source even for failed test. Since failed test only show lines of fail assertion

Agreed on of this particular reason.

leobalter:
https://github.com/jquery/qunit/compare/master...leobalter:Source

I've wrote the tests and at some point it was necessary to avoid them on unsupported browsers, the good point was to remove the source information on those browsers avoiding a "Source: undefined".

gauravmittal1995:
@jzaefferer @leobalter Added the commits for tests done by leobalter

leobalter:
This last commit from @gauravmittal1995 convinced me it's interesting to keep the test source info on failing tests:

captura de tela 2015-03-04 18 19 20

test/reporter-html.js
```diff
@@ -100,27 +68,60 @@ QUnit.module( "timing", {
...
+	}
+
+	assert.ok( /(^| )qunit-source( |$)/.test( source.className ), "Source element exists" );
+	assert.equal( source.firstChild.innerHTML, "Source: " );
+
+	// test/reporter-html.js is a direct reference to this test file
+	assert.ok( /\/test\/reporter-html\.js\:\d+/.test( source.innerHTML ),
+		"Source references to the current file and line number"
+	);
+});
```
leobalter: @gauravmittal1995, here are the tests for the source information that I was talking about.

gauravmittal1995: @leobalter Thanks a lot for your help on these tests :)

reporter/html.js
```diff
@@ -723,6 +723,20 @@ QUnit.testDone(function( details ) {
 		time.innerHTML = details.runtime + " ms";
 		testItem.insertBefore( time, assertList );
 	}
+
+	// Only add this information if stackTrace is supported
+	if ( Error().stack ) {
```
jzaefferer: Can't you check for `details.source`? This check has unnecessary overhead and relies too much on implementation details.

test/reporter-html.js
```diff
@@ -100,27 +68,60 @@ QUnit.module( "timing", {
...
+	assert.ok( parseInt( setup.innerHTML, 10 ) > 100,
 		"Runtime includes beforeEach"
 	);
 });
 
-})();
+QUnit.module( "source" );
+
+QUnit.test( "setup", function( assert ) {
+	assert.expect( 0 );
```
jzaefferer: Why 0?

test/reporter-html.js
```diff
@@ -100,27 +68,60 @@ QUnit.module( "timing", {
...
+});
+
+QUnit.test( "logs location", function( assert ) {
+	var source,
+		previous = document.getElementById( "qunit-test-output-" + assert.test.testId  )
+			.previousSibling;
+
+	source = previous.lastChild;
+
+	if ( !Error().stack ) {
```
jzaefferer: Same gotcha as above.

test/reporter-html.js
leobalter: same setup test on the [timing module](https://github.com/jquery/qunit/pull/758/files#r25956696) 

This test block is required in the next block.

jzaefferer: Oh, this is a test, not a `setup`. Misread that, nevermind.

test/reporter-html.js
leobalter: This test will happen only if the env has support for this implementation. If it doesn't, just assert no source info was inserted.

reporter/html.js
leobalter: details.source can be made by more kinds of implementation, I didn't want to bring the complexity from core's [extractStacktrace](https://github.com/jquery/qunit/blob/master/src/core.js#L474-L516) to test it, neither change that method causing a backwards incompatibility.

jzaefferer: There may be browsers that don't generate the `stack` property when calling just `Error()`, but will provide the property when catching an error using `try/catch`. We either have to take that into account and use a different check, or we have to verify that the behaviour isn't actually happening anywhere. In that case we could also simplify `sourceFromStacktrace`, removing the `try/catch`.

leobalter: simplifying sourceFromStacktrace would cause a backward compatibility issue on `details.source`. It would be better to do this for 2.0

gauravmittal1995:
@leobalter Any change needed? or should i squash the commits?

leobalter:
@gauravmittal1995, let the squash to happen only before merging. So far we need to wait @jzaefferer to answer my comments or you can try the approach he suggested, in a new commit in the PR.

If @jzaefferer agrees with my arguments, we're probably good to land, but we need to wait the next time he's available. Different time zones makes it take more time.

gauravmittal1995:
@leobalter I think i will wait for @jzaefferer to reply to your comments before making the changes he asked for

reporter/html.js
jzaefferer: So let's not change `sourceFromStacktrace` and fix this check, either by only looking for `details.source` (I think that's the correct way) or replicating the functionality of `sourceFromStacktrace` correctly. The current check will only output the source location of tests for a subset of browsers (as far as we know).

jzaefferer:
Discussed this with @leobalter in a private chat. He's going to do some tests to verify support for `Error().stack`, which should help making an informed decision here.

gauravmittal1995:
@jzaefferer ok ... thanks

jzaefferer:
@leobalter since #773 doesn't change much, what's the verdict for this PR?

leobalter:
I don't like to add tests based on the log to verify the feature. But we can replace that `if ( Error.stack() ) by something related to `details.source` if is defined in a testDone after the "setup" test (which would require to hook removal later).

Other way, without adding logging hooks: 

In the tests just copy the content from `sourceFromStacktrace` (without the return statement) for to replace the if condition in the tests.

jzaefferer:
> I don't like to add tests based on the log to verify the feature. But we can replace that if ( Error.stack() ) by something related to details.source if is defined in a testDone after the "setup" test (which would require to hook removal later).

I'm not sure what you're suggesting here. Can you please clarify, maybe using an example?

> In the tests just copy the content from sourceFromStacktrace (without the return statement) for to replace the if condition in the tests.

That seems fine, since all the complexity lives in the `extract...` method.

leobalter:
I didn't write an example before because it would be long and ugly, but as you agreed with the second option I would just stick with that.

jzaefferer:
:+1:

leobalter:
@gauravmittal1995, #773 is landed, let's finish this one!

leobalter:
I've got the branch and fixed this, but I got one remaining issue.

Safari 5.1 fails the tests as the stack is not defined because of [this line](https://github.com/jquery/qunit/blob/master/src/core.js#L516-L518)

I can't have the same detection on the test js file. There's nothing beyond the userAgent to avoid this.

leobalter:
https://github.com/jquery/qunit/compare/jquery:master...leobalter:gauravmittal1995-Source?expand=1

Here's the 2 commits fixing this PR. My branch is also rebased with master.

The [second one](640ef86) is not the best thing as it determines Safari 5.1 (the sourceURL exception) via userAgent. I can't find anything else to avoid this.

jzaefferer:
jQuery Core exports `$.support`, even though it mostly uses those flags internally. How about we export `QUnit.stacktraceSupportIsThere` (not my naming day) and use that in tests? Maybe we can use the flag internally as well, not sure.

leobalter:
`QUnit.stack` should be fine and would export a standardized Error stack. I'll push this in a separate PR so we can proper discuss it.

For now, we can land this as the flags are already internal (right?), and replace the tests when QUnit.stack is ready, or wait until the other PR to land.

jzaefferer:
Let's create a separate issue and merge this.

gauravmittal1995:
@leobalter Do i rebase this with https://github.com/jquery/qunit/compare/jquery:master...leobalter:gauravmittal1995-Source?expand=1 ???

PR #760, opened 2015-03-04
Assert: Fix missing test on exported throws

leobalter:
Fixes regression found at #752 and add tests for exported throws assertions.


jzaefferer:
:+1:

PR #763, opened 2015-03-05
HTML Reporter: Change display text for bad tests

### LINKED ISSUE #693: Make it more obvious why test execution order changes

shivamdixit:
When displaying text for running tests, if
QUnit.config.reorder is enabled then previously
failed tests will run first. However in summary
there is no indication for the same. This commit
changes the message text in HTML reporter.

Fixes #693

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -612,12 +612,10 @@ QUnit.testStart(function( details ) {
 
 	running = id( "qunit-testresult" );
 	if ( running ) {
-		if ( details.bad === true ) {
-			running.innerHTML = "Rerunning failed: <br />" +
+		running.innerHTML = details.wasBad ?
+			"Rerunning previously failed test: <br />" :
+			"Running: <br />" +
 				getNameHtml( details.name, details.module );
-		} else {
-			running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
-		}
 	}
 
 });
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -612,7 +612,12 @@ QUnit.testStart(function( details ) {
 
 	running = id( "qunit-testresult" );
 	if ( running ) {
-		running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
+		if ( details.bad === true ) {
+			running.innerHTML = "Rerunning failed: <br />" +
+				getNameHtml( details.name, details.module );
+		} else {
+			running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
+		}
 	}
 
 });
--- a/src/test.js
+++ b/src/test.js
@@ -80,7 +80,7 @@ Test.prototype = {
 			name: this.testName,
 			module: this.module.name,
 			testId: this.testId,
-			bad: this.wasBad
+			wasBad: this.wasBad
 		});
 
 		if ( !config.pollution ) {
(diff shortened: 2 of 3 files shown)


shivamdixit:
@jzaefferer Please review.

reporter/html.js
```diff
@@ -612,7 +612,12 @@ QUnit.testStart(function( details ) {
 
 	running = id( "qunit-testresult" );
 	if ( running ) {
-		running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
+		if ( details.bad === true ) {
+			running.innerHTML = "Rerunning failed: <br />" +
```
jzaefferer: Could be simplified to just "Rerunning:", or to be more explicit and correct, "Rerunning previously failed test:".

reporter/html.js
```diff
@@ -612,7 +612,12 @@ QUnit.testStart(function( details ) {
 
 	running = id( "qunit-testresult" );
 	if ( running ) {
-		running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
+		if ( details.bad === true ) {
```
jzaefferer: No need to compare a boolean to a boolean, just use `if ( details.bad )`

reporter/html.js
```diff
@@ -612,7 +612,12 @@ QUnit.testStart(function( details ) {
...
+			running.innerHTML = "Rerunning failed: <br />" +
+				getNameHtml( details.name, details.module );
+		} else {
+			running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
```
jzaefferer: This could be refactored to only change the string literal depending on the `details.bad` property. The `getNameHtml` call doesn't change.

src/test.js
```diff
@@ -78,7 +79,8 @@ Test.prototype = {
 		runLoggingCallbacks( "testStart", {
 			name: this.testName,
 			module: this.module.name,
-			testId: this.testId
+			testId: this.testId,
+			bad: this.wasBad
```
jzaefferer: `wasBad` is okay, but `bad` is a poor variable name. Might as well use `wasBad` for the logging callback.

shivamdixit:
@jzaefferer  Refactored the code as instructed.

jzaefferer:
Thanks for the update, looks good to me. I'll let someone else review this as well.

leobalter:
I don't like to have this feature at the logging callback API as it solves only a browser problem. Also, this `.wasBad` assumes you are rerunning a test because it failed before, and that might not be always truth.

Instead of messing with the API, I would use the browser `localStorage` API to handle it.

jzaefferer:
> Instead of messing with the API, I would use the browser localStorage API to handle it.

Can you be a bit more explicit? QUnit "core" uses `sessionStorage` to track failed tests, are you suggesting the HTML reporter should access that information as well?

leobalter:
`s/localStorage/sessionStorage`, that's what I meant.

>  are you suggesting the HTML reporter should access that information as well?

yes, or at least handle it by itself. It's better than adding a new API property

jzaefferer:
Okay, makes sense. @shivamdixit could you look into implementing that approach?

shivamdixit:
Sure, I'll try to implement using `sessionStorage`.

shivamdixit:
Created new PR #780

Issue #348, opened 2012-11-06
Confusing line breaks in object diff


rodneyrehm:
![confusing line breaks in object diff](http://rodneyrehm.de/t/github-issue-qunit-object-diff.png)

values should start at the same horizontal offset or be shown in sequence. Current visual is confusing, as it looks like the value could be a key or something.

AnishChandran:
http://fiddle.jshell.net/sj6kdj9g/ 
Is this an simple example about confusing line breaks in diff ? since the "gre": "foo" should be aligned with "foo": "bar ?

jzaefferer:
@AnishChandran that fiddle only has "foo"/"bar" as actual/expected - without linebreaks, there's nothing confusing there.

PR #764, opened 2015-03-05
HTML Reporter: Fixed the confusing Line Breaks in Diffs

gauravmittal1995:
Regarding https://github.com/jquery/qunit/issues/348
In the Below picture:
1st assert has same keys but different values,
2nd assert is for two strings,
3rd assert has different keys but similar values.

diff_pic

### DIFF
--- a/external/jsdiff/jsdiff.js
+++ b/external/jsdiff/jsdiff.js
@@ -115,6 +115,22 @@ QUnit.diff = (function() {
 			nSpace.push( " " );
 		}
 
+		for ( i = 0; i < oSpace.length; i++ ) {
+			if ( oSpace[i].length > 1 ) {
+				if ( ( oSpace[i][0] == "\n") && (oSpace[i][1] == " " ) ) {
+					oSpace[i] = "\n";
+				}
+			}
+		}
+
+		for ( i = 0; i < nSpace.length; i++ ) {
+			if ( nSpace[i].length > 1 ) {
+				if ( ( nSpace[i][0] == "\n") && (nSpace[i][1] == " " ) ) {
+					nSpace[i] = "\n";
+				}
+			}
+		}
+
 		if ( out.n.length === 0 ) {
 			for ( i = 0; i < out.o.length; i++ ) {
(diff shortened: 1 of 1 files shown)


gauravmittal1995:
@leobalter @jzaefferer Please Review

jzaefferer:
Can you provide a screenshot of those same diffs without the changes of this PR applied?

jzaefferer:
Also #348 had an example where an object was replaced by an array. Could you cover that as well?

gauravmittal1995:
@jzaefferer for the object replaced by array:

list_array

gauravmittal1995:
Now for without the change (i.e originally): 

pic1
pic2

gauravmittal1995:
@jzaefferer Hey, I also tried another way and got the following diff:

new_diff1
new_diff2

Please Let me know which one according to you is better.

jzaefferer:
Both approaches are definitely an improvement over the current implementation. The "another way" looks even better to me. We may need more tests to avoid regressions though.

@leobalter what do you think?

gauravmittal1995:
@jzaefferer the another way uses google's diff-patch-match library.

jzaefferer:
Ah, good to know. Could you create a separate PR for that?

leobalter:
I liked these changes, but I would like to see this other PR as it looks better.

gauravmittal1995:
@jzaefferer @leobalter Creating the new PR. Its taking time because have to refactor as it doesnt use cameCase and vars are scattered throughout the function. Will send it soon

leobalter:
As it is an external library, it would be good to ignore it on jshint and jscs checks. Also, getting the code from a npm dependency would be nice.

gauravmittal1995:
@leobalter @jzaefferer PR opened https://github.com/jquery/qunit/pull/772

gauravmittal1995:
@jzaefferer I think its safe to close this PR as we r mostly gonna stick with https://github.com/jquery/qunit/pull/772 ??

leobalter:
Yes, @gauravmittal1995, thanks!

Issue #765, opened 2015-03-06
Failed assertions throws an AssertionError 


leobalter:
based on http://wiki.commonjs.org/wiki/Unit_Testing/1.0, we could change QUnit to throw an `AssertionError` for failed assertions. 

```js
new assert.AssertionError({message: message, actual: actual, expected: expected})
assert.AssertionError instanceof Error
```

This way, we would not only be closer to a standard but we would be able write proper tests for failed assertions, which we don't have today.

The other side of doing this is not running the following assertions in the same test block. This might be considered as a bad thing, but regarding a test block should more specific, this is not that bad, IMHO.

leobalter:
The other way to test failing assertions would be mocking `QUnit.assert.push`, that would change any feature on QUnit.

The counter argument for this issue is that not throwing an exception on failed assertions is one of the things that make QUnit unique, but I still want to find something to test them properly.

jzaefferer:
CommonJS doesn't matter, so "compatibility with CommonJS" shouldn't be a priority for us.

As for the suggested change, its a common design in other test frameworks, which is a good argument for adopting it. From my point of view its also a user-hostile limitation that we should continue to ignore.

leobalter:
As the #767 is in the process, I'm already solving the final issue I wanted to solve with this before, which is something to test the failing assertions.

After some time rethinking about the exceptions, QUnit is indeed good to not do it, as it doesn't need to be like every other frameworks.

PR #767, opened 2015-03-06
Tests: Add missing tests for assertions

leobalter:
Also test failed assertions by mocking `QUnit.assert.push`

These changes were also good to fix assert.throws behaviour.

The other way to test failing assertions I have in mind is reported at #765, but I believe this is the best way to go.

Ref #765 (which might be closed after landing this patch).

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -165,11 +165,9 @@ QUnit.assert = Assert.prototype = {
 				expectedOutput = null;
 				ok = true;
 			}
-
-			currentTest.assert.push( ok, actual, expectedOutput, message );
-		} else {
-			currentTest.pushFailure( message, null, "No exception was thrown." );
 		}
+
+		currentTest.assert.push( ok, actual, expectedOutput, message );
 	}
 };
 
--- a/test/assert.js
+++ b/test/assert.js
@@ -1,5 +1,51 @@
 QUnit.module( "assert" );
 
+QUnit.test( "ok", function( assert ) {
+	assert.ok( true );
+	assert.ok( 1 );
+	assert.ok( "1" );
+	assert.ok( Infinity );
+	assert.ok( {} );
+	assert.ok( [] );
+});
+
+QUnit.test( "notOk", function( assert ) {
+	assert.notOk( false );
+	assert.notOk( 0 );
+	assert.notOk( "" );
+	assert.notOk( null );
+	assert.notOk( undefined );
+	assert.notOk( NaN );
+});
+
+QUnit.test( "equal", function( assert ) {
+	assert.equal( 1, 1 );
+	assert.equal( "foo", "foo" );
+	assert.equal( "foo", [ "foo" ] );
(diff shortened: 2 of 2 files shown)


leobalter:
@arschmitz already did something [really similar](https://gist.github.com/arschmitz/9bdcf8065c25ece5748b) on the failing assertions a few days ago. I didn't see that before and it's a crazy coincidence it happened so close.

It might be good to document this as a suggested pattern to test custom assertions.

arschmitz:
+1 for documenting this it took me quite a while to come up with this for my tests on qunit-assert-classes.

test/assert.js
```diff
@@ -237,3 +283,129 @@ QUnit.test( "raises, alias for throws", function( assert ) {
...
 	});
 });
+
+QUnit.module( "failing assertions", {
+	beforeEach: function( assert ) {
+		var originalPush = QUnit.assert.push;
+
+		this.originalPush = originalPush;
+
+		QUnit.assert.push = function( result, actual, expected, message ) {
```
jzaefferer: This should use `assert.push`, that way you're only modifying the current assert instance without requiring access to the global one. May remove the need for the `afterEach` below.

src/assert.js
```diff
@@ -165,11 +165,9 @@ QUnit.assert = Assert.prototype = {
 				expectedOutput = null;
 				ok = true;
 			}
-
-			currentTest.assert.push( ok, actual, expectedOutput, message );
-		} else {
-			currentTest.pushFailure( message, null, "No exception was thrown." );
 		}
+
+		currentTest.assert.push( ok, actual, expectedOutput, message );
```
jzaefferer: This change belongs in a separate commit. Currently its not at all mentioned in the commit message.

leobalter: Done

test/assert.js
```diff
@@ -237,3 +283,122 @@ QUnit.test( "raises, alias for throws", function( assert ) {
 		throw "my error";
 	});
 });
+
+QUnit.module( "failing assertions", {
+	beforeEach: function( assert ) {
+		assert.push = function( result, actual, expected, message ) {
+
+			// inverts the result so we can test failing assertions
+			QUnit.assert.push.call( assert, !result, actual, expected, message );
```
jzaefferer: Would be good to get rid of this global as well, instead saving a reference to the original `assert.push` and calling the original here.

leobalter: As the QUnit.assert is a reference for Assert's prototype and we can just call it with assert, I believe it's better than storing a variable in the memory, also it takes less loc.

jzaefferer: I'd like to get rid of all references to `QUnit.assert` outside of QUnit's source code, except those adding custom assertions.

In this case it should also remove the need to use `call()`.

leobalter:
done

jzaefferer:
Looks good to land!

leobalter:
Thanks!

Issue #554, opened 2014-04-26
Expose version in API and display in html reporter


Krinkle:
It would be helpful when working on a project to more easily identify what version of QUnit is being used (e.g through a property on the `QUnit` object).

While at it, maybe also display it somewhere in the html reporter (near where we dump the user agent).

shivamdixit:
@Krinkle @jzaefferer Please review the PR.

jzaefferer:
Btw. there's no need to ask people to review a PR just after you've sent it. Watchers of the repo get notifications anyway.

PR #768, opened 2015-03-08
Expose QUnit version as property and display it in HTML Reporter

shivamdixit:
Screenshot:

qunit

Ref #554

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -453,13 +453,14 @@ function storeFixture() {
 }
 
 function appendUserAgent() {
-	var userAgent = id( "qunit-userAgent" ),
-		versionInfo = "QUnit " + QUnit.version;
+	var userAgent = id( "qunit-userAgent" );
 
 	if ( userAgent ) {
 		userAgent.innerHTML = "";
 		userAgent.appendChild(
-			document.createTextNode( navigator.userAgent + ": " + versionInfo )
+			document.createTextNode(
+				navigator.userAgent + "; " + "QUnit " + QUnit.version
+			)
 		);
 	}
 }
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -28,6 +28,7 @@ QUnit.init = function() {
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
 			"<h2 id='qunit-banner'></h2>" +
 			"<div id='qunit-testrunner-toolbar'></div>" +
+			"<h2 id='qunit-version'></h2>" +
 			"<h2 id='qunit-userAgent'></h2>" +
 			"<ol id='qunit-tests'></ol>";
 	}
@@ -454,9 +455,20 @@ function storeFixture() {
 
 function appendUserAgent() {
 	var userAgent = id( "qunit-userAgent" );
+
 	if ( userAgent ) {
 		userAgent.innerHTML = "";
 		userAgent.appendChild( document.createTextNode( navigator.userAgent ) );
+	}
+}
+
+function appendVersion() {
+	var version = id( "qunit-version" ),
+		value = "@VERSION";
+
+	if ( version ) {
+		version.innerHTML = "";
+		version.appendChild( document.createTextNode( "QUnit " + value ) );
 	}
 }
 
@@ -518,6 +530,7 @@ QUnit.begin(function( details ) {
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
 			"<h2 id='qunit-banner'></h2>" +
 			"<div id='qunit-testrunner-toolbar'></div>" +
+			"<h2 id='qunit-version'></h2>" +
 			"<h2 id='qunit-userAgent'></h2>" +
 			"<ol id='qunit-tests'></ol>";
 	}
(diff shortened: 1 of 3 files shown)


jzaefferer:
Showing the QUnit version in a separate line takes up too much visual space. Maybe put it in front of the user-agent, separated by a semicolon?

shivamdixit:
@jzaefferer The length becomes too large if we display it in same line. Is it OK?

src/core.js
```diff
@@ -184,6 +185,10 @@ config.modules.push( config.currentModule );
 
 	// Figure out if we're running the tests from a server or not
 	QUnit.isLocal = location.protocol === "file:";
+
+	// Expose the current QUnit version
+	// `version` initialized at top of scope
+	QUnit.version = version;
```
jzaefferer: Since `version` isn't used anywhere else, it can be inlined, that is, use `"@VERSION"` here.

reporter/html.js
```diff
@@ -454,12 +455,23 @@ function storeFixture() {
...
+
 	if ( userAgent ) {
 		userAgent.innerHTML = "";
 		userAgent.appendChild( document.createTextNode( navigator.userAgent ) );
 	}
 }
 
+function appendVersion() {
+	var version = id( "qunit-version" ),
+		value = "@VERSION";
```
jzaefferer: Can this use `QUnit.version` instead?

jzaefferer:
> The length becomes too large if we display it in same line. Is it OK?

What do you consider "too large"? It looks like it would fit fine on your screenshot. Vertical screenspace is more important than horizontal.

shivamdixit:
Sure, I'll change it. (It looks little clumsy in my vertical monitor, for horizontal its perfectly fine)

shivamdixit:
Refactored the code. And changed the display text. Screenshot:

qunit

reporter/html.js
```diff
@@ -453,10 +453,14 @@ function storeFixture() {
...
 function appendUserAgent() {
-	var userAgent = id( "qunit-userAgent" );
+	var userAgent = id( "qunit-userAgent" ),
+		versionInfo = "QUnit " + QUnit.version;
+
 	if ( userAgent ) {
 		userAgent.innerHTML = "";
-		userAgent.appendChild( document.createTextNode( navigator.userAgent ) );
+		userAgent.appendChild(
+			document.createTextNode( navigator.userAgent + ": " + versionInfo )
```
jzaefferer: Let's replace the colon with a semicolon. Could also inline the `versionInfo` variable.

src/core.js
```diff
@@ -184,6 +184,10 @@ config.modules.push( config.currentModule );
 
 	// Figure out if we're running the tests from a server or not
 	QUnit.isLocal = location.protocol === "file:";
+
+	// Expose the current QUnit version
+	// `version` initialized at top of scope
```
jzaefferer: This comment is now outdated and can be removed

src/qunit.css
```diff
@@ -11,17 +11,17 @@
 
 /** Font Family and Sizes */
 
-#qunit-tests, #qunit-header, #qunit-banner, #qunit-testrunner-toolbar, #qunit-userAgent, #qunit-testresult {
+#qunit-tests, #qunit-header, #qunit-banner, #qunit-testrunner-toolbar, #qunit-version, #qunit-userAgent, #qunit-testresult {
```
jzaefferer: This looks like it can be reverted, since we're not creating a separate element anymore.

src/qunit.css
```diff
@@ -68,6 +68,13 @@
 	overflow: hidden;
 }
 
+#qunit-version {
```
jzaefferer: Same here, and for the other changes above.

jzaefferer:
Thanks, the update is an improvement. Just need to fix a few details.

shivamdixit:
I'm not sure why the build is failing. Is the browserstack test are failing because of the UI changes that I made?

jzaefferer:
Looks like the Travis build now passed. We run tests on browserstack, which sometimes fails with timeouts. Those are usually non-deterministic, which makes them difficult to debug and fix. Any persistent failures should be obvious enough though...

jzaefferer:
This PR now looks good to land to me. Will let @leobalter take a look as well.

leobalter:
Looks great, I have only one question: wouldn't it be better to show the QUnit version before the User Agent? It's a smaller text and would help locating every information on that line faster.

```js
"QUnit " + QUnit.version + " @ " + navigator.userAgent

jzaefferer:
I had considered the same, since you brought it up: Yes, let's switch the order.

shivamdixit:
Changed the order.
qunit

jzaefferer:
I missed this in Leo's comment: Let's stick with the semicolon as the separator. "@" works fine when it refers to the version of something, like "npm install package@version", but here the QUnit version and userAgent are independent. (bikeshedded this with @arthurvr on IRC as well)

shivamdixit:
@jzaefferer Okay. Fixed it.

jzaefferer:
Thanks! Merged this, squashing commits into one. I also made a tiny edit, removing the space in front of the semicolon.

Issue #664, opened 2014-09-13
Collapse failed tests in HTML reporter


mmun:
Often when I have many failing tests I like to collapse the failed tests with something like

```
$('.fail .qunit-assert-list').addClass('qunit-collapsed');
```

This let's me quickly scan and find the simplest failing tests that I should be working on first.

It would be great if this feature were officially added somewhere in the UI.

shivamdixit:
@leobalter Are you working on this issue?

jzaefferer:
Work in progresS: #769

nlfurniss:
@jzaefferer from https://github.com/qunitjs/qunit/issues/664#issuecomment-55515813: `That sounds like a usecase we can optimize for, though I'd prefer a solution that doesn't require an extra button or similar.`

I'd also love to be able to see the stacktraces for all failing tests; could you be persuaded this is a good idea?

jzaefferer:
@nlfurniss please create a new issue with a bit more context and background. A 4 year old quote in a closed ticket is not enough to kick anything off.

PR #769, opened 2015-03-08
HTML Reporter: Collapse failed tests in HTML reporter except first

gauravmittal1995:
Regarding https://github.com/jquery/qunit/issues/664
@leobalter @jzaefferer Please Review

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -64,7 +64,7 @@ if ( typeof window === "undefined" ) {
 
 var config = QUnit.config,
 	collapse = config.collapse,
-	collapse_check = collapse,
+	collapseCheck = collapse,
 	hasOwn = Object.prototype.hasOwnProperty,
 	defined = {
 		document: window.document !== undefined,
@@ -686,7 +686,7 @@ QUnit.log(function( details ) {
 
 QUnit.testDone(function( details ) {
 	var testTitle, time, testItem, assertList,
-		good, bad, testCounts, skipped, local_collapse,
+		good, bad, testCounts, skipped, localCollapse,
 		tests = id( "qunit-tests" );
 
 	if ( !tests ) {
@@ -709,10 +709,10 @@ QUnit.testDone(function( details ) {
 		}
 	}
 
-	local_collapse = config.collapse;
+	localCollapse = config.collapse;
 
-	if (local_collapse != collapse_check) {
-		collapse = local_collapse;
+	if (localCollapse !== collapseCheck) {
+		collapse = localCollapse;
 	}
 
 	if ( bad === 0 ) {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -63,6 +63,7 @@ if ( typeof window === "undefined" ) {
 }
 
 var config = QUnit.config,
+	collapse = config.collapse,
 	hasOwn = Object.prototype.hasOwnProperty,
 	defined = {
 		document: window.document !== undefined,
@@ -710,11 +711,11 @@ QUnit.testDone(function( details ) {
 	if ( bad === 0 ) {
 		addClass( assertList, "qunit-collapsed" );
 	} else {
-		if ( config.collapse === false ) {
-			config.collapse = -1;
-		} else if ( config.collapse === true ) {
-			config.collapse = 1;
-		} else if ( config.collapse === 1 ) {
+		if ( collapse === false ) {
+			collapse = -1;
+		} else if ( collapse === true ) {
+			collapse = 1;
+		} else if ( collapse === 1 ) {
 			addClass( assertList, "qunit-collapsed" );
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -100,6 +100,9 @@ config = {
 	// by default, scroll to top of the page when suite is done
 	scrolltop: true,
 
+	// By default, only expand the first failed test as opposed to all failed tests
+	collapse: true,
```
jzaefferer: Making this configurable seems sane, though based on the comment its unclear what `collapse: false` would do - collapse nothing? Collapse all tests?

reporter/html.js
```diff
@@ -697,6 +697,14 @@ QUnit.testDone(function( details ) {
 
 	if ( bad === 0 ) {
 		addClass( assertList, "qunit-collapsed" );
+	} else {
+		if ( config.collapse === false ) {
+			config.collapse = -1;
```
jzaefferer: We shouldn't change the configuration from the HTML reporter. If you need to keep track of state, use local state (variables) within the reporter.

gauravmittal1995:
@jzaefferer @leobalter Please Review

src/core.js
```diff
@@ -100,6 +100,10 @@ config = {
 	// by default, scroll to top of the page when suite is done
 	scrolltop: true,
 
+	// By default, only expand the first failed test
+	// When disabled (False), all the failed tests are expanded
```
jzaefferer: "When set to false, all failed tests are expanded"

reporter/html.js
```diff
@@ -63,6 +63,7 @@ if ( typeof window === "undefined" ) {
 }
 
 var config = QUnit.config,
+	collapse = config.collapse,
```
jzaefferer: This makes it impossible to actually change the behaviour via `QUnit.config`, since any changes to `QUnit.config.collapse` after `qunit.js` was loaded will be ignored. Tested by adding this to the top of `test/test.js`:

```js
QUnit.config.collapse = false;
```

Should be easy to fix by not making a copy of `config.collapse`, but still using a local variable to keep track of the need to collapse or not.

gauravmittal1995:
@jzaefferer Please Review. I think this solves the problem.

gauravmittal1995:
@leobalter Please Review

reporter/html.js
```diff
@@ -707,8 +709,22 @@ QUnit.testDone(function( details ) {
 		}
 	}
 
+	localCollapse = config.collapse;
+
+	if (localCollapse !== collapseCheck) {
+		collapse = localCollapse;
+	}
```
gauravmittal1995: Checks for change in QUnit.config

jzaefferer: This can be simplified. Don't make a copy of `config.collapse`, instead always reference that directly. Then track if the first test was collapsed with a simple boolean flag, eg. `collapsedFirst`.

gauravmittal1995:
Updated: As suggested

gauravmittal1995:
@jzaefferer Please Review

reporter/html.js
```diff
@@ -707,8 +710,18 @@ QUnit.testDone(function( details ) {
 		}
 	}
 
+	if (config.collapse !== collapseCheck) {
+		collapse = config.collapse;
+	}
```
leobalter: as already mentioned by Jörn, it's not necessary to make this variable copy. You can remove all this block and the `collape` declaration on the top of the file.

reporter/html.js
```diff
@@ -707,8 +710,18 @@ QUnit.testDone(function( details ) {
...
+
 	if ( bad === 0 ) {
 		addClass( assertList, "qunit-collapsed" );
+	} else {
+		if (( collapse === true ) && (collapseFirst === 0 )) {
```
leobalter: here you can use directly `config.collapse` instead of this new variable.

leobalter: also, there are missing spaces between the parenthesis. basically: replace `((` and `))` to `( (` and `) )`

reporter/html.js
```diff
@@ -707,8 +710,18 @@ QUnit.testDone(function( details ) {
...
+			collapseFirst = 1;
```
leobalter: as the class name used below, it would be good to rename this variable to `collapsed`. And you can use booleans.

doing so everything can be simplified:

```js
if ( config.collapse && collapseFirst ) {
```

src/core.js
```diff
@@ -100,6 +100,10 @@ config = {
 	// by default, scroll to top of the page when suite is done
 	scrolltop: true,
 
+	// By default, only expand the first failed test
+	// When set to false, all failed tests are expanded
```
leobalter: This description can be improved.

Maybe:

```js
// HTML Reporter: collapse every test except the first failing test
// If false, all failing tests will be expanded
```

jzaefferer:
@gauravmittal1995 are you interested in updating this once more, to address Leo's feedback?

leobalter:
I'll see if I can apply the fixes, unless @gauravmittal1995 wants to update it.

leobalter:
Closing this, work in progress on #867.

PR #770, opened 2015-03-09
Assert: changed object equivalence test to make it faster

BraulioVM:
It's certainly a minor change, but it helped my computer run the tests provided with qunit 0.3 seconds faster on average.

This "trick" is widely used to prove that two sets are equal, that is, if one set A is contained in B and both have the same size, then they must be equal. It's quite intuitive I guess. The advantage of using this approach instead of the old one is that we are able to change an O(n * log n) operation with a O(1) operation, which would be kind of a big deal if QUnit.equiv was used to compare large objects.

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -28,6 +28,24 @@ QUnit.equiv = (function() {
 			/* jshint camelcase: false, proto: true */
 			return obj.__proto__;
 		},
+
+		objectsHaveSameKeys = function( b, a ) {
+			var i;
+			for ( i in b ) {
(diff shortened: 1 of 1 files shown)


BraulioVM:
Ok, the tests doesn't seem to be working while they were in my computer. I'll look into what is happening here

BraulioVM:
Ok, I saw why the tests were failing and fixed the bug properly. However, the fix no longer is O(1) but O(n), and it's not that much faster.

jzaefferer:
Okay, thanks for the update. Was worth a try!

Issue #636, opened 2014-08-26
Should use `Error#stack` without throwing when available


JamesMGreene:
In Chrome, at least, a newly created `Error` instance already has its `stack` property set, so there is no need to `throw` it in that situation.  Moreover, `throw`ing it results in both a performance hit as well as debugging hell for any devs with the "Pause on all exceptions" setting enabled in Chrome's Dev Tools.

I am doing this in another open source library (polyfill) of mine... works well: [JamesMGreene/document.currentScript@dist/document.currentScript.js#L124-137](https://github.com/JamesMGreene/document.currentScript/blob/8c58df6f038caa6d12b6ec153c23e2dce8132758/dist/document.currentScript.js#L124-137)

Basics:

``` js
var e = new Error(),
    stack = e.stack;
if (!e.stack) {
  try {
    throw e;
  }
  catch (err) {
    stack = err.stack;
  }
}
// Use `stack`....
```

I've also used this technique in non-open source projects as well.

cc: @stefanpenner

PR #773, opened 2015-03-09
Core: Simplify stack trace methods

leobalter:
Ref #758 

I did a test on every supported browser and so far we can drop the `e.stacktrace` part and simplify the way we get the stack information.

The try catch blocks are necessary for IE10 and IE11, otherwise, we could get rid of them.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -483,20 +483,14 @@ function done() {
 	});
 }
 
-// Doesn't support IE6 to IE9
+// Doesn't support IE6 to IE9, it will return undefined on these browsers
 // See also https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error/Stack
 function extractStacktrace( e, offset ) {
 	offset = offset === undefined ? 4 : offset;
 
 	var stack, include, i;
 
-	if ( e.stacktrace ) {
-
-		// Opera 12.x
-		return e.stacktrace.split( "\n" )[ offset + 3 ];
-	} else if ( e.stack ) {
-
-		// Firefox, Chrome, Safari 6+, IE10+, PhantomJS and Node
+	if ( e.stack ) {
 		stack = e.stack.split( "\n" );
 		if ( /^error$/i.test( stack[ 0 ] ) ) {
 			stack.shift();
@@ -514,9 +508,10 @@ function extractStacktrace( e, offset ) {
 			}
 		}
 		return stack[ offset ];
+
+	// Support: Safari <=6 only
 	} else if ( e.sourceURL ) {
 
-		// Safari < 6
 		// exclude useless self-reference for generated Error objects
 		if ( /qunit.js$/.test( e.sourceURL ) ) {
 			return;
@@ -528,16 +523,19 @@ function extractStacktrace( e, offset ) {
 }
 
 function sourceFromStacktrace( offset ) {
-	var e = new Error();
-	if ( !e.stack ) {
+	var error = new Error();
+
+	// Support: Safari <=7 only, IE <=10 - 11 only
+	// Not all browsers generate the `stack` property for `new Error()`, see also #636
+	if ( !error.stack ) {
 		try {
-			throw e;
+			throw error;
 		} catch ( err ) {
-			// This should already be true in most browsers
-			e = err;
+			error = err;
 		}
 	}
-	return extractStacktrace( e, offset );
+
+	return extractStacktrace( error, offset );
 }
 
 function synchronize( callback, last ) {
--- a/browserstack.json
+++ b/browserstack.json
@@ -10,18 +10,33 @@
 	],
 	"browsers": [
 		"chrome_previous",
-		"chrome_latest",
+		"chrome_current",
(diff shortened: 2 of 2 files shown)


scottgonzalez:
You should update this to use proper support comments (see https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197).

src/core.js
```diff
@@ -516,16 +512,11 @@ function extractStacktrace( e, offset ) {
 }
 
 function sourceFromStacktrace( offset ) {
-	var e = new Error();
-	if ( !e.stack ) {
```
jzaefferer: Please see #636 for why this should be kept.

leobalter: Thanks

src/core.js
```diff
@@ -516,16 +511,17 @@ function extractStacktrace( e, offset ) {
...
+			throw error;
 		} catch ( err ) {
-			// This should already be true in most browsers
-			e = err;
+			error = err;
 		}
 	}
-	return extractStacktrace( e, offset );
+
+	return extractStacktrace( error, offset );
```
leobalter: @jzaefferer is that ok to at least bring these changes? Also, the removed comment is useless.

src/core.js
```diff
@@ -478,13 +478,7 @@ function extractStacktrace( e, offset ) {
 
 	var stack, include, i;
 
-	if ( e.stacktrace ) {
-
-		// Opera 12.x
-		return e.stacktrace.split( "\n" )[ offset + 3 ];
-	} else if ( e.stack ) {
-
-		// Firefox, Chrome, Safari 6+, IE10+, PhantomJS and Node
```
leobalter: @scottgonzalez: here we basically support everything but IE6 to 9, support for Safari 6 and 7 is at the next branch.

I wasn't able to identify how to mention it "Doesn't support IE6 to IE9", as line [477](https://github.com/jquery/qunit/pull/773/files#diff-7eb52b366866677666470e019283c8eaR477)

scottgonzalez: You wouldn't mark that with a support comment; the existing comment seems fine. However, it's not at all clear what happens in the unsupported browsers. That should be documented in a comment.

src/core.js
scottgonzalez: When a reference like this is needed in a PR, it's needed in the code as well.

leobalter: agreed. I'll fix it

src/core.js
leobalter: done, thanks

src/core.js
```diff
@@ -516,16 +511,18 @@ function extractStacktrace( e, offset ) {
 }
 
 function sourceFromStacktrace( offset ) {
-	var e = new Error();
-	if ( !e.stack ) {
+	var error = new Error();
+
+	// Ref https://github.com/jquery/qunit/issues/636
```
jzaefferer: How about this?

```js
// Not all browsers generate the `stack` property for `new Error()`, see also #636
```

Though that makes me wonder which browsers those are - can we list them?

src/core.js
jzaefferer: Yeah, remaining changes look good

jzaefferer:
Apart from the one comment, this looks good.

src/core.js
jzaefferer: Just noticed the detail in the PR description. Let's make that into another support comment.

leobalter: good catch:

- Safari 7 (safari_previous)
- IE10
- IE11

If we drop support for Safari 6 (current is 8), the `.sourceURL` can also be dropped.

jzaefferer:
Since the previous discussion is now hidden:

Our `browserstack.json` doesn't match our actual browser support, which is supposed to match [jQuery 1.x](http://jquery.com/browser-support/). Our website links to that page, then claims "That's IE6+ and Current - 1 for Chrome, Firefox, Safari and Opera.", which isn't correct.

I think we should just add the missing browsers to `browserstack.json`, fix the website, and hopefully @leobalter can extend his `stack` test to those missing browsers.

leobalter:
I made the changes to include the following browsers on the browserstack list:

- Safari 5.1 and 6
- iOS Safari 6 to 8
- Android (whatever browserstack uses as the default browser), on versions 2.3, 4.1, 4.2, 4.3, 4.4 and 5
- Opera 12.10

The tests are now failing on Opera 12.10 due to the promise feature. 

ATM, we need to move forward on the QUnit 2.0 to drop this compatibility.

leobalter:
All Android tests returns a timeout, even when the max timeout value is set.

Should I drop it?

leobalter:
I removed the Android tests, otherwise we need Browserstack to fix the timeout issue.

leobalter:
ok to merge this without the Android tests on browserstack? #785 depends on that.

jzaefferer:
Yeah, let's land this. We'll see if the browserstack-runner performance is gonna get worse with more workers. We definitely need to do something about those timeouts.

PR #776, opened 2015-03-11
Prop equal against circular structures

### LINKED ISSUE #411: assert.propEqual not guarded against circular structures
As a result of #343 we got a new "propEqual" method which is usefully capable of comparing object trees in a "constructor-blind" manner, using a weaker semantic than QUnit.deepEqual. However, this method is not up to the standard of the rest of the framework - in particular, QUnit.deepEqual itself offers support for
1. correctly comparing cyclic structures without bombing the stack (#100 - recently improved  by #397)
2. not being confused by primitive arguments such as strings - these are compared properly and highlighted in the difference output.

"propEqual" was implemented using a utility function "objectValues" which performs a clone of the arguments before they are dispatched to QUnit.deepEqual, and it is deficiencies in this algorithm which are responsible for deficiencies in propEqual as compared to deepEqual.
1. is caused by the cloning operation itself bombing out on cyclic structures, and
2. is caused by the lack of a check for primitive arguments before copying - strings are copied as if they were arrays. The comments at the head of objectValues are clear about the latter restriction, but this contract limitation is not clear when passed up to propEqual itself.

I enclose a screenshot of the results of comparing two string using propEqual. deepEqual by contrast shows a suitable result.
propEqualStrings

BraulioVM:
Made this pull request to fix issue #411 

The problem within this issue was that propEqual was using the utility function objectValues, which does not check against circular structures, and then compared the result of both objectValues with a normal QUnit.equiv. As this approach does not work, I decided to split QUnit.equiv into  two separate functions:
- `QUnit.equiv.deep`: which performs the old QUnit.equiv computations
- `QUnit.equiv.props`: which compares object properties, ignoring their constructor. This function does check against circular structures and allows propEqual to work as it should. If this function is passed a non-object parameter, a `QUnit.equiv.deep` will be called.

I also created some tests for this new function (mostly copied them from deepEqual.js tests).

Fixes #411

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -60,7 +60,8 @@ grunt.initConfig({
 		},
 		all: [
 			"<%= jshint.all %>",
-			"!test/deepEqual.js"
+			"!test/deepEqual.js",
+			"!test/propEqual.js"
 		]
 	},
 	search: {
--- a/src/assert.js
+++ b/src/assert.js
@@ -87,23 +87,19 @@ QUnit.assert = Assert.prototype = {
 	},
 
 	propEqual: function( actual, expected, message ) {
-		actual = objectValues( actual );
-		expected = objectValues( expected );
-		this.push( QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( QUnit.equiv.props( actual, expected ), actual, expected, message );
 	},
 
 	notPropEqual: function( actual, expected, message ) {
-		actual = objectValues( actual );
-		expected = objectValues( expected );
-		this.push( !QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( !QUnit.equiv.props( actual, expected ), actual, expected, message );
 	},
 
 	deepEqual: function( actual, expected, message ) {
-		this.push( QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( QUnit.equiv.deep( actual, expected ), actual, expected, message );
 	},
 
 	notDeepEqual: function( actual, expected, message ) {
-		this.push( !QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( !QUnit.equiv.deep( actual, expected ), actual, expected, message );
 	},
 
 	strictEqual: function( actual, expected, message ) {
--- a/src/assert.js
+++ b/src/assert.js
@@ -165,11 +165,9 @@ QUnit.assert = Assert.prototype = {
 				expectedOutput = null;
 				ok = true;
 			}
-
-			currentTest.assert.push( ok, actual, expectedOutput, message );
-		} else {
-			currentTest.pushFailure( message, null, "No exception was thrown." );
 		}
+
+		currentTest.assert.push( ok, actual, expectedOutput, message );
 	}
 };
 
--- a/src/assert.js
+++ b/src/assert.js
@@ -117,7 +117,8 @@ QUnit.assert = Assert.prototype = {
 	"throws": function( block, expected, message ) {
 		var actual, expectedType,
 			expectedOutput = expected,
-			ok = false;
+			ok = false,
+			currentTest = ( this instanceof Assert && this.test ) || QUnit.config.current;
 
 		// 'expected' is optional unless doing string comparison
 		if ( message == null && typeof expected === "string" ) {
@@ -125,13 +126,13 @@ QUnit.assert = Assert.prototype = {
 			expected = null;
 		}
 
-		this.test.ignoreGlobalErrors = true;
+		currentTest.ignoreGlobalErrors = true;
 		try {
-			block.call( this.test.testEnvironment );
+			block.call( currentTest.testEnvironment );
 		} catch (e) {
 			actual = e;
 		}
-		this.test.ignoreGlobalErrors = false;
+		currentTest.ignoreGlobalErrors = false;
 
 		if ( actual ) {
 			expectedType = QUnit.objectType( expected );
@@ -165,9 +166,9 @@ QUnit.assert = Assert.prototype = {
 				ok = true;
 			}
 
-			this.push( ok, actual, expectedOutput, message );
+			currentTest.assert.push( ok, actual, expectedOutput, message );
 		} else {
-			this.test.pushFailure( message, null, "No exception was thrown." );
+			currentTest.pushFailure( message, null, "No exception was thrown." );
 		}
 	}
 };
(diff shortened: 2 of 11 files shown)


BraulioVM:
I have signed the CLA several times now, but don't know why it doesn't seem so to jquerybot

Gruntfile.js
```diff
@@ -60,7 +60,8 @@ grunt.initConfig({
 		},
 		all: [
 			"<%= jshint.all %>",
-			"!test/deepEqual.js"
+			"!test/deepEqual.js",
+			"!test/propEqual.js"
```
jzaefferer: Let's not add more exceptions. We should fix any issues in new files.

src/assert.js
```diff
@@ -87,23 +87,19 @@ QUnit.assert = Assert.prototype = {
...
 	},
 
 	deepEqual: function( actual, expected, message ) {
-		this.push( QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( QUnit.equiv.deep( actual, expected ), actual, expected, message );
 	},
 
 	notDeepEqual: function( actual, expected, message ) {
-		this.push( !QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( !QUnit.equiv.deep( actual, expected ), actual, expected, message );
```
jzaefferer: Let's keep `QUnit.equiv` backwards-compatible. You can still add the `props` property to `QUnit.equiv`. Should remove the need to update the `test/deepEqual.js` tests as well.

Gruntfile.js
BraulioVM: I added test/propEqual.js to the exceptions because the tests in both files are almost the same

src/core.js
```diff
@@ -74,6 +74,24 @@ var QUnit,
...
 		}
 		return vals;
+	},
+	/**
+	 * Gets the keys of an object which are both enumerable and own
+	 *
+	 * @param {Object} obj
+	 * @return {Array} Array with the enumerable and own keys of the object given
+	 */
+	ownObjectKeys = function( obj ) {
```
jzaefferer: This is used only in `src/equiv.js`. Let's add it there. We want to reduce circular dependencies between QUnit modules, should add even more.

src/assert.js
```diff
@@ -87,23 +87,19 @@ QUnit.assert = Assert.prototype = {
 	},
 
 	propEqual: function( actual, expected, message ) {
-		actual = objectValues( actual );
-		expected = objectValues( expected );
-		this.push( QUnit.equiv( actual, expected ), actual, expected, message );
+		this.push( QUnit.equiv.props( actual, expected ), actual, expected, message );
```
jzaefferer: Since this doesn't use `objectValues` anymore, and it isn't used anywhere else, we should remove that function.

jzaefferer:
It looks like `propEqual` now replicates a lot of the `deepEqual` implementation. Maybe we're on the wrong track here. Maybe we should drop `propEqual` and remove the prototype check in `deepEqual`. That's what the io.js project went for their `assert` module: https://github.com/iojs/io.js/pull/621

Gruntfile.js
jzaefferer: Right, but we shouldn't duplicate crappy code.

jzaefferer:
Not sure why the CLA check is failing, as far as I can tell your signature is fine. You can ignore that, I've pinged someone about that.

BraulioVM:
As you said, I already removed ````objectValues```` and put ````ownObjectKeys```` in equiv.js. I also made ````QUnit.equiv```` backwards compatible. I didn't fix the formatting issues in test/propEqual.js as that could take a really long long time.

Talking about https://github.com/jquery/qunit/pull/776#issuecomment-78273500, I think that when comparing objects, what most people actually want is comparing like propEqual does, but can't really know for sure.

src/equiv.js
```diff
@@ -212,8 +236,75 @@ QUnit.equiv = (function() {
...
+				callers.pop(); // unstack, we are done
+
+				// ensure both objects have the same properties
+				return eq && deepEquiv(aProperties.sort(), bProperties.sort());
+			}
+		}( args[ 0 ], args[ 1 ] ) ) &&
+			propEquiv.apply(this, args.splice( 1, args.length - 1 ) ) );
+
+	},
+	innerEquiv = deepEquiv; // default equivalence check
```
jzaefferer: Might as well drop `innerEquiv` and assign `deepEquiv.props` directly, then return that.

test/deepEqual.js
```diff
@@ -1,4 +1,4 @@
-QUnit.module( "equiv" );
+QUnit.module( "deepEquiv" );
```
jzaefferer: Let's revert this - there is no exported method called "deepEquiv".

test/propEqual.js
```diff
@@ -0,0 +1,708 @@
+QUnit.module( "propEquiv" );
```
jzaefferer: "QUnit.equiv.props"

jzaefferer:
@Krinkle @JamesMGreene @leobalter any input on the discussion about general changes to `deepEqual` (or `propEqual`)?

src/equiv.js
BraulioVM: I'm not sure that would totally make sense. How could ````propEquiv```` be a member of ````deepEquiv````? They are kind of opposite. Something that could possibly be more natural would be to drop ````deepEquiv```` and assign ````innerEquiv```` and ````innerEquiv.props```` directly. The point of this is not having that "linguistic contradiction" with ````propEquiv```` being a subproperty of ````deepEquiv````.

Does that make any sense?

test/propEqual.js
BraulioVM: Would "equiv.props" be fine? It would be the same pattern used in test/deepEqual.js

src/equiv.js
jzaefferer: Sticking with `innerEquiv` is fine.

leobalter:
I'm sorry I didn't give the proper attention to this PR, should we continue and update?

jzaefferer:
I still think this is a valuable PR and we should try to pick it up and finish it. I just looked over the current stage again and while there's probably some minor details to fix (I noticed a missing end-of-file newline, so there are probably more style issues), overall this should be good to land.

jzaefferer:
Before we put more time into this though, we should figure out what to do about #865. If we contribute our implementation to an external module and make use of that, we can (as easily) share code between `propEqual` and `deepEqual` anymore.

leobalter:
does anyone want to continue the work on this? Should we close this at this point?

trentmwillis:
I think this is still worth pursuing, but unless @BraulioVM wants to take this up again, it'd be best to close and open a new PR.

trentmwillis:
It's been a couple more months, so I'm going to close this and hopefully we can address it in another PR.

PR #780, opened 2015-03-16
HTML Reporter: Change display text for bad tests

### LINKED ISSUE #693: Make it more obvious why test execution order changes

shivamdixit:
When displaying text for running tests, if
QUnit.config.reorder is enabled then previously
failed tests will run first. However in summary
there is no indication for the same. This commit
changes the message text in HTML reporter.

Fixes #693

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -611,7 +611,7 @@ function getNameHtml( name, module ) {
 }
 
 QUnit.testStart(function( details ) {
-	var running, testBlock;
+	var running, testBlock, bad;
 
 	testBlock = id( "qunit-test-output-" + details.testId );
 	if ( testBlock ) {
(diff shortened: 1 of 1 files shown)


reporter/html.js
```diff
@@ -624,7 +624,13 @@ QUnit.testStart(function( details ) {
 
 	running = id( "qunit-testresult" );
 	if ( running ) {
-		running.innerHTML = "Running: <br />" + getNameHtml( details.name, details.module );
+		bad = QUnit.config.reorder && defined.sessionStorage &&
+			+sessionStorage.getItem( "qunit-test-" + details.module + "-" + details.name );
+
+		running.innerHTML = bad ?
+			"Rerunning previously failed test: <br />" :
```
jzaefferer: This needs parens. Currently the test name is only added to "Running", not to the new string.

Btw. I'm testing this by making an async test fail (change `expect()` call), then increasing the timeout to 2 seconds.

jzaefferer:
This looks much better. I didn't realize before that the HTML reporter already does most of the `sessionStorage` access anyway.

Btw. if you haven't already, you should start writing your proposal! See also https://github.com/jquery/gsoc/issues/1

shivamdixit:
I've added the parenthesis and squashed commits into one. Yeah, I've already started with my proposal :smile:

jzaefferer:
Thanks! Merged, with only a small update to the commit message, adding the reference to the PR and using the full 80 character width for the description.

PR #783, opened 2015-03-16
Typed arrays comparison with standard arrays

### LINKED ISSUE #672: propEqual for TypedArrays (was deepEqual for TypeArrays)
I've notice that qunit gives incorrect results when using typedArrays (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays).
I think it will be a great improvement if it will let you compare standard array and typedarrays like for example:

``` javascript
var ta = new Uint8Array(4);
deepEqual(ta, [0, 0, 0, 0], "Typed array and array equals");
```

I was checking the code and I guess one solution could be to change `core.js`, in the line when you know that the typeof is an `object`, and do something like:

``` javascript
...
if ( typeof obj === "object" ) {
    if ( obj instanceof Int8Array ||
        obj instanceof Uint8Array ||
        obj instanceof Uint8ClampedArray ||
        obj instanceof Int16Array ||
        obj instanceof Uint16Array ||
        obj instanceof Int32Array ||
        obj instanceof Uint32Array ||
        obj instanceof Float32Array ||
        obj instanceof Float64Array ) {
        return "typedarray";
    }
    return "object";
}
```

And then on `equiv.js` returns a new function for the new type.

shivamdixit:
Allow `deepEqual` comparison of typed arrays and standard arrays. Since phantomjs don't support `Uint8ClampedArray` and `Float64Array` hence they are not present in the comparison.

Fixes #672

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -17,6 +17,7 @@ var QUnit,
 	defined = {
 		document: window.document !== undefined,
 		setTimeout: window.setTimeout !== undefined,
+		typedArray: typeof( Int8Array ) === "function",
 		sessionStorage: (function() {
 			var x = "qunit-test-string";
 			try {
@@ -348,13 +349,13 @@ extend( QUnit, {
 				return type.toLowerCase();
 		}
 		if ( typeof obj === "object" ) {
-			if ( obj instanceof Int8Array ||
-				obj instanceof Uint8Array ||
-				obj instanceof Int16Array ||
-				obj instanceof Uint16Array ||
-				obj instanceof Int32Array ||
-				obj instanceof Uint32Array ||
-				obj instanceof Float32Array ) {
+			if ( defined.typedArray && obj instanceof Int8Array ||
+					obj instanceof Uint8Array ||
+					obj instanceof Int16Array ||
+					obj instanceof Uint16Array ||
+					obj instanceof Int32Array ||
+					obj instanceof Uint32Array ||
+					obj instanceof Float32Array ) {
 				return "typedarray";
 			}
 
--- a/src/core.js
+++ b/src/core.js
@@ -348,6 +348,16 @@ extend( QUnit, {
 				return type.toLowerCase();
 		}
 		if ( typeof obj === "object" ) {
+			if ( obj instanceof Int8Array ||
+				obj instanceof Uint8Array ||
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -348,6 +348,16 @@ extend( QUnit, {
 				return type.toLowerCase();
 		}
 		if ( typeof obj === "object" ) {
+			if ( obj instanceof Int8Array ||
```
jzaefferer: Looking at the failing tests in IE, you need a feature detect before referencing any of these. Also applies to the unit tests.

src/core.js
```diff
@@ -348,6 +348,16 @@ extend( QUnit, {
...
+				obj instanceof Uint8Array ||
```
jzaefferer: Multi-line if-statements get extra indent, so all of these `instanceof` lines except the first need one more tab.

jzaefferer:
A few issues that should be easy to fix. Also a conceptual issue that I commented on in #672 - would appreciate your input there as well.

jzaefferer:
@shivamdixit: mostly off-topic: When are you going to submit your proposal through the GSoc site?

shivamdixit:
@jzaefferer will submit it today itself, just finished with exams.

shivamdixit:
Why Travis-CI build is not running?

shivamdixit:
One question -  Is it safe to assume that if `Int8Array` exists, then others exists as well?

jzaefferer:
Considering the discussion in #672, can you create a 2nd PR that updates `propEqual` instead? If that turns out to be a terrible idea, we can come back to this.

PS: You can do `git commit --allow-empty` or amend an existing commit to trigger a new Travis build.

shivamdixit:
Hi @jzaefferer, my exams are going on, I'll submit the updated patches soon.

jzaefferer:
@shivamdixit are you still interested in doing the PR for `propEqual`?

shivamdixit:
@jzaefferer Yes definitely. Just finished up with my exams, will start contributing again from Tuesday.

leobalter:
@shivamdixit I'm closing this due to inactivity. We can reopen if you want to continue working on it.

PR #784, opened 2015-03-16
Optimizations on test folder

### LINKED ISSUE #781: Refactor test/ folder
By now we have too many files in a single directory and its getting hard to find specific files when I don't know the exact name. We should refactor the `test/` folder.

I suggest moving files referenced from `index.html` into a `default/` folder (or some other name). The remaining files should be `.html` files with their matching `.js` test file. We can then see from there if we need other changes.

Should probably land some more PRs first, at least those making changes to tests.

leobalter:
Closes #781


Gruntfile.js
```diff
@@ -92,13 +92,13 @@ grunt.initConfig({
 			}
 		},
 		qunit: [
-			"test/index.html",
```
jzaefferer: Can you move this file back? I very much like to stick with opening `qunit/test` to have tests run...

jzaefferer:
Apart from keeping `test/index.html` in place, this looks good.

leobalter:
@jzaefferer good to land? Should I squash them? different changes on each one.

leobalter:
ping

Gruntfile.js
```diff
@@ -60,7 +60,7 @@ grunt.initConfig({
 		},
 		all: [
 			"<%= jshint.all %>",
-			"!test/deepEqual.js"
+			"!test/default/deepEqual.js"
```
jzaefferer: I feel like there's a better name for this, but can't come up with anything I like. `index` would match `index.html`, but doesn't make much sense otherwise. `main` or `core` isn't great either.

jzaefferer:
Looks good. If we can't come up with a better name, let's land this as-is (separate commits).

leobalter:
`main` looks good, it tells those are the main tests for the core functionality. As the other tests are for the supportive roles.

jzaefferer:
Let's go with `main` then.

leobalter:
rebased.

jzaefferer:
Looks good. I hope we're not wreaking too much havoc on open PRs by landing this.

leobalter:
the files are flagged as `moved`, I hope git manage that in a friendly way.

PR #786, opened 2015-03-17
HTML Reporter: Show diff only when it helps

### LINKED ISSUE #335: Show diff only when it helps
If the length of the diff is just the length of `expected.length + actual.length`, its likely to be useless. Its only useful if it can reduce the input.

shivamdixit:
Don't show diff if expected and actual values are completely different and there is nothing in common.

However when working with two different arrays or objects, the braces remains common because of which the length changes and diff is displayed. Is this behavior accepted? It works fine if we compare object and array. Screenshot attached:

qunit

Fixes #335

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -629,9 +629,15 @@ QUnit.testStart(function( details ) {
 
 });
 
+function stripHtml( string ) {
+	// strip tags, html entity and whitespaces
+	return string.replace(/<\/?[^>]+(>|$)/g, "").replace(/\&quot;/g, "").replace(/\s+/g, "");
+}
+
 QUnit.log(function( details ) {
 	var assertList, assertLi,
-		message, expected, actual,
+		message, expected, actual, diff,
+		lengthChanged = true,
 		testItem = id( "qunit-test-output-" + details.testId );
 
 	if ( !testItem ) {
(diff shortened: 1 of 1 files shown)


jzaefferer:
This is also something I'd like to evaluate after we've landed a new diff implementation (#772), since that is very likely to affect this optimization.

shivamdixit:
Output with new diff library (Google diff patch match):

qunit6

I think this PR is still valid. In case of completely different output no diff is displayed. However even if one character is same, diff is displayed.

What do you think @jzaefferer ?

leobalter:
There's only one thing missing here, maybe introduced by the recent changes: we need to check and ommit the diff when one of the values (actual or expected) is a boolean. 

The diff: falstrue is awful.

shivamdixit:
@leobalter Yes, I agree. Showing diff for boolean don't make much sense.

leobalter:
https://github.com/jquery/qunit/compare/master...leobalter:shivamdixit-335-show-diff-only-when-it-helps?expand=1

I've got that branch rebased and did the other commit checking the boolean value.

shivamdixit:
Looks good to me. Created a new PR. Ref https://github.com/jquery/qunit/pull/802

leobalter:
You could have updated your branch, but let's move on the new PR now.

PR #790, opened 2015-03-19
Core: Implement a node stdout reporter

leobalter:
With this reporter we can use QUnit directly on npm environments:

```js
var QUnit = require( "../dist/QUnit" );

QUnit.reporter();

require( "./unit-tests.js" );

QUnit.load();
```

The stdout reporter is only activated when called, this way we keep the retrocompatibility.

Log example: https://www.dropbox.com/s/07k3funzdwp0y7r/Screenshot%202015-03-19%2016.18.25.png?dl=0

The reporter function also accepts an object with the options, the only current available options is `output`:

```js
// Will print a minimal report, with only dots for each assertion,
// some extra information comes on failures to help locate the issue
QUnit.reporter({
  output: "minimal"
});
```

```js
// Will print a more detailed report, listing each passing assertion.
QUnit.reporter({
  output: "verbose"
});
```


scottgonzalez:
Is the use of chalk the only thing specific to node? You might consider changing this from a node stdout reporter to a console reporter.

leobalter:
The stdout reporter is also using Node's `process.stdout` and `process.error`. AFAIK, the use of console.log is limited to one line for each log, which is not what I'm aiming for here.

Maybe, it will be more clear to replace the `QUnit.reporter` for `QUnit.stdout`.

My goal in this PR is to provide a full operational reporter for Node, so we may use QUnit without any adapter.

leobalter:
The stdout reporter may be followed by a TAP reporter. The one at [Mocha](https://github.com/mochajs/mocha/blob/master/lib/reporters/tap.js) may be used as reference.

I'll take another look to see if I detach this code from the build, and make it an individual module, so we can set it as an separate module and just `require` the module when `QUnit.stdout` is called.

leobalter:
The last commit changed this PR to use the stdout reporter feature from an external module. We can turn the module a jQuery project, as you prefer.

package.json
```diff
@@ -31,7 +31,9 @@
     "qunit/qunit.css",
     "LICENSE.txt"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "chalk": "1.0.0"
```
jzaefferer: Is this still needed?

package.json
```diff
@@ -45,6 +47,7 @@
     "grunt-qunit-istanbul": "0.4.5",
     "grunt-search": "0.1.6",
     "load-grunt-tasks": "0.3.0",
+    "qunit-reporter-stdout": "0.1.0",
```
jzaefferer: If this is only a devDependency, how does `QUnit.stdout()` actually add any value? Is all we're removing from userland a single require-call?

leobalter: my bad here, it needs to be a dependency, not for development.

package.json
leobalter: removing it.

test/reporter-stdout.js
```diff
@@ -0,0 +1,21 @@
+/*globals QUnit:true*/
+/*jshint node:true*/
+var QUnit = require( "../dist/qunit" );
+
+// Options: { output: "minimal" || "verbose" }
+QUnit.stdout();
+
+// Load QUnit tests
+require( "./logs" );
```
jzaefferer: How do these tests get access to `QUnit`? We must be assigning `global.QUnit` somewhere for that to work, right?

leobalter: It will use the QUnit variable assigned above. This current file `test/reporter-stdout.js` will work as a context module for its following requirements.

jzaefferer: Required files in node don't have access to the scope of the file that is requiring it. So the only way for the required file to have `QUnit` in scope is for a global (right?). Where does the global come from?

scottgonzalez: See line 3 above.

jzaefferer: That `var QUnit` shouldn't be accessible in the scope of any other files required here. Its a local variable, not made global in any way.

leobalter: This is the same thing already happening on Grunt [`test-on-node` task](https://github.com/jquery/qunit/blob/master/Gruntfile.js#L178).

jzaefferer: Sure, I understand *that* it works, I don't understand *how*.

leobalter: It's certainly a not documented feature, and because of this it might be a bad practice. I'll insert an extra line exporting QUnit to the global scope.

leobalter: done via @9bea625

jzaefferer: That's okay, but still doesn't explain why it was working before.

leobalter:
bumped the reporter module to 0.2.0 where I included output tests. (the minor version was due to some differences on the stderr output on failing tests)

test/reporter-stdout.js
leobalter: as I told before, this looks like an undocumented feature. Node modules seems to be leaking the current file scope to their following required modules.

scottgonzalez: Are you sure about that? QUnit does some crazy stuff with globals and different environments.

leobalter: No, I'm not. I'll check this out on an example repo.

jzaefferer: Looks like the `this` from outro.js refers in node to `global`, which we then alias as `window`, and in exports.js we then assign `window.QUnit`, which in node is the same as `global.QUnit`. Let's fix that.

leobalter:
By several arguments, I'm closing this PR.

We are on the way to have a release of js-reporters, as many improvement ideas came after this PR and now rebasing this is not easy.

PR #791, opened 2015-03-20
HTML Reporter: More graceful handling of AMD

scottgonzalez:
Still needs tests, but @jzaefferer and I wanted to get this in front of more eyes in case there's something horribly wrong with this approach.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -756,13 +756,15 @@ QUnit.testDone(function( details ) {
 	}
 });
 
-if ( !defined.document || document.readyState === "complete" ) {
+if ( defined.document ) {
+	if ( document.readyState === "complete" ) {
+		QUnit.load();
+	} else {
+		addEvent( window, "load", QUnit.load );
+	}
+} else {
 	config.pageLoaded = true;
 	config.autorun = true;
-}
-
-if ( defined.document ) {
-	addEvent( window, "load", QUnit.load );
 }
 
 })();
--- a/src/export.js
+++ b/src/export.js
@@ -61,3 +61,10 @@ if ( typeof module !== "undefined" && module && module.exports ) {
 if ( typeof exports !== "undefined" && exports ) {
 	exports.QUnit = QUnit;
 }
+
+if ( typeof define === "function" && define.amd ) {
+	define( function() {
+		return QUnit;
+	} );
+	QUnit.config.autostart = false;
+}
--- a/test/amd.html
+++ b/test/amd.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit AMD Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+	<script src="../node_modules/requirejs/require.js"></script>
+</head>
+<body>
+	<div id="qunit"></div>
+	<script>
+	var beginData;
+
+	require.config({
+		paths: {
+			qunit: "../dist/qunit"
+		}
+	});
+
+	require( [ "qunit", "./amd" ], function( QUnit, tests ) {
+		QUnit.begin( function( data ) {
+			beginData = data;
+		} );
+		tests();
+		QUnit.start();
+	} );
+	</script>
+	<script src="swarminject.js"></script>
+</body>
+</html>
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,6 +17,7 @@
 
 	"globals": {
 		"QUnit": false,
+		"define": false,
 		"exports": false,
 		"module": false
 	}
--- a/browserstack.json
+++ b/browserstack.json
@@ -6,7 +6,8 @@
 		"test/index.html",
 		"test/startError.html",
 		"test/logs.html",
-		"test/setTimeout.html"
+		"test/setTimeout.html",
+		"test/amd.html"
 	],
 	"browsers": [
 		"chrome_previous",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   ],
   "dependencies": {},
   "devDependencies": {
-    "browserstack-runner": "0.2.3",
+    "browserstack-runner": "0.3.0",
     "commitplease": "2.0.0",
     "grunt": "0.4.2",
     "grunt-contrib-concat": "0.3.0",
@@ -45,6 +45,7 @@
     "grunt-qunit-istanbul": "0.4.5",
     "grunt-search": "0.1.6",
     "load-grunt-tasks": "0.3.0",
+    "requirejs": "2.1.16",
     "testswarm": "1.1.0"
   },
   "scripts": {
--- a/test/amd.js
+++ b/test/amd.js
@@ -1,0 +1,13 @@
+/* global beginData */
+define( [ "qunit" ], function( QUnit ) {
+
+return function() {
+	QUnit.module( "AMD autostart" );
+
+	QUnit.test( "Prove the test run started as expected", function( assert ) {
+		assert.expect( 1 );
+		assert.strictEqual( beginData.totalTests, 1, "Should have expected 1 test" );
+	} );
+};
+
+} );


reporter/html.js
```diff
@@ -747,13 +747,19 @@ QUnit.testDone(function( details ) {
 	}
 });
 
-if ( !defined.document || document.readyState === "complete" ) {
+if ( defined.document ) {
+	if ( document.readyState === "complete" ) {
+		if ( window.define && window.define.amd ) {
+			QUnit.config.autostart = false;
```
jugglinmike: Thanks for this! It's definitely a common gotcha for AMD users. I have one question, though: if it is an AMD environment, shouldn't auto-start be disabled regardless of the document's `readyState`?

scottgonzalez: We could certainly make it do that. The thought was that if the loading isn't complete, then you probably didn't load QUnit through AMD, even if AMD is available. But I think it's probably ok to assume that if AMD is available that you'll always want auto-start disabled.

scottgonzalez:
I added a test for this. It's a little ghetto in the sense that it pretends to be an AMD module loader instead of actually using one, but I think that's ok.

The real problem is that for some reason the test won't run in grunt.

scottgonzalez:
I'm now always setting `QUnit.config.autostart = false;` when AMD is detected. I've also added a `define()` call so that when requiring QUnit in AMD, you actually get a module.

JamesMGreene:
Whoa... we seriously didn't have the `define` call in our exports? Weeeeeird. :astonished:

scottgonzalez:
Does anyone have ideas on why the tests don't work in Grunt?

test/amd.html
```diff
@@ -0,0 +1,50 @@
...
+		window.define = function() {};
+		window.define.amd = true;
+		setTimeout(function() {
+			var script = document.createElement( "script" );
+			script.src = "../dist/qunit.js";
+			script.onload = function loadTests() {
+				QUnit.load = (function( _load ) {
+					times.pageLoaded = now();
+					_load.call( QUnit );
+				})( QUnit.load );
```
leobalter:
```js
QUnit.load = (function( _load ) {
	return function() {
		times.pageLoaded = now();
		load.call( QUnit );
	}
})( QUnit.load );
```

leobalter: It doesn't solve the problem on Grunt, but it doesn't replace QUnit.load for a undefined value.

leobalter:
The grunt test isn't passing because of the grunt task for qunit, which requires QUnit to be loaded directly, not via AMD.

leobalter:
I've set a `console.log( typeof QUnit );` before [this line](https://github.com/asciidisco/grunt-qunit-istanbul/blob/master/phantomjs/bridge.js#L14) and yes, QUnit is undefined at the phantomjs bridge.

I'll use the official grunt-contrib-qunit task to see if it works or maybe the alternative might be loading QUnit via a regular script tag and the tests via AMD.

test/amd.html
scottgonzalez: Oh, good catch. This is actually just copied from the existing `autostart.html` file, so we'll need to update that as well.

Gruntfile.js
```diff
@@ -94,6 +94,7 @@ grunt.initConfig({
 		qunit: [
 			"test/index.html",
 			"test/autostart.html",
+			"test/amd.html",
```
scottgonzalez: @leobalter pointed out that I should add this to the Browserstack tests as well. Is there a reason the two lists aren't in sync? I found this list by searching for "autostart.html".

scottgonzalez:
Sounds like we might want to just land this without the grunt test (and only include it in BrowserStack) for now. The grunt task can be updated to handle this later.

JamesMGreene:
Phantom is notorious for having AMD + QUnit issues. :-1:

Gruntfile.js
leobalter: There's no specific reason. It would be certainly better if possible to import this list from a central point, but it can't be done this way as browserstack.json is static, and not the best place to make Gruntfile import this list.

leobalter:
Ok to land this out of grunt tasks, but should we leave a comment note, maybe on Gruntfile with an issue reference?

scottgonzalez:
> Ok to land this out of grunt tasks, but should we leave a comment note, maybe on Gruntfile with an issue reference?

Sure. If we're good to land this. I can file an issue about this and add a comment before merging.

test/amd.html
```diff
@@ -0,0 +1,52 @@
...
+	<title>QUnit AMD Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+</head>
+<body>
+	<div id="qunit"></div>
+	<script>
+	var beginData,
+		asyncDelay = 1000,
+		times = {
+			pageLoaded: 0,
```
leobalter: We can change this value to null or false to improve the assertion, otherwise it won't break the tests.

src/export.js
```diff
@@ -61,3 +61,10 @@ if ( typeof module !== "undefined" && module && module.exports ) {
 if ( typeof exports !== "undefined" && exports ) {
 	exports.QUnit = QUnit;
 }
+
+if ( typeof define === "function" && define.amd ) {
+	define( [], function() {
+		return QUnit;
+	} );
```
leobalter: we can remove the `[]` argument. Ref http://www.requirejs.org/docs/api.html#deffunc

Also, I'm not up to date on the docs, but it shouldn't close as `});` instead of `} );`?

scottgonzalez: > we can remove the [] argument. Ref http://www.requirejs.org/docs/api.html#deffunc

ok

> Also, I'm not up to date on the docs, but it shouldn't close as }); instead of } );?

Not anymore: http://contribute.jquery.org/style-guide/js/#spacing

test/amd.html
scottgonzalez: These actually only have values because I started from a copy of the autostart test. If I had written this from scratch, this would just be an empty object.

scottgonzalez:
I just pushed another update. Note that it's actually impossible to proxy `QUnit.load()` in the AMD case because it's invoked immediately upon loading.

This is probably why the autostart test "fix" didn't work properly either.

src/export.js
leobalter: thanks.

test/amd.html
```diff
@@ -0,0 +1,43 @@
...
+			script.onload = function loadTests() {
+				QUnit.begin(function( data ) {
+					beginData = data;
+					times.runStarted = now();
+				});
+				var script = document.createElement( "script" );
+				script.src = "amd.js";
+				document.getElementsByTagName( "head" )[ 0 ].appendChild( script );
+			};
+			times.pageLoaded = now();
```
leobalter: I'm sorry going so deep here, but I believe this pageLoaded < runStarted won't assert anything, pageLoaded will always be lower than runStarted in this case.

At this point, it's good to get rid of the times object and the respective assertion. The beginData assertion will be enough until we find a better approach.

scottgonzalez: Removed the time assertions.

leobalter:
LGTM

jzaefferer:
Looks good to me, too. Since the reporter still doesn't care about window.define, "Exports" is a better component for this.

scottgonzalez:
The tests are timing out with browserstack-runner, but they're working for me manually via BrowserStack.

jzaefferer:
So the Travis build also failed with multiple timeouts: https://travis-ci.org/jquery/qunit/builds/55530371

But there's something else wrong. Multiple tests are reported as passed multiple times, for example this one:

```
[Windows 8, Internet Explorer 10.0, test/logs.html] Passed: 50 tests, 50 passed, 0 failed; ran for 38ms
[Windows 8, Internet Explorer 10.0, test/logs.html] Passed: 52 tests, 52 passed, 0 failed; ran for 5059ms
[Windows 8, Internet Explorer 10.0, test/logs.html] Passed: 54 tests, 54 passed, 0 failed; ran for 5063ms
```

I noticed that for both `setTimeout.html` and `logs.html`. The last successful build from master didn't have that issue.

scottgonzalez:
> Phantom is notorious for having AMD + QUnit issues but I've never seen a specific resolution that worked for everyone either.

I have AMD + grunt-contrib-qunit working locally right now. I'll be posting updates soon.

scottgonzalez:
PR to get us one step closer to proper AMD tests with grunt-contrib-qunit: https://github.com/gruntjs/grunt-contrib-qunit/pull/94

jzaefferer:
Released grunt-contrib-qunit 0.6.0 and browserstack-runner 0.3.0 with fixes for AMD support.

scottgonzalez:
I've rebased and squashed the commits. Tests are now passing in Travis. I think we're ready to merge :-)

leobalter:
Merged.

scottgonzalez:
:tada:

jugglinmike:
Awesome!

jzaefferer:
Finally figured out what we were missing: We didn't add the new `test/amd.html` to run via PhantomJS (add in `Gruntfile.js`), and of course adding it results in a timeout. I've put together [a patch for `grunt-qunit-istanbul`](https://github.com/asciidisco/grunt-qunit-istanbul/compare/master...jzaefferer:amd-bridge?expand=1) to apply the same treatment as for `grunt-contrib-qunit`, but I still get a timeout. Not sure what I'm missing.

jzaefferer:
The timeout came from me forgetting to run `npm install`, so `require.js` was missing. Sent PRs to patch grunt-qunit-istanbul for AMD support, and another one for that and the original module to log PhantomJS errors.

PR #792, opened 2015-03-21
HTML Reporter: Escape backslash when parsing string

### LINKED ISSUE #478: Backslash in string inconsistency in diff
Hi,

Please have a look at http://jsfiddle.net/qEdGD/3/ .
The strings I compare contain both an escaped `"` and an escaped `\`.

The diff shows the escaped `"` as `\"`, but the escaped `\` as `\`. Thus: `'\" versus \\'` becomes `"\" versus \"`. I think this is confusing (especially if there is another character behind the escaped `\`), and it would be better to also show the `\` as `\\` due to it's special meaning within strings.

Kind regards,
Famlam

```javascript
test("A failing test", function () {
    deepEqual('\" versus \\', "", "A test that fails");
});
```

shivamdixit:
Screenshot:
qunit

Fixes #478

### DIFF
--- a/src/dump.js
+++ b/src/dump.js
@@ -2,7 +2,7 @@
 // http://flesler.blogspot.com/2008/05/jsdump-pretty-dump-of-any-javascript.html
 QUnit.dump = (function() {
 	function quote( str ) {
-		return "\"" + str.toString().replace( /"/g, "\\\"" ) + "\"";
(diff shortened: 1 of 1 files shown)


jzaefferer:
This looks fine, but without any tests that could help prevent regressions its hard to tell if we're actually making things worse.

jzaefferer:
Landed this anyway, its a simple change that we can improve later if its turns out to be problematic. Rebased the commit and update the commit message, now using the right component.

PR #794, opened 2015-03-22
Dump maxdepth

gauravmittal1995:
Making the MaxDepth configurable. We can change it using:
<pre>
QUnit.config.maxDepth = 6;
</pre>

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -103,6 +103,9 @@ config = {
 	// when enabled, all tests must call expect()
 	requireExpects: false,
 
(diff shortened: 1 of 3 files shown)


jzaefferer:
The current property is actually documented here: http://api.qunitjs.com/QUnit.dump.parse/

Its a rather recent addition and inconsistent with other configurable features. Considering the other issues/PRs about showing only relevant diffs and diffing big trees, we may need to rework a lot more than just moving this property from one place to the other.

jzaefferer:
This was addressed by #795.

Issue #793, opened 2015-03-21
Diff is not shown even when two objects are different


shivamdixit:
For the following deepEqual comparison, no diff is displayed. Is this an expected behavior?

```javascript
assert.deepEqual(
    {
        d: {
            c: {
                g: {
                    j: {
                        k: {
                            n: {
                                r: "s",
                                s: [ 1, 2, 3, 4 ],
                                t: undefined,
                                u: 1,
                                v: {
                                    w: {
                                        x: {
                                            y: "Hola!",
                                            z: "1"
                                        }
                                    }
                                }
                            },
                            q: [],
                            p: 1 / 0,
                            o: 99
                        },
                        l: undefined,
                        m: null
                    }
                }
            }
        }
    },
    {
        d: {
            c: {
                g: {
                    j: {
                        k: {
                            n: {
                                r: "p",
                                t: undefined,
                                u: 0,
                                s: [ 1, 2, 3 ],
                                v: {
                                    w: {
                                        x: {
                                            z: null,
                                            y: "Yahoo!"
                                        }
                                    }
                                }
                            },
                            o: 99,
                            p: 1 / 0,
                            q: []
                        },
                        l: undefined,
                        m: null
                    }
                }
            }
        }
    }
, true);
```

Output:
![qunit](https://cloud.githubusercontent.com/assets/3061095/6765017/f0e86d3c-cff3-11e4-8285-375999f2f9aa.png)

Diff is also not show in various other complex object assertions too.

gauravmittal1995:
I think that this is because https://github.com/jquery/qunit/blob/master/src/dump.js#L163#L165. Here since your dept at which the difference is seen is more than the default MaxDepth(5) , It gives [Object object]. Also here the diff is not seen because even though its different (and that's why it gives an error), the displayed text is essentially the same(i.e object Object). If you change the MaxDepth (https://github.com/jquery/qunit/blob/master/src/dump.js#L130) to 7 or 8, it will show a diff.

shivamdixit:
@gauravmittal1995 Thanks, I was not aware of `maxDepth` property.

scottgonzalez:
This would require exposing some additional logic, but it might be interesting to show a notice next to the assertion. Potentially even providing a link to rerun the test with a higher max depth.

shivamdixit:
@scottgonzalez I think that's a great idea. Shall I open an issue for the same?

PR #795, opened 2015-03-22
HTML Reporter: Make it more obvious why diff is suppressed

shivamdixit:
If the depth of an object is more than 'Max Depth' specified in `Qunit.dump` then the diff is not displayed. This behavior is confusing. This PR will display a message to the user so that user knows why it is not displayed.

Refer: https://github.com/jquery/qunit/issues/793#issuecomment-84495125

I'm confused with what text to display to the user since the user might not be aware of max depth. Any suggestions?

Screenshot:
qunit4

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -668,7 +668,7 @@ QUnit.log(function( details ) {
 					expected.indexOf( "[object Object]" ) !== -1 ) {
 				message += "<tr class='test-message'><th>Message: </th><td>" +
 					"Diff suppressed as the depth of object is more than current max depth (" +
-					QUnit.config.maxDepth + ").<p>Hint: Use 'QUnit.config.maxDepth' to run " +
+					QUnit.config.maxDepth + ").<p>Hint: Use 'QUnit.dump.maxDepth' to run " +
 					" with a higher max depth or <a href='?maxDepth=-1'>" +
 					"Rerun</a> without max depth</p></td></tr>";
 			}
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -668,8 +668,8 @@ QUnit.log(function( details ) {
 					expected.indexOf( "[object Object]" ) !== -1 ) {
 				message += "<tr class='test-message'><th>Message: </th><td>" +
 					"Diff suppressed as the depth of object is more than current max depth (" +
-					QUnit.config.maxDepth + ").<p>Hint: Use 'QUnit.dump.maxDepth' to run " +
-					" with a higher max depth or <a href='?maxDepth=-1'>" +
+					QUnit.config.maxDepth + ").<p>Hint: Use <code>QUnit.dump.maxDepth</code> to " +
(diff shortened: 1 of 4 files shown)


scottgonzalez:
Ideally it says what the current max depth is and provides a link to rerun with a higher max depth or no max depth.

shivamdixit:
@scottgonzalez I've updated the PR. Screenshot:

qunit5

1. Added `maxDepth` in url parameter, `maxDepth = -1` implies run the tests without any `maxDepth`.

2. Although `maxDepth = someVal` in URL parameters will work, but not *explicitly* showing the option of inputting the value from user because the UI will be cluttered as we will need to add input box, button etc.

3. Instead showing `QUnit.dump.maxDepth` as hint, in case user wants to change `maxDepth`.

4. The color of `Rerun` looks dull in the message, any suggestions for a different color?

5. As `QUnit.dump` is not in scope when we are parsing URL params, hence I've added extra property `QUnit.config.maxDepth`.

@jzaefferer Please provide your valuable suggestions as well.

shivamdixit:
I think the tests are failing because of the timeout. How can I re-run the tests?

reporter/html.js
```diff
@@ -663,6 +663,15 @@ QUnit.log(function( details ) {
 				actual + "</pre></td></tr>" +
 				"<tr class='test-diff'><th>Diff: </th><td><pre>" +
 				QUnit.diff( expected, actual ) + "</pre></td></tr>";
+		} else {
+			if ( expected.indexOf( "[object Array]" ) !== -1 ||
+					expected.indexOf( "[object Object]" ) !== -1 ) {
+				message += "<tr class='test-message'><th>Message: </th><td>" +
+					"Diff suppressed as the depth of object is more than current max depth (" +
+					QUnit.config.maxDepth + ").<p>Hint: Use 'QUnit.dump.maxDepth' to run " +
+					" with a higher max depth or <a href='?maxDepth=-1'>" +
```
jzaefferer: This should use `setUrl()` to generate QUnit specific URLs that include other URL parameters already present (filters etc.).

jzaefferer:
@shivamdixit thanks for the PR. I have very little bandwidth to review anything this week. I added one comment above. I'd like to discuss the general concept for tackling this issue, but that has to wait a few days.

scottgonzalez:
That screenshot looks great. You should wrap `QUnit.config.maxDepth` in `<code>` instead of quotes.

Fixing the styling of the rerun link should be very simple, but let's wait for the team to discuss.

shivamdixit:
@jzaefferer @scottgonzalez Thanks for your suggestions. I've updated the code.

jzaefferer:
Let's add some CSS to display the Rerun link properly, just change `color` to something with more contrast.

shivamdixit:
qunit5

How is it looking?

leobalter:
LGTM

PR #797, opened 2015-03-23
Test: Fix override of QUnit.load()

scottgonzalez:
As pointed out in https://github.com/jquery/qunit/pull/791#discussion_r26896124.

### DIFF
--- a/test/autostart.html
+++ b/test/autostart.html
@@ -36,8 +36,10 @@
 		}, asyncDelay );
 
 		QUnit.load = (function( _load ) {
-			times.pageLoaded = now();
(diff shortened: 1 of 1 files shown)


scottgonzalez:
So this didn't work and actually points out a problem with the existing test. `QUnit.load()` was being called immediately and `times.pageLoaded` was the time of the override, not the time of the "real" call (which apparently never happened).

scottgonzalez:
I'm going to close this PR because I don't want to spend any more time digging into the crazy loading process, but I'll file an issue about this test being incorrect.

Issue #363, opened 2012-12-06
Diff only leafs for (big) objects


jzaefferer:
When comparing big object structures, the current diff output has too much overhead. Comparing each leaf and showing the path, actual, expected (and diff) for just the leaf would be much more useful. It won't replace the regular diff, e.g. comparing (long) strings wouldn't profit from this.

In other words, we'd have to find the properties that are different, serialize and diff those, and not output the rest, or at least compress it.

shivamdixit:
Only outputting properties that are different might become confusing sometimes as the structure of object will not be maintained. Instead we can make the objects **collapsible** to compress the output of the diff. For example:
![qunit3](https://cloud.githubusercontent.com/assets/3061095/6765178/3cf77180-cffb-11e4-9d39-13adebd5b72f.png)
Can be collapsed to:

![qunit2](https://cloud.githubusercontent.com/assets/3061095/6765181/44b620b0-cffb-11e4-9b2b-c2175872467a.png)

When showing the diff, the unchanged properties can be collapsed by default. It will compress the output significantly. We can also have configuration options for it. For example:
* By default collapse only changed properties
* By default collapse all - `expected` or `actual` or `diff` output  etc

Possible challenges:
1. Determining which properties have changed. As we are dumping the objects first and then diff-ing them, it might not be straightforward. Maybe we can check the properties which are containing the diff tags `<ins>` or `<del>` and make them collapsible. Not sure if it is a good way.
2. The output of objects and other types will have to be handled differently in HTML reporter as strings, int etc won't be "collapsible".

What are your views?

gibson042:
I think it depends on the case(s) we're trying to improve. Diffs of objects with many properties, which I suspect is a principle pain point, would benefit most from hiding completely-unchanged properties. Deep diffs, on the other hand, would probably benefit more from compressing the structure _above_ changes (e.g., replacing `arr: [\n ...,\n {\n ...,\n differs: {\n ...` literal representation with `["arr"][N]["differs"]: {\n ...` object/array reference notation)... the challenge there being preserving enough context in objects (but probably not arrays) containing the difference.

gibson042:
That said, I really like the dynamic collapse concept for conditionally showing excess data. But I think it would apply to lists of unchanged properties instead of whole objects/arrays, at least for the "many properties, few diffs" case.

shivamdixit:
[...]

PR #799, opened 2015-03-24
HTML Reporter: Display only the different part in diff

gauravmittal1995:
This is a very basic implementation.
Here the diff removes the unnesseray part (in big objects or arrays), and only displays the diff part.
It is like the github's diff, where common parts are hidden and only diff is seen.

Also there is no styling done
This PR is mostly for feedbacks.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -63,6 +63,8 @@ if ( typeof window === "undefined" ) {
 }
 
 var config = QUnit.config,
+	depthExpand = 0,
+	expanderiator = 0,
 	hasOwn = Object.prototype.hasOwnProperty,
 	defined = {
 		document: window.document !== undefined,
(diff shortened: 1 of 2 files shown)


gauravmittal1995:
Will post Screenshots soon

gauravmittal1995:
Initially like: 
diff_collapse1

On clicking on the first [+] symbol, expands the common part before the first difference. : 
diff_collapse2

On clickicking the first [+] again, collapses it back.
Similarly for the second:
diff_collapse3

Note: There is no styling done, and also its nowhere near complete. Its just a basic implementation, mostly for feedbacks.
Issue at: https://github.com/jquery/qunit/issues/363

gauravmittal1995:
Code needs to be refactored, quiet messy.
Also, with this implementation, can we remove the maxDepth from the dump.js?

gauravmittal1995:
@jzaefferer Please review

jzaefferer:
@leobalter can you review this?

reporter/html.js
```diff
@@ -610,6 +611,114 @@ function getNameHtml( name, module ) {
...
+			}
+		}
+	}
+	str = temp_array;
+	diff = "";
+	for(iter = 0;iter<str.length;iter++) {
+		diff += str[iter];
+	}
+	return diff;
+}
```
leobalter: this is unpleasantly complex. I'll need more time to review the interactions on each loop.

There isn't a way to simplify this function? It would also be easier to review this if following the style guide.

Also, the very similar names like `split_array1` and `split_array` make this function harder to follow along.

leobalter:
I liked the idea, but it is looking very complex with a lot to improve before it's good to merge. Also, the style would be indeed good.

Maybe, instead of this approach, we could show a simplified diff (focused only on the different parts) with a toggle link to show the full diff (without changing the maxDepth).

gauravmittal1995:
@leobalter Hey, this PR is mostly for the feedback on the idea. Once its clear, i will properly style everything and refactor the code.

I didn't quiet understand the other approach. Can u explain a bit more?
In this (this PR) approach, only the different parts are shown and a toggle button is given to expand the difference between the previous different part and the current different part. Somewhat like the github's diff

reporter/html.js
gauravmittal1995: @leobalter Don't review this code, as its just used to implement the idea in a basic way. This PR is mostly for feedback of the idea. Once its clear, i will refactor everything and then u can properly review it.

leobalter:
I just told an non elaborated idea, most because I didn't like the complexity level of this implementation.

For the general idea presented (by you) in this PR: **I like it**, but I would need to see a cleaner code to review it.

gauravmittal1995:
@leobalter i will send a cleaner code by tomorrow. Thanks

gauravmittal1995:
Screenshots:

1) Inital Look, Only the diff parts + one line before and after will be shown. Rest will be hidden and can be toggled with the [+] symbol.
collapse1

2) Expanding the first part. Toggles [+] symbol to [-] and if clicked, hides the common part. Has a background grey color similar to github. Question: does the positioning of the [+] and [-] symbol look right?
collapse2

3) On expanding everything.
collapse3

@jzaefferer @leobalter Please Review

gauravmittal1995:
@leobalter Ping

leobalter:
@gauravmittal1995 I'm sorry I've been very busy since then.

After some work on QUnit, and the PR #850, it seems these would be better at the diff tool.

I also miss the tests here.

leobalter:
Due to the inactivity, I'll close this PR. We can open it back if you wanna continue working on it.

PR #800, opened 2015-03-25
Core: Nested modules

### LINKED ISSUE #543: Allow nested suites (modules)?

Askelkana:
Allows nesting of modules by enabling passing function as third argument
to QUnit.module method.

Fixes #543

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -278,10 +278,10 @@ extend( QUnit, {
 	},
 
 	skip: function( testName ) {
-		var test = new Test( {
+		var test = new Test({
 			testName: testName,
 			skip: true
-		} );
+		});
 
 		test.queue();
 	},
@@ -407,7 +407,7 @@ extend( QUnit, {
 			resumeProcessing();
 		}
 	}
-} );
+});
 
 // Register logging callbacks
 (function() {
@@ -504,7 +504,7 @@ function done() {
 		passed: passed,
 		total: config.stats.all,
 		runtime: runtime
-	} );
+	});
 }
 
 // Doesn't support IE6 to IE9
--- a/src/core.js
+++ b/src/core.js
@@ -128,6 +128,9 @@ config = {
 	// Set of all modules.
 	modules: [],
 
+	// Stack of nested modules
+	moduleStack: [],
+
 	// The first unnamed module
 	currentModule: {
 		name: "",
@@ -194,12 +197,15 @@ config.modules.push( config.currentModule );
 extend( QUnit, {
 
 	// call on start of module test to prepend name to all tests
-	module: function( name, testEnvironment ) {
-		var currentModule = {
-			name: name,
-			testEnvironment: testEnvironment,
-			tests: []
-		};
+	module: function (name, testEnvironment, executeNow) {
+		var currentModule = config.currentModule;
+		var module;
+		if (arguments.length === 2) {
+			if (testEnvironment instanceof Function) {
+				executeNow = testEnvironment;
+				testEnvironment = undefined;
+			}
+		}
 
 		// DEPRECATED: handles setup/teardown functions,
 		// beforeEach and afterEach should be used instead
@@ -212,8 +218,33 @@ extend( QUnit, {
 			delete testEnvironment.teardown;
 		}
 
-		config.modules.push( currentModule );
-		config.currentModule = currentModule;
+		module = createModule();
+		if (executeNow instanceof Function) {
+			config.moduleStack.push( module );
+			setCurrentModule( module );
+			executeNow();
+			config.moduleStack.pop();
+			module = module.parentModule || currentModule;
+		}
+		setCurrentModule( module );
+
+		function createModule() {
+			var parentModule = config.moduleStack.length ? config.moduleStack.slice( -1 )[ 0 ] : null;
+			var moduleName = parentModule !== null ? [ parentModule.name, name ].join( " > " ) : name;
+			var module = {
+				name: moduleName,
+				parentModule: parentModule,
+				testEnvironment: testEnvironment,
+				tests: []
+			};
+			config.modules.push( module );
+			return module;
+		}
+
+		function setCurrentModule(module) {
+			config.currentModule = module;
+		}
+
 	},
 
 	// DEPRECATED: QUnit.asyncTest() will be removed in QUnit 2.0.
--- a/src/core.js
+++ b/src/core.js
@@ -2,15 +2,14 @@ var QUnit,
 	config,
 	onErrorFnPrev,
 	loggingCallbacks = {},
-	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//,
-		"" ),
+	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//, "" ),
 	toString = Object.prototype.toString,
 	hasOwn = Object.prototype.hasOwnProperty,
-// Keep a local reference to Date (GH-283)
+	// Keep a local reference to Date (GH-283)
 	Date = window.Date,
 	now = Date.now || function() {
-			return new Date().getTime();
-		},
+		return new Date().getTime();
+	},
 	globalStartCalled = false,
 	runStarted = false,
 	setTimeout = window.setTimeout,
@@ -24,8 +23,7 @@ var QUnit,
 				sessionStorage.setItem( x, x );
 				sessionStorage.removeItem( x );
 				return true;
-			}
-			catch ( e ) {
+			} catch ( e ) {
 				return false;
 			}
 		}())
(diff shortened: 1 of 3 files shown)


Askelkana:
I signed the CLA prior to creating the pull request (losing my virginity with this). Don't know what happens after that...

arthurvr:
@Askelkana You used your username when signing the CLA, not your full name.

Askelkana:
Actually, I didn't. But I'll have another go...

arthurvr:
@Askelkana Now the CLA appears to be signed with your full name, thanks. You'll have to update your git config to include your full name as well. Normally it should be fine then.

Askelkana:
By way of interest, the CLA page reads as follows:

> The email address used to sign the CLA must match the email address in your git config. The author information for each commit in your Contributions will be cross-checked with this signature.

It does not mention that the name given is significant.

Askelkana:
Ok, git config updated and commited. I assume the CI build fail is not critical, since there are several errors being thrown which have nothing to do with my changes.

leobalter:
I'll review this tomorrow morning but you also need to run grunt in order to lint and check the code style.

This PR looks interesting, I'm looking forward to see it passing the on CI.

leobalter:
current tests are working, but I also made this following test in your branch:

```js
QUnit.module( "multiple hooks", {
	beforeEach: function( assert ) {
		assert.ok( true, "root module beforeEach" );
	},
	afterEach: function( assert ) {
		assert.ok( true, "root module afterEach" );
	}
}, function() {
	QUnit.module( "nested module", {
		beforeEach: function( assert ) {
			assert.ok( true, "nested module beforeEach" );
		},
		afterEach: function( assert ) {
			assert.ok( true, "nested module afterEach" );
		}
	});

	QUnit.test( "all hooks must be called", function( assert ) {
		assert.expect( 4 );
	});
});
```

It failed as it only triggered the nested module hooks. This test should pass triggering all the chained modules' hooks.

IMHO, the only real advantage on having nested modules is to stack the hooks. Otherwise it would just create a fancy report.

cc @JamesMGreene.

Askelkana:
I think your tests are incorrect: surely the `afterEach` would be called after the test has run, and therefore any assertions fired there will not be caught by the `expect` method. However, I have integrated the intent of the check into my original tests.

I wanted to wait for feedback before attempting this, even though it was mentioned in the original issue's comments. So thanks for the confirmation.

I would add that another advantage to nesting modules is that it clears up one's testing code no end, since it allows for much better arrangement of the tests.

I'll push a commit through soon.

leobalter:
the expect also counts the assertions in the following afterEach hook.

Some other issues I noticed:
- the nested modules are not listed on the HTML Reporter
- selecting the root modules on the HTML Reporter filter doesn't run the tests on its respective nested modules

Askelkana:
You are right. Tests still holding.

As for the HTMLReporter, at present the modules are not highlighted in any way. I didn't wish to presume how this should be handled, so I plumbed for simply showing the module path on the test row, as before.

I agree that ideally, some nesting mechanism should be portrayed by the reporter.

src/core.js
```diff
@@ -67,8 +68,8 @@ var QUnit,
 	objectValues = function( obj ) {
 		var key, val,
 			vals = QUnit.is( "array", obj ) ? [] : {};
-		for ( key in obj ) {
-			if ( hasOwn.call( obj, key ) ) {
+		for (key in obj) {
+			if (hasOwn.call( obj, key )) {
```
leobalter: something is breaking the code style, and it's becoming hard to check the real diffs here.

Askelkana: Ok, I'll check it out. Everything passed the jshint and jscs tests, so I didn't notice these.

The actual changes this time are in the test.run and test.hook methods.

Askelkana:
> Some other issues I noticed:
> 
> the nested modules are not listed on the HTML Reporter
> selecting the root modules on the HTML Reporter filter doesn't run the tests on its respective nested modules

Well, that may be so, but currently QUnit doesn't provide any support for modules in the HTML Reporter. The features you highlight here are really independent of my proposed changes, in that they could (and should) already be there.

Askelkana:
Ok, I just pushed an addition that all tests belonging to a super-module are run when selecting the said module from the filter. Unless you have any other issues, I think this is good to go.

src/core.js
```diff
@@ -2,14 +2,15 @@ var QUnit,
 	config,
 	onErrorFnPrev,
 	loggingCallbacks = {},
-	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//, "" ),
+	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//,
```
jzaefferer: There are lots of whitespace or linebreak changes like this. They don't belong in this PR, most of them seem just plain wrong. Can you address that? We can then do a proper review (more timely, hopefully).

src/test.js
```diff
@@ -19,15 +19,16 @@ function Test( settings ) {
 
 	this.testId = generateHash( this.module.name, this.testName );
 
-	this.module.tests.push({
+	this.module.tests.push( {
 		name: this.testName,
 		testId: this.testId
-	});
+	} );
```
jzaefferer: These changes are fine, since they actually match the styleguide

src/test.js
```diff
@@ -42,13 +43,13 @@ Test.prototype = {
 		if (
 
 			// Emit moduleStart when we're switching from one module to another
-			this.module !== config.previousModule ||
+		this.module !== config.previousModule ||
```
jzaefferer: This isn't, it just makes things worse.

src/test.js
jzaefferer: But they should happen in a separate PR, not in one that introduces a new feature.

Askelkana:
Apologies for the delay. Slowed by poor health and a garden to get ready for planting...

Anyway, I've fixed (or think I have) the style issues. I take your point about not trying to solve more than one problem at a time. But the code really could stand some cleaning: it's quite inconsistent in many cases, and certainly doesn't pass the jslint test.

Askelkana:
Yes, all styles issues have now been resolved. Happy reviewing!

jzaefferer:
@leobalter what do you think?

leobalter:
I believe fixing the conflict issues won't be too hard. 

I'll merge this and then I'll propose an extra PR to make the QUnit.module functions to come with the `test` and `QUnit` as the arguments.

---

The `test` argument is just to have it as a local reference (avoiding globals without using namespaces, aka `QUnit.test`); The `QUnit` argument is to allow access to additional features, as `QUnit.skip`, `QUnit.config`, the proposed `QUnit.only`, etc.

A second following PR proposal is to also alias QUnit.module to the QUnit. This is more insane, but would allow a simple and dynamic use with these nested modules:

```js
import QUnit from 'qunitjs';

QUnit( "my first batch", ( test ) => {
  test( "test this", ( t ) => {
    t.ok( true, "it pass" );
  } );
} );
```

I'll open the respective issues to discuss them as soon as I merge this PR.

gibson042:
This PR exhibits some potentially surprising behaviors:
* executing all of a module's own tests before entering its submodules (regardless of test/submodule definition order)
* always minting a new environment definition for every submodule, even if its `hooks` argument is `undefined` or not even specified

If they are intentional, there should be assertions covering them. If they are unintentional and undesired, there should be assertions covering the desired behavior.

leobalter:
The squashed and rebased branch is here: https://github.com/jquery/qunit/compare/jquery:master...leobalter:nested-modules?expand=1

---

@gibson042 

> executing all of a module's own tests before entering its submodules (regardless of test/submodule definition order)

This might seem weird, but with the idea of each test should be atomic, an specific order shouldn't be a major concern. The real world prevents me to keep this argument as this will indeed blow out eventually without assertions with the intentional behavior.

> always minting a new environment definition for every submodule, even if its hooks argument is undefined or not even specified

I don't see it clearly, but I believe this and along with the other behavior could be improved in a next PR?

The main feature seems ok and working fine, and I am late giving the proper attention to this work. So if you agree, I would let these parts to be treated in a next commit bit.

gibson042:
> The main feature seems ok and working fine, and I am late giving the proper attention to this work. So if you agree, I would let these parts to be treated in a next commit bit.

:+1:

I just didn't want to lose these thoughts (or the one regarding [explicit context sensitivity of `beforeEach`/`afterEach`/`test`/etc.](https://github.com/jquery/qunit/pull/670#discussion_r26321413) via arguments or context, which should be covered by your above followup).

test/modules.js
```diff
@@ -147,3 +147,51 @@ QUnit.module( "Deprecated setup/teardown", {
...
+				assert.expect( 3 );
+			});
+		});
+	QUnit.module( "second outer" );
+	QUnit.test( "test after all nesting modules processed and new module declared",
+		function( assert ) {
+			var module = assert.test.module;
+			assert.equal( module.name, "nested modules > second outer" );
+		});
+});
```
leobalter: I found an interesting and possibly blocking behavior:

`QUnit.module` calls with functions are not affecting the following tests written outside the function.

Like this following example right after this line where I am commenting:

```js
QUnit.test( "test foo", function( a ) {
	a.equal( a.test.module.name, "nested modules" );
} );
```

This test is failing, telling the actual value is `"Deprecated setup/teardown"`, referencing the module written above `"nested modules"`.

leobalter: @jzaefferer The implementation is tied to this behavior.

I believe this can still be shipped but mentioning that calling `QUnit.module` with a nested function for tests will not spread the current module for consequent `QUnit.test` our of its nested function.

This should be tested, documented and we might want to deprecate the usage of `QUnit.module` without a function parameter.

leobalter: also: calling `QUnit.module` without a function inside a `QUnit.module` function will cause a strong mess with the nested behavior. That's a reason it should be deprecated.

leobalter: please, see #859

leobalter:
Closing this, further work in progress on #859

PR #802, opened 2015-03-27
Show diff only when it helps

shivamdixit:
Follow up of #786

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -643,7 +643,7 @@ function stripHtml( string ) {
 QUnit.log(function( details ) {
 	var assertList, assertLi,
 		message, expected, actual, diff,
-		lengthChanged = true,
+		showDiff = false,
 		testItem = id( "qunit-test-output-" + details.testId );
 
 	if ( !testItem ) {
@@ -665,16 +665,24 @@ QUnit.log(function( details ) {
 			"</pre></td></tr>";
 
 		if ( actual !== expected ) {
-			diff = QUnit.diff( expected, actual );
-			lengthChanged = stripHtml( diff ).length !==
-				stripHtml( expected ).length +
-				stripHtml( actual ).length;
 
-			// don't show diff if expected and actual are totally different
 			message += "<tr class='test-actual'><th>Result: </th><td><pre>" +
-				actual + "</pre></td></tr>" +
-				( lengthChanged ? "<tr class='test-diff'><th>Diff: </th><td><pre>" +
-					diff + "</pre></td></tr>" : "" );
+				actual + "</pre></td></tr>";
+
+			// Don't show diff if actual or expected are booleans
+			if ( !( /^(true|false)$/.test( actual ) ) &&
+					!( /^(true|false)$/.test( expected ) ) ) {
+				diff = QUnit.diff( expected, actual );
+				showDiff = stripHtml( diff ).length !==
+					stripHtml( expected ).length +
+					stripHtml( actual ).length;
+			}
+
+			// Don't show diff if expected and actual are totally different
+			if ( showDiff ) {
+				message += "<tr class='test-diff'><th>Diff: </th><td><pre>" +
+					diff + "</pre></td></tr>";
+			}
 		} else if ( expected.indexOf( "[object Array]" ) !== -1 ||
 				expected.indexOf( "[object Object]" ) !== -1 ) {
 			message += "<tr class='test-message'><th>Message: </th><td>" +
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -635,9 +635,15 @@ QUnit.testStart(function( details ) {
 
 });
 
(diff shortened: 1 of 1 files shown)


leobalter:
> Note: this is a follow up to #786 and also address an issue introduced on the last changes showing a whacky `falstrue` diff when one of the assertion values (actual or expected) was a Boolean.

shivamdixit:
Seems like the test timed out.

leobalter:
it worked now

shivamdixit:
qunit7

Booleans inside objects/arrays are still displayed like `falstrue`.

shivamdixit:
Ping.

jzaefferer:
The hack to show true/false differently is bad. This is likely to blow up with a similar combination of single word pairs, or maybe with numbers.

We want to character-by-character diff for syntax, but maybe not for literals like booleans?

Since we already landed the new diff, can we keep the "show diff only when it helps" and "show better diff for booleans" apart?

leobalter:
the regexp works only for Booleans, as strings are represented between double quotes and the regexp will search from the start to the end of the value: `/^(true|false)$/`.

leobalter:
btw, `falstrue` is a good name for a band.

leobalter:
as told on the biweekly meeting, the boolean regexp check is good as-is and @jzaefferer agreed to land it.

shivamdixit:
Great. So this PR is ready to be merged?

leobalter:
LGTM, I'll just wait for @jzaefferer :+1: here to confirm. Then I'll squash the commits and merge them.

reporter/html.js
```diff
@@ -659,19 +665,31 @@ QUnit.log(function( details ) {
...
+			// Don't show diff if expected and actual are totally different
+			if ( showDiff ) {
+				message += "<tr class='test-diff'><th>Diff: </th><td><pre>" +
+					diff + "</pre></td></tr>";
 			}
+		} else if ( expected.indexOf( "[object Array]" ) !== -1 ||
+				expected.indexOf( "[object Object]" ) !== -1 ) {
+			message += "<tr class='test-message'><th>Message: </th><td>" +
+				"Diff suppressed as the depth of object is more than current max depth (" +
+				QUnit.config.maxDepth + ").<p>Hint: Use <code>QUnit.dump.maxDepth</code> to " +
```
jzaefferer: Let's fix the `<code>QUnit.dump.maxDepth</code>` to use `QUnit.config.maxDepth` as well.

reporter/html.js
```diff
@@ -659,19 +665,31 @@ QUnit.log(function( details ) {
...
-					QUnit.config.maxDepth + ").<p>Hint: Use <code>QUnit.dump.maxDepth</code> to " +
-					" run with a higher max depth or <a href='" + setUrl({ maxDepth: -1 }) + "'>" +
-					"Rerun</a> without max depth.</p></td></tr>";
+				actual + "</pre></td></tr>";
+
+			// Don't show diff if actual or expected are booleans
+			if ( !( /^(true|false)$/.test( actual ) ) &&
+					!( /^(true|false)$/.test( expected ) ) ) {
+				diff = QUnit.diff( expected, actual );
+				showDiff = stripHtml( diff ).length !==
```
jzaefferer: This will be false when the diff is exactly as long as actual length plus expected length, right? Is that really a good measure?

jzaefferer:
Noticed two things, not sure why I missed them before.

reporter/html.js
leobalter:
```js
QUnit.diff( "1234567890", "abcdefghij" );
"<del>1234567890</del><ins>abcdefghij</ins>"
```

AFAIK, this happens when there's no diff, the objects are completely different.

leobalter: it makes sense when we already expose the actual and expected and there's nothing equal from them.

jzaefferer: Okay, we can tune it later.

reporter/html.js
shivamdixit: I just tried to change it to `QUnit.config.maxDepth` but strangely it din't work, whereas using `QUnit.dump.maxDepth` works fine. I guess this is happening because `dump.maxDepth` is initialized with default `config.maxDepth` in file `dump.js` and then the value of `config.maxDepth` is changed. Hence I got a message which looks like:

```
Diff suppressed as the depth of object is more than current max depth (infinity)
```
However specifying `maxDepth` in GET parameters works fine because if parameters are present, the value of `config.maxDepth` is updated in file `core.js` before initialization in `dump.js`.

reporter/html.js
shivamdixit: I agree with @leobalter , if objects are completely different then only the condition will be false.

jzaefferer:
Squashed the two commits into one and merged. Thanks again!

PR #803, opened 2015-03-30
Fix autostart test

### LINKED ISSUE #798: Invalid test for autostart
`test/autostart.html` currently contains an invalid test for the autostart feature. The `QUnit.load()` override is implemented incorrectly, resulting in a false positive. The method isn't actually overloaded and `QUnit.load()` is actually called instantly.

BraulioVM:
Fixes #798.

Once the ````QUnit.load```` function was overriden, the old ````QUnit.load```` had already been suscribed to the ````window.onload```` event. This PR removes the old ````QUnit.load```` from those event listeners and adds the new one.

### DIFF
--- a/test/autostart.html
+++ b/test/autostart.html
@@ -18,6 +18,54 @@
 		};
 
 	(function() {
+		/**
+		 * @param {HTMLElement} elem
+		 * @param {string} type
+		 * @param {Function} fn
+		 */
+		function addEvent( elem, type, fn ) {
+			if ( elem.addEventListener ) {
+
+				// Standards-based browsers
+				elem.addEventListener( type, fn, false );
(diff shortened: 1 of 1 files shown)


scottgonzalez:
I feel like there has to be a better change.

BraulioVM:
I'd love if you could elaborate a bit more on your response. What is it that you don't like:

* Having ````addEvent```` and ````removeEvent```` added into autostart.html?
* Removing the ````load```` event suscription?
* The general approach used to test this feature?
* Any other thing?

Thank you very much for your feedback

scottgonzalez:
Mostly the `addEvent` and `removeEvent` logic. See https://github.com/jquery/qunit/pull/791#discussion_r26958221 for related discussion.

PR #785, opened 2015-03-17
Split core.js into smaller chunks

BraulioVM:
This PR tries to close #782.

I make the pull request now so that anyone can see how I'm approaching the task and can tell me whether I'm doing something wrong. As it is work in progress, it may not pass the tests

PR #809, opened 2015-04-09
Build: Remove Opera 12.1 from browserstack tests

leobalter:
Although the support is not over, the browserstack tests on that browser
is returning some random errors on Travis-CI, which might be due to
timeout errors, common sometimes, but with a misleading feedback.

Ref #785


jzaefferer:
:+1:

PR #810, opened 2015-04-10
Test: Register source for every assertion

leobalter:
- [ ] missing logs tests for the

Meanwhile, my only concern here is impacting the suite performance. I'll also give a check.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -260,32 +260,22 @@ Test.prototype = {
 	},
 
 	push: function( result, actual, expected, message ) {
-		var source,
-			details = {
-				module: this.module.name,
-				name: this.testName,
-				result: result,
-				message: message,
-				actual: actual,
-				expected: expected,
-				testId: this.testId,
-				runtime: now() - this.started
-			};
+		runLoggingCallbacks( "log", {
+			module: this.module.name,
+			name: this.testName,
+			result: result,
+			message: message,
+			actual: actual,
+			expected: expected,
+			testId: this.testId,
+			runtime: now() - this.started,
+			source: sourceFromStacktrace()
+		} );
 
-		if ( !result ) {
-			source = sourceFromStacktrace();
-
-			if ( source ) {
-				details.source = source;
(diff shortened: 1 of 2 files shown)


jzaefferer:
Seems fine to me, if there is no measurable performance overhead.

leobalter:
fixed tests

leobalter:
On QUnit's test suite there's no impact. I'll need to check this on larger libs, like jQuery UI.

leobalter:
testing jQuery UI:

- current master branch using QUnit 1.18.0: 40078ms
- same master branch using built QUnit from this PR branch: 39718ms

Using `time grunt qunit`:

- QUnit 1.18.0:
```
real	1m13.838s
user	0m36.713s
sys	0m6.223s
```
```
real	1m19.045s
user	0m40.239s
sys	0m6.627s
```

- this PR's branch:
```
real	1m18.697s
user	0m39.693s
sys	0m6.841s
```
```
real	1m15.563s
user	0m38.219s
sys	0m6.546s
```

jzaefferer:
Does phantomjs even support stacktraces? Testing in some actual browsers should be more valuable.

leobalter:
> Does phantomjs even support stacktraces? 

yes. 

I agree about testing on actual browsers is more valuable, but as PhantomJS support stacktrace I thought it was reasonable enough to check for an average difference on a large test suite like the 40s on jQuery UI.

jzaefferer:
Looks like there's still plenty of valid failures on Travis.

leobalter:
After QUnit.stack and reviewing this PR again I believe this is not necessary anymore.

The main reporter doesn't use this information and as the example given at jquery/api.qunitjs.com#106, it's very simple to get this information via QUnit.stack for custom reporters, without adding this extra information when it's not necessary.

The performance wasn't compromised on a larger test suite I used to compare the runtime (jQuery UI) but I still believe this might affect other suites where the performance is critical.

PR #811, opened 2015-04-14
Tests: Remove Opera broken test

leobalter:
Opera 12.1 will not return a proper stacktrace line and will make the following test to fail.

This issue was caught only after TestSwarm run, where Opera 12.1 is returning a different line from the call stack.

Still need to proper fix this error instead of only removing this very specific test.


jzaefferer:
:+1:

leobalter:
I don't know what did by committing that equal assertion, but that wasn't my intention, just garbage from me trying an alternative solution.

I'll wait for another confirmation before landing this.

jzaefferer:
Its a trap? Anyway, as-is it looks good, let's land.

jzaefferer:
Rebased into one commit and merged.

PR #812, opened 2015-04-16
Tests: Avoid loosen errors on autostart test

### LINKED ISSUE #798: Invalid test for autostart

leobalter:
Also remove the QUnit.load as they were invalid.

Closes #798
Closes #803

### DIFF
--- a/browserstack.json
+++ b/browserstack.json
@@ -8,6 +8,7 @@
 		"test/logs.html",
 		"test/setTimeout.html",
 		"test/amd.html",
+		"test/autostart.html",
 		"test/reporter-html/index.html"
 	],
 	"browsers": [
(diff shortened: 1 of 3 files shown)


leobalter:
Worth to mention I re-included them to the brow

jzaefferer:
iOS 8.0 timed out on the added test. Can you verify that it works "in general"? Otherwise looks good.

jzaefferer:
Verified the test passes on iOS 8, merged.

PR #813, opened 2015-04-22
Core: Optimize utilization of global and window variables

leobalter:
Ref #790

QUnit was found exported to the global object on Node environment.

This PR also removes QUnit.diff and QUnit.init functions on Node, they were always browser functions.

This might look as a breaking change, so we should also discuss if this if good to land only on 2.0 or maybe I could set a stub function for them printing a warning about their usage on non-browser envs, which is not functional.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,19 +1,7 @@
-var QUnit,
-	onErrorFnPrev,
-	fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//, "" ),
-	globalStartCalled = false,
-	runStarted = false,
-	Date = window.Date,
-	now = Date.now || function() {
-		return new Date().getTime();
-	},
-	location = window.location || { search: "", protocol: "file:" };
-
-QUnit = {};
-
 QUnit.urlParams = urlParams;
 
-QUnit.isLocal = location.protocol === "file:";
+// Figure out if we're running the tests from a server or not
+QUnit.isLocal = !( defined.document && window.location.protocol !== "file:" );
 
 // Expose the current QUnit version
 QUnit.version = "@VERSION";
@@ -145,38 +133,6 @@ extend( QUnit, {
 });
 
 registerLoggingCallbacks( QUnit );
-
-// `onErrorFnPrev` initialized at top of scope
-// Preserve other handlers
-onErrorFnPrev = window.onerror;
-
-// Cover uncaught exceptions
-// Returning true will suppress the default browser handler,
-// returning false will let it run.
-window.onerror = function( error, filePath, linerNr ) {
-	var ret = false;
-	if ( onErrorFnPrev ) {
-		ret = onErrorFnPrev( error, filePath, linerNr );
-	}
-
-	// Treat return value as window.onerror itself does,
-	// Only do our handling if not suppressed.
-	if ( ret !== true ) {
-		if ( QUnit.config.current ) {
-			if ( QUnit.config.current.ignoreGlobalErrors ) {
-				return true;
-			}
-			QUnit.pushFailure( error, filePath + ":" + linerNr );
-		} else {
-			QUnit.test( "global failure", extend(function() {
-				QUnit.pushFailure( error, filePath + ":" + linerNr );
-			}, { validTest: true } ) );
-		}
-		return false;
-	}
-
-	return ret;
-};
 
 function begin() {
 	var i, l,
--- a/src/core/initialize.js
+++ b/src/core/initialize.js
@@ -2,8 +2,8 @@ var QUnit = {};
 
 var Date = global.Date;
 var now = Date.now || function() {
-			return new Date().getTime();
-		};
+	return new Date().getTime();
+};
 
 var setTimeout = global.setTimeout;
 var clearTimeout = global.clearTimeout;
--- a/src/core/initialize.js
+++ b/src/core/initialize.js
@@ -1,0 +1,31 @@
+var QUnit = {};
+
+var Date = global.Date;
+var now = Date.now || function() {
+			return new Date().getTime();
+		};
+
+var setTimeout = global.setTimeout;
+var clearTimeout = global.clearTimeout;
+
+// Store a local window from the global to allow direct references.
+var window = global.window;
+
+var defined = {
+	document: window && window.document !== undefined,
+	setTimeout: setTimeout !== undefined,
+	sessionStorage: (function() {
+		var x = "qunit-test-string";
+		try {
+			sessionStorage.setItem( x, x );
+			sessionStorage.removeItem( x );
+			return true;
+		} catch ( e ) {
+			return false;
+		}
+	}() )
+};
+
+var fileName = ( sourceFromStacktrace( 0 ) || "" ).replace( /(:\d+)+\)?/, "" ).replace( /.+\//, "" );
+var globalStartCalled = false;
+var runStarted = false;
--- a/src/core/onerror.js
+++ b/src/core/onerror.js
@@ -1,4 +1,7 @@
-if ( defined.document ) {
+( function() {
+	if ( !defined.document ) {
+		return;
+	}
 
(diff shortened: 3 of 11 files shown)


src/export.js
```diff
@@ -1,5 +1,6 @@
 // For browser, export only select globals
-if ( typeof window !== "undefined" ) {
+if ( window.document && window.HTMLDocument &&
+		window.document instanceof window.HTMLDocument ) {
```
leobalter: this is the most precise check I've got, we might want to simplify if to just check window.document.

leobalter:
I also want to land another commit here in this PR renaming window to global to avoid this kind of confusion.

src/diff.js
```diff
@@ -29,6 +29,10 @@
  *
  * QUnit.diff( "the quick brown fox jumped over", "the quick fox jumps over" ) === "the  quick <del>brown </del> fox jump<ins>s</ins><del>ed</del over"
  */
+
+// Don't load QUnit.diff on non-Browser environments
+if ( typeof window !== "undefined" ) {
```
jzaefferer: I get why we exclude the HTML reporter in node, but how does this make a difference?

src/export.js
jzaefferer: This seems like the wrong approach. We shouldn't alias `global` to `window` in the first place.

src/diff.js
leobalter: when QUnit is not exposed as a global value anymore, the following QUnit.difff line would fail.

jzaefferer: I see. I guess with ES6 modules this problem could go away? If so, let's extend the comment a bit and leave the `if` as-is.

leobalter: I've change the if to check this and also extended it to check if window has truth `document` property (avoiding Node apps with a window variable on the global).

src/export.js
leobalter: It is different now and window is now renamed to global.

src/core.js
```diff
@@ -146,7 +146,7 @@ config.modules.push( config.currentModule );
 // Initialize more QUnit.config and QUnit.urlParams
 (function() {
 	var i, current,
-		location = window.location || { search: "", protocol: "file:" },
+		location = global.location || { search: "", protocol: "file:" },
```
jzaefferer: We should leave this as a check on `window`, since we can assume that the `location` property only exists there, just like we do for `document`.

src/core.js
```diff
@@ -429,18 +429,18 @@ extend( QUnit, {
 
 // `onErrorFnPrev` initialized at top of scope
 // Preserve other handlers
-onErrorFnPrev = window.onerror;
+onErrorFnPrev = global.onerror;
```
jzaefferer: Same as above might apply here - do we really expect `onerror` to be available outside of any browser?

src/diff.js
```diff
@@ -29,6 +29,10 @@
  *
  * QUnit.diff( "the quick brown fox jumped over", "the quick fox jumps over" ) === "the  quick <del>brown </del> fox jump<ins>s</ins><del>ed</del over"
  */
+
+// Don't load QUnit.diff on non-Browser environments
+if ( typeof window !== "undefined" && window.document ) {
```
jzaefferer: This seems strange. Just because `QUnit.diff` currently outputs HTML doesn't mean we can't use it for CLI reporters in the future. Even ignoring that, what's the issue of running the code in node?

jzaefferer: I found the previous discussion about this, inside the outdated discussion blob above. Rereading this:

> when QUnit is not exposed as a global value anymore, the following QUnit.difff line would fail.

Can't we just move the diff module inside the intro/outro closure?

leobalter: > Can't we just move the diff module inside the intro/outro closure?

yes, sure.

I would rather avoid this and apply something solid to QUnit.diff for node environments.

src/core.js
leobalter: good catch, I'll fix it.

src/diff.js
jzaefferer: What do you have in mind?

leobalter: nothing consistent. I'll probably move QUnit.diff back and solve this node diff later.

jzaefferer: Okay, let me know when I can review again.

leobalter:
@jzaefferer after a hard rebasing, I've fixed this branch with some more optimizations.

The commits will be squashed using this (new) PR title.

src/diff.js
```diff
@@ -400,7 +400,7 @@ QUnit.diff = (function() {
         if (longtext.length < 4 || shorttext.length * 2 < longtext.length) {
             return null; // Pointless.
         }
-        dmp = this; // 'this' becomes 'window' in a closure.
+        dmp = window;
```
leobalter: This is wrong. It took me a while to understand the purpose. I'll remove this.

src/core/initialize.js
```diff
@@ -0,0 +1,31 @@
+var QUnit = {};
+
+var Date = global.Date;
+var now = Date.now || function() {
+			return new Date().getTime();
+		};
```
jzaefferer: Reduce indent here and on line above.

src/core/onerror.js
```diff
@@ -0,0 +1,34 @@
+if ( defined.document ) {
```
jzaefferer: Can we wrap this module in a IEFE? Helps ensure `onErrorFnPrev` isn't visible anywhere else.

src/core/utilities.js
```diff
@@ -59,7 +57,7 @@ function extend( a, b, undefOnly ) {
 		if ( hasOwn.call( b, prop ) ) {
 
 			// Avoid "Member not found" error in IE8 caused by messing with window.constructor
-			if ( !( prop === "constructor" && a === window ) ) {
+			if ( !( prop === "constructor" && a === global ) ) {
```
jzaefferer: The comment above says this is a browser specific (IE8) issue. Sounds like using `window` is correct here.

src/core/utilities.js
```diff
@@ -113,11 +111,12 @@ function is( type, obj ) {
...
 var getUrlParams = function() {
-	var i, current,
-		location = window.location || { search: "", protocol: "file:" },
-		params = location.search.slice( 1 ).split( "&" ),
-		length = params.length,
-		urlParams = {};
+	var i, current, location, params, length;
+	var urlParams = {};
+
+	location = window.location;
```
jzaefferer: This can be initialised immediately, no need for separate declaration of the variable.

src/core/utilities.js
```diff
@@ -113,11 +111,12 @@ function is( type, obj ) {
...
+	params = location.search.slice( 1 ).split( "&" );
+	length = params.length;
```
jzaefferer: Also applies to these two, as long as the order stays the same. Just stop using onevar...

jzaefferer:
A few small issues left to address, overall looking good.

src/core/utilities.js
leobalter: `window` is correct here, but this leads to a different check. This block runs on every env and global is set to avoid errors on node.

src/core/utilities.js
```diff
@@ -59,7 +57,7 @@ function extend( a, b, undefOnly ) {
 		if ( hasOwn.call( b, prop ) ) {
 
 			// Avoid "Member not found" error in IE8 caused by messing with window.constructor
-			if ( !( prop === "constructor" && a === window ) ) {
+			if ( prop !== "constructor" || a !== global ) {
```
jzaefferer: > This block runs on every env and global is set to avoid errors on node.

Can you add a comment for that?

leobalter: 0bba6d5

leobalter:
I rebased this branch with master and now it's crashing on the amd tests, I'll check and fix.

leobalter:
Realized the error was on my machine only with a missing node module.

jzaefferer:
Looks good to land.

leobalter:
thanks!

PR #822, opened 2015-05-30
HTML Reporter: HTML reporter enhancements for negative asserts

platinumazure:
HTML reporter will prepend "NOT" to expected value for notEqual, notStrictEqual, notDeepEqual, and notPropEqual assertions.

In addition, notOk assertions are also marked as negative for consistency, though the HTML reporter special-cases that assertion.

### DIFF
--- a/test/reporter-html/index.html
+++ b/test/reporter-html/index.html
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="UTF-8">
 	<title>QUnit HTML Reporter Test Suite</title>
-	<link rel="stylesheet" href="../dist/qunit.css">
+	<link rel="stylesheet" href="../../dist/qunit.css">
 	<script src="../../dist/qunit.js"></script>
 	<script src="reporter-html.js"></script>
 	<script src="../swarminject.js"></script>
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -652,11 +652,11 @@ QUnit.log(function( details ) {
 	// when it calls, it's implicit to also not show expected and diff stuff
 	// Also, we need to check details.expected existence, as it can exist and be undefined
 	if ( !details.result && hasOwn.call( details, "expected" ) ) {
-        if ( details.negative ) {
-            expected = escapeText( "NOT " + QUnit.dump.parse( details.expected ) );
-        } else {
-            expected = escapeText( QUnit.dump.parse( details.expected ) );
-        }
+		if ( details.negative ) {
+			expected = escapeText( "NOT " + QUnit.dump.parse( details.expected ) );
+		} else {
+			expected = escapeText( QUnit.dump.parse( details.expected ) );
+		}
 
 		actual = escapeText( QUnit.dump.parse( details.actual ) );
 		message += "<table><tr class='test-expected'><th>Expected: </th><td><pre>" +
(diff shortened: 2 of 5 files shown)


platinumazure:
I signed the CLA a little bit ago. How do I get the error to go away?

And I'm looking for guidance on how best to test the HTML reporter part itself. To be honest it seems like that code could be refactored (i.e., there would be an HTMLReporter object that knows how to spit out the appropriate HTML given a logging context, and that could be unit tested independent of the actual HTML test page.) but that is a separate issue.

jzaefferer:
Thank you for the contribution!

Regarding the CLA: It looks like you signed twice, the first time the email didn't match your git email. I'm not really sure why the check still fails, though. We currently don't have a way to recheck manually, so the only option is to add another (empty) commit to this PR.

Regarding the Travis failure, you're right that its not related to your changes. I haven't seen that particular failure before...

As for the actual changes: Adding another argument to the already long `push` method is bad, but I also don't see an alternative here. Changing the signature to use an object argument would break any custom assertions.

@leobalter what do you think?

platinumazure:
Could we break that contract in 2.0?

jzaefferer:
Kind of. We'd have to deprecate the current method, add a new method in a 1.x release, then remove the deprecated method in 2.0 (same process es Emberjs is following).

If we came up with a better name, that effort would be easier to justify.

platinumazure:
Could I get some guidance on how to add an empty commit to my PR (to re-trigger the CLA and Travis checks)? Thanks!

jzaefferer:
`git commit --allow-empty` should do.

platinumazure:
@jzaefferer Thanks, that's done and now I'm just waiting for Travis to run.

platinumazure:
Looks like the CLA is good but the Travis build failed for the same reason as before.

Seems like a browserstack issue maybe?

leobalter:
@platinumazure the CI fail was due to a browserstack-runner instability. It's not related to your code.

@jzaefferer:
> Adding another argument to the already long push method is bad, but I also don't see an alternative here. 

Agreed. We cannot pass an object with the arguments to push as its first argument is always coerced to a boolean value. That's what we need to deprecate, forcing all push calls to have `result` (the first argument) as a boolean. Further, we'll be able to call `.push` with a single object argument, and that will be more cool in a later future using ES6: `.push({ result, actual, expected, message, negative })`.

test/reporter-html/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="UTF-8">
 	<title>QUnit HTML Reporter Test Suite</title>
-	<link rel="stylesheet" href="../dist/qunit.css">
+	<link rel="stylesheet" href="../../dist/qunit.css">
```
leobalter: nit.

leobalter:
LGTM.

test/reporter-html/index.html
platinumazure: Do you want me to remove that? I couldn't run the QUnit HTML Reporter suite without that change.

leobalter: You don't need to remove that. That was something I didn't notice before because I'm running tests only from the cli (grunt and browserstack).

jzaefferer:
Landed this with a rebase. I've filed #827 for a new push method, or at least a new signature.

platinumazure:
PR status is "Closed with unmerged commits". I assume maybe some cherry-picking happened (e.g., my empty commit to re-trigger CLA and Travis was intentionally ignored) but I want to be sure nothing important was missed.

@jzaefferer have I got it? If so I can delete my source branch. Thanks!

jzaefferer:
Yes, I've rebased your commits. You can find a reference to the commit that landed above.

Issue #833, opened 2015-07-12
Testing Maps/Sets with QUnit


godchuanz:
Most browsers already implement ES6 Map/Set to some degree. Right now, we have the strange issue of:

        obj1 = new Map(); obj1.set(1,2);
        obj2 = new Map(); obj2.set(3,4);
        assert.deepEqual(obj1, obj2);    // passed

I understand QUnit is not meant to process ES6 Map/Set or iterables yet, but as browsers continue to implement ES6 features, this can actually become a real problem.

leobalter:
on [Weak]{Map,Set}, list items doesn't become properties of the collection instantiated object.

deepEqual checks for the values equality of 2 objects' own and extended (prototype) properties. 

so `assert.deepEqual(obj1, obj2);` will indeed pass.

What we need is not break `deepEqual`, but **maybe extend QUnit to compare ES6 collections**. This should regard it's impossible to compare WeakMap and WeakSet collections if you don't know what they in their collection as they doesn't have a `size` or `forEach` methods and [can't be cast to an Array](http://jsbin.com/datalapavo/edit?js,output).

This leads to another topics: 

- should a assertion to compare ES6 collections be shipped with QUnit or should it come as a plugin?
- we can improve our diff on fails to specify JS native collections instead of just `{}`.

godchuanz:
The way I see it, `deepEqual` should be extended to cater for ES6 Set and Map.From the src, it already specially catered for "array" objects. What needs to be done is to update the code in `equiv()` for "arrays" to use `forEach()`, and for Set/Map to be classified under that switch case. For better code clarity, "arrays" could be renamed "collections" or something.

jzaefferer:
Extending `equiv()` to support `Map` and `Set` seems fine, as long as that is backwards compatible. @godchuanz could you look into putting a patch together for that, and send it as a pull request?

As for WeakMap and WeakSet: If those are uncomparable, we could either consider them unequal or throw an exception. Seems like a separate issue.

For both it might be good to see if other frameworks already implement support for any of these new types. Maybe we can just use their approach.

godchuanz:
Hi, yes. I think I have gotten something together. Can you add me as a collaborator so I can make a pull request?

leobalter:
You can open a PR without being a collaborator, just push to your fork.

godchuanz:
Ok done that. Sorry, not used to forking (still noob to github, especially for jQuery). Hopefully the CLA check and CI passes soon.

[...]

PR #834, opened 2015-07-14
Added ES6 Set and Map support

godchuanz:
The main idea is to reduce ES6 Set and Map comparisons into Array comparisons.

Ref: https://github.com/jquery/qunit/issues/833

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -129,12 +129,12 @@ QUnit.equiv = (function() {
 				
 				"set": function( b, a ) {
 					var aArray, bArray;
-					
+
 					// b could be an object literal here
 					if ( QUnit.objectType( b ) !== "set" ) {
 						return false;
 					}
-					
+
 					// reduce sets to arrays
 					aArray = [];
 					a.forEach( function( v ) {
@@ -144,18 +144,18 @@ QUnit.equiv = (function() {
 					b.forEach( function( v ) {
 						bArray.push(v);
 					});
-					
+
 					return innerEquiv( bArray, aArray );
 				},
 				
 				"map": function( b, a ) {
 					var aArray, bArray;
-					
+
 					// b could be an object literal here
 					if ( QUnit.objectType( b ) !== "map" ) {
 						return false;
 					}
-					
+
 					// reduce sets to arrays
 					aArray = [];
 					a.forEach( function( v, k ) {
@@ -165,7 +165,7 @@ QUnit.equiv = (function() {
 					b.forEach( function( v, k ) {
 						bArray.push([k, v]);
 					});
-					
+
 					return innerEquiv( bArray, aArray );
 				},
 
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -126,6 +126,48 @@ QUnit.equiv = (function() {
 					parentsB.pop();
 					return true;
 				},
+				
+				"set": function( b, a ) {
+					var aArray, bArray;
+					
+					// b could be an object literal here
+					if ( QUnit.objectType( b ) !== "set" ) {
+						return false;
+					}
+					
+					// reduce sets to arrays
+					aArray = [];
+					a.forEach( function( v ) {
+						aArray.push(v);
+					});
+					bArray = [];
+					b.forEach( function( v ) {
+						bArray.push(v);
+					});
+					
+					return innerEquiv( bArray, aArray );
+				},
+				
+				"map": function( b, a ) {
+					var aArray, bArray;
+					
+					// b could be an object literal here
+					if ( QUnit.objectType( b ) !== "map" ) {
+						return false;
+					}
+					
+					// reduce sets to arrays
+					aArray = [];
+					a.forEach( function( v, k ) {
+						aArray.push([k, v]);
+					});
+					bArray = [];
+					b.forEach( function( v, k ) {
+						bArray.push([k, v]);
+					});
+					
+					return innerEquiv( bArray, aArray );
+				},
 
 				"object": function( b, a ) {
 
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -130,7 +130,7 @@ QUnit.equiv = (function() {
 				"set": function( b, a ) {
 					var aArray, bArray;
 
-					// b could be an object literal here
+					// b could be any object here
 					if ( QUnit.objectType( b ) !== "set" ) {
 						return false;
 					}
@@ -150,7 +150,7 @@ QUnit.equiv = (function() {
 				"map": function( b, a ) {
 					var aArray, bArray;
 
-					// b could be an object literal here
+					// b could be any object here
 					if ( QUnit.objectType( b ) !== "map" ) {
 						return false;
 					}
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -126,7 +126,7 @@ QUnit.equiv = (function() {
 					parentsB.pop();
 					return true;
 				},
-				
+
 				"set": function( b, a ) {
 					var aArray, bArray;
 
@@ -146,7 +146,7 @@ QUnit.equiv = (function() {
 
 					return innerEquiv( bArray, aArray );
 				},
-				
+
 				"map": function( b, a ) {
 					var aArray, bArray;
 
@@ -157,11 +157,11 @@ QUnit.equiv = (function() {
 
 					aArray = [];
 					a.forEach( function( v, k ) {
-						aArray.push([k, v]);
+						aArray.push( [ k, v ] );
 					});
 					bArray = [];
 					b.forEach( function( v, k ) {
-						bArray.push([k, v]);
+						bArray.push( [ k, v ] );
 					});
 
 					return innerEquiv( bArray, aArray );
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -135,7 +135,6 @@ QUnit.equiv = (function() {
 						return false;
 					}
 
-					// reduce sets to arrays
 					aArray = [];
 					a.forEach( function( v ) {
 						aArray.push(v);
@@ -156,7 +155,6 @@ QUnit.equiv = (function() {
 						return false;
 					}
 
-					// reduce sets to arrays
 					aArray = [];
 					a.forEach( function( v, k ) {
 						aArray.push([k, v]);
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -1531,24 +1531,44 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
 
 QUnit.module( "equiv Maps and Sets");
 
-// - in IE 11, QUnit.objectType( new Set() ) === "object"
 var hasES6Set = (function() {
 	if ( typeof Set !== "function" ) {
 		return false;
 	}
 
-	var s = new Set();
-	return ( QUnit.objectType( s ) === "set" );
+	try {
+		// some platforms don't support iterables in Set constructors
+		var s = new Set([ 1, 2, 3 ]);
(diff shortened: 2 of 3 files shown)


src/equiv.js
```diff
@@ -126,6 +126,48 @@ QUnit.equiv = (function() {
 					parentsB.pop();
 					return true;
 				},
+				
+				"set": function( b, a ) {
+					var aArray, bArray;
+					
+					// b could be an object literal here
```
leobalter: actually b could be any other kind of object.

This might be removed, its clear we are checking both arguments are both instances of Set.

src/equiv.js
```diff
@@ -126,6 +126,48 @@ QUnit.equiv = (function() {
...
+						bArray.push(v);
+					});
+					
+					return innerEquiv( bArray, aArray );
+				},
+				
+				"map": function( b, a ) {
+					var aArray, bArray;
+					
+					// b could be an object literal here
```
leobalter: same thing here

src/equiv.js
```diff
@@ -126,6 +126,48 @@ QUnit.equiv = (function() {
...
+					if ( QUnit.objectType( b ) !== "map" ) {
+						return false;
+					}
+					
+					// reduce sets to arrays
```
leobalter: this is not a set. Also, this comment might be deleted.

leobalter:
Beyond the comments, LGTM

src/equiv.js
godchuanz: I was actually hesitant to add this in, but I saw that the "array" case had this liner. I didn't quite understand why "array" had it, but I thought there's no harm checking it other than slight inefficiency. If it's indeed useless, I can remove it.

godchuanz:
I have no idea why this is failing CI. Looks like an unrelated problem.

Same with CLA. I've signed and ensure the name and email are right...

src/equiv.js
godchuanz: Inspecting the code further, I think the check is necessary, since `objectType()` is only run on `a`, and never on `b` prior to executing the function `callbacks['set']` / `callbacks['map']`.

godchuanz: On further inspection and tracing of the code, I believe it's impossible that a Set and a "non-Set" ends up in the switch case, as `a` is guaranteed to be same type as `b` (condition on line 245 of equiv.js). This begs the question: why does the "array" case include this check?

jzaefferer:
The CLA signature looks fine, no idea why the automatic check is failing.

For now, could you rebase your PR on top of master? We landed a bigger refactoring there that we couldn't hold off any longer.

godchuanz:
Done. If everything is fine, please just merge this PR in despite the CLA.

jzaefferer:
This PR/branch looks very messed up now. I could do a `--squash` merge to keep the actual map/set changes, but that still leaves us without any tests - I'm not sure why neither Leo nor me noticed that before.

Are there polyfills for map/set? Maybe we can use one to have tests run in all browsers, not just the ones with native support.

leobalter:
> Are there polyfills for map/set? Maybe we can use one to have tests run in all browsers, not just the ones with native support.

I would say https://github.com/paulmillr/es6-shim is the best tested polyfill option, but I believe it's best to just check the existence of Map and Set objects, and use `skip` if they're absent.

@godchuanz: I can handle the squash, but I would appreciate if we have the tests on `test/main/deepEqual.js`.

the if condition could be like: 

```js
var hasMap = typeof Map === "function";
var hasSet = typeof Set === "function";

QUnit.module( "equiv Maps and Sets" );
QUnit[ hasMap ? "test" : "skip" ]( "Maps", function( assert ) {
    var m1 = new Map( [ [ 1, "foo" ], [ 2, "bar" ] ] );
    var m2 = new Map( [ [ 1, "foo" ], [ 2, "bar" ] ] );
    var m3 = new Map( [ [ 1, "foo" ] ] );

    assert.deepEqual( m1, m2 );
    assert.notDeepEqual( m1, m3 );

    ...
} );

...
```

scottgonzalez:
> The CLA signature looks fine, no idea why the automatic check is failing.

"godchuanz" is not a valid name.

godchuanz:
@leobalter I'm adding in the tests for Sets/Maps, but I noticed something strange in the process of the tests.

IE11 seems to report Sets and Maps as `"object"` when you run `QUnit.objectType()`, e.g.

    var s = new Set();
    var type = QUnit.objectType( s ); // returns "object"
    var isSet = s instanceof Set;        // returns true

Because IE's implementation of Set/Map is also incomplete (does not support iterators, e.g. Array, `arguments` as parameters to constructor), I think it's best to consider IE as having no support for ES6 Set/Map.

If a proper polyfill is included, IE11 would run and pass those tests.

godchuanz:
@scottgonzalez Hi, why is it not a valid name? :(

scottgonzalez:
Because the CLA checker expects a full name to contain at least one space. Is godchuanz your legal name?

godchuanz:
@scottgonzalez  No it's not. I think it's gonna be tedious to change the name of all the commits. I think I'll just leave it to the owners to merge this PR in if they think it's fine.

leobalter:
@godchuanz:

> I think I'll just leave it to the owners to merge this PR in if they think it's fine.

@scottgonzalez is an owner.

The CLA check is a necessary step on any project from the jQuery Foundation.

We need you to set it with valid values, you can see a quick how to here: http://contribute.jquery.org/CLA/status/?repo=qunit&sha=413d5150cf31f976621e6bca26c72c19e1474963

godchuanz:
@leobalter @scottgonzalez Ok, I have added in real names and rebased the entire branch's commits. CLA is now passing.

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+		if (s.size !== 3 || !s.has(2)) {
+			return false;
+		}
+	}
+	catch (e) {
+		return false;
+	}
+
+	return true;
+})();
```
leobalter: These flags should be conditioned to the same kind of test from QUnit.objectType, so if Set is existent, callable with a new target and `(new Set()).toString()` returns `"[object Set]"`, we can assume ES6's Set is implemented. Same thing on Map.

This way, we can also reduce the comments to say only `in IE 11, QUnit.objectType( new Set() ) === 'object'`.

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+QUnit[ hasES6Set ? "test" : "skip" ]("Sets", function ( assert ) {
+	var s1, s2, s3, s4, o1, o2, o3, m1, m2, m3;
+
+	// Empty sets
+	s1 = new Set();
+	s2 = new Set();
+	assert.equal( QUnit.equiv( s1, s2 ), true, "Empty Sets");
+	s1 = new Set([]);
+	s2 = new Set([]);
+	assert.equal( QUnit.equiv( s1, s2 ), true, "Empty sets init with empty arrays");
```
leobalter: `new Set();` and `new Set([]);`

Both returns empty Set objects, this is tested on ES6. We may replace this by only: 

```js
s1 = new Set();
s2 = new Set( [] );
assert.equal( QUnit.equiv( s1, s2 ), true, "Empty Sets" );
```

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	s1 = new Set();
+	s1.add( undefined );
+	s1.add( null );
+	s1.add( false );
+	s1.add( 0 );
+	s1.add( NaN );
+	s1.add( Infinity );
+	s1.add( -Infinity );
+	s2 = new Set([ undefined, null, false, 0, NaN, Infinity, -Infinity ]);
+	assert.equal( QUnit.equiv( s1, s2 ), true, "Mutiple-element sets of tricky values");
```
leobalter: also, there's no need to test Set.prototype.add and Set elements added via the constructor's iterable argument.

We can merge assertions from ```// Simple cases``` to ```// Sets containing sets```

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	// Sets containing maps
+	m1 = new Map([ [ 1,1 ] ]);
+	m2 = new Map([ [ 1,1 ] ]);
+	m3 = new Map([ [ 1,3 ] ]);
+	s3 = new Set([ m1 ]);
+	s4 = new Set([ m2 ]);
+	assert.equal( QUnit.equiv( s3, s4 ), true, "Sets containing different but deeply-equal maps");
+	s3 = new Set([ m1 ]);
+	s4 = new Set([ m3 ]);
+	assert.equal( QUnit.equiv( s3, s4 ), false, "Sets containing different maps");
```
leobalter: I would leave assertions for nested Set and Map objects to other blocks of tests.

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+});
+
+QUnit[ hasES6Map ? "test" : "skip" ]("Maps", function ( assert ) {
```
leobalter: There are some minor whitespace issues, too: ```("Maps"``` => ```( "Maps"```

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	var m1, m2, m3, m4, o1, o2, o3, s1, s2, s3;
+
+	// Empty maps
+	m1 = new Map();
+	m2 = new Map();
+	assert.equal( QUnit.equiv(m1, m2), true, "Empty maps");
+	m1 = new Map([]);
```
leobalter:
```js
assert.equal( QUnit.equiv(m1, m2), true, "Empty maps");
m1 = new Map([]);
```
to:

```js
assert.equal( QUnit.equiv( m1, m2 ), true, "Empty maps" );
m1 = new Map( [] );
```

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	m2 = new Map([
+		[ 1, o1 ],
+		[ 2, o3 ]
+	]);
+	assert.equal( QUnit.equiv( m1, m2 ), true, "Maps containing same objects");
+	m1 = new Map([
+		[ 1, o1]
+	]);
+	m2 = new Map([
+		[ 1, o2]
```
leobalter: missing space before the square bracket

test/main/deepEqual.js
```diff
@@ -1528,3 +1528,226 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	
+	// Maps containing different sets
+	m1 = new Map([
+		[1, s1]
+	]);
+	m2 = new Map([
+		[1, s3]
+	]);
+	assert.equal( QUnit.equiv( m1, m2 ), false, "Maps containing diffrent sets");
+});
```
leobalter: missing eof

test/main/deepEqual.js
godchuanz: I think Sets containing numbers (in the `// Simple cases`) should not be conflated with Sets containing objects, including Sets (in the `// Sets containing sets`).

leobalter:
Thanks, @godchuanz! I've rebased your branch and removed minor whitespace that remained. It's on master now and will land within the next release.

godchuanz:
@leobalter that's great +1

Issue #815, opened 2015-04-25
browserstack-runner improvements


jzaefferer:
Currently we run all our testsuites against most of our supported browsers for PRs and commits to master. With plenty of failures due to timeouts, this doesn't provide near as much value as it should. I'd like to use this issue to discuss some options for getting more values from this setup.

One idea: Only run the main tests for "all" commits, then set up a separate browserstack.json (or browserstack.js, might work) that runs all suites in (more|all) browsers, but only for some commits, e.g., all commits to master. The latter could also do multiple runs, each running just one of the testsuites, if that helps avoid the timeouts.

Meanwhile we should use https://browserstack.com/automate to track failing tests, check which ones take forever to run (causing later suites to time out, since the 5 minute window is over), along with any other failures, and consistently report those back to support@browserstack.com. If we keep ignoring these issues, we won't make any progress.

PR #836, opened 2015-07-17
Build: Update browserstack settings

leobalter:
This update remove some browsers from the default set in order
to avoid timeout errors.

It also includes the Microsoft Edge browser.

Ref #815 

I'll also use this PR to check for possible timeouts.


browserstack-all.json
```diff
@@ -0,0 +1,50 @@
...
+		},
+		{
+			"os": "iOS",
+			"os_version": "8.3"
+		},
+		{
+			"os": "Windows",
+			"os_version": "10",
+			"browser": "edge",
+			"browser_version": "0.11"
```
mzgol: I wouldn't specify this if not needed; Edge build numbers may change quite frequently.

browserstack-all.json
```diff
@@ -0,0 +1,50 @@
...
+			"os": "iOS",
+			"os_version": "7.0"
+		},
+		{
+			"os": "iOS",
+			"os_version": "8.3"
+		},
+		{
+			"os": "Windows",
+			"os_version": "10",
```
mzgol: Edge implies Windows 10 so this shouldn't be needed either. Can't you just write `edge_current`?

browserstack.json
```diff
@@ -29,16 +27,10 @@
...
-			"os": "iOS",
-			"os_version": "7.0"
-		},
-		{
-			"os": "iOS",
-			"os_version": "8.3"
+			"os": "Windows",
+			"os_version": "10",
+			"browser": "edge",
+			"browser_version": "0.11"
```
mzgol: See above remarks

browserstack-all.json
leobalter: There isn't an alias for Edge yet.

Setting it with the alias or omitting one of these properties will make browserstack to break and close silently.

mzgol: Ah, I see.

leobalter:
Just a note: this pr doesn't update browserstack-runner to 0.3.4 anymore as I've got some issues with it. I'll need to reflect this on the commit message

jzaefferer:
@leobalter afaik BrowserStack fixed some issue with Edge in the API, can you update this to address @mzgol's comments?

jzaefferer:
Also 0.3.5 is out and works fine on my machine. Worth a try here as well.

leobalter:
trying browserstack 0.3.6

leobalter:
@jzaefferer browserstack-runner is breaking on the newer versions. I'll isolate this problem to another issue.

leobalter:
@mzgol I was able to use edge_latest, but `edge_current` and just `edge` breaks browserstack-runner (from 0.3.0 to 0.3.6)

PR #839, opened 2015-07-28
Fix bug while no document(nodejs, react-native)

steel1990:
document.createElement('span') will throw an error while in nodejs or react-native.

so I move that line of code to the below.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -392,12 +392,13 @@ function toolbarModuleFilterHtml() {
 
 function toolbarModuleFilter() {
 	var toolbar = id( "qunit-testrunner-toolbar" ),
-		moduleFilter = document.createElement( "span" ),
 		moduleFilterHtml = toolbarModuleFilterHtml();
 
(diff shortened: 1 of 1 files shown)


jzaefferer:
Can you please sign our CLA? http://contribute.jquery.org/CLA/

leobalter:
@steel1990, your changes looks good, but I need you to sign the CLA before I merge.

jzaefferer:
@leobalter can you apply this patch manually, thanks the author in the commit message?

leobalter:
yes.

leobalter:
I was working on this but with the html reporter preventing to run on an [env without a `window.document` object](https://github.com/jquery/qunit/blob/master/reporter/html.js#L4-L6) I wonder if this suggesting isn't related to an older version of QUnit.

Also eaae69b might also solve the problem.

As I don't have any feedback from the author, I'll close this assuming this is not an issue anymore.

Issue #844, opened 2015-08-06
Expected at least one assertion, but none were run - Bounty Available


mattparlane:
This is one of those annoying intermittent issues...  I have a test which has two async calls in series, and I'm getting this error around 30% of the time:

![screen shot 2015-08-07 at 8 17 54 am](https://cloud.githubusercontent.com/assets/176572/9122419/fbb26b9a-3cdc-11e5-9a47-36367d4014e4.png)

This is the test:

    QUnit.test 'Backbone save - new object has attributes', (assert) ->
    	done = assert.async()
    	search = new Search
    	search.save(Description: 'My test').done ->
    		testSearch = new Search(id: search.id)
    		testSearch.fetch().done ->
    			assert.equal(testSearch.get('Description'), 'My test')
    			done()

Unfortunately the save/fetch calls depend on a server so it's not easily reproducible. I've tried creating a reproducible test by using setTimeout instead of the save/fetch calls but it works as expected.

What I've found is that I can make the problem go away by increasing the 13ms timeout in resumeProcessing(), the errors get fewer and fewer as the timeout gets higher, until they seem to completely stop at around 100ms.

Any ideas?

leobalter:
I don't know if this code is CoffeeScript or anything else different from JS, but this line doesn't look right:

```
search.save(Description: 'My test').done ->
```

As you told, I've also failed trying to reproduce it with setTimeout: http://jsbin.com/suxuka/edit?js,output

Maybe these .done functions are not being triggered.

mattparlane:
They are definitely being triggered -- I've added console.log statements inside them and they are there. I used console.logs to get the timing in my attempt to reproduce it with setTimeout, but that didn't work.

Is there any chance that the 13ms value is incorrect? I don't really understand what that code is doing, but like I said, raising it to 100 fixes my problem. Perhaps it could be made an overridable setting?

mattparlane:
Also, the CoffeeScript code is correct, search.save() is a Backbone method which returns a jQuery XHR, so .done() is the same function you'd call if you were doing `$.get(...).done(...)`.

Also, there's no confusion between the local variable done and the method name done, that was just bad naming on my part -- I tried changing the variable name even though I knew it was nothing to do with it, but still no joy.

mattparlane:
I now have a reproducible test! It still depends on a server but should be trivial to set up. I have it visible here:

https://admin.wb.gs/test.html

[...]

PR #845, opened 2015-08-07
Test: Don't pass Promise fulfillment value to QUnit.start

gibson042:
Closes gh-844


leobalter:
LGTM

The CI timeouts were due to browserstack issues, confirmed on a local run for the same envs.

PR #849, opened 2015-08-10
Build: Update JSHint and remove testSwarm

leobalter:
This is a workaround fix for our testSwarm checks which are using an updated version of JSHint from grunt-contrib-jshint and failing on some tests.

Ref http://jenkins.jquery.com/job/QUnit/545/console

---

Edit: 

This PR also removes the testSwarm job (as seen on the following comments).


Gruntfile.js
```diff
@@ -1,4 +1,5 @@
 /*jshint node:true */
+/*globals JSON:false*/
```
leobalter: this was necessary as we have the `es3: true` setting on `.jshintrc`

jzaefferer:
Looks good. 

At this point, I think we can turn off the TestSwarm job for QUnit, since browserstack-runner works reasonably well. What do you think?

leobalter:
Agreed. I'll do this.

jzaefferer:
Changes look good. Can't check right now if you've missed anything in this repo. One external change requires Jenkins access: Remove the part that submits the Jenkins job, only the CDN deploy should be left then.

leobalter:
I checked with some `git grep -i`s for `testSwarm` and `swarm`. From the code I believe it's done.

I'm not sure if those other parts you mentioned prevents this to merge.

jzaefferer:
No blockers, just something to do now or as soon as possible. Lots of people have Jenkins access and should be able to help.

jzaefferer:
I've just removed the grunt-testswarm call from the Jenkins job, back to green: http://jenkins.jquery.com/job/QUnit/548/

mzgol:
@jzaefferer should I remove TestSwarm QUnit configuration from the infra repo as well?

jzaefferer:
@mzgol yes, thanks.

mzgol:
@jzaefferer Done.

PR #850, opened 2015-08-11
QUnit.diff cleanup

leobalter:
Trying to cleanup QUnit.diff, goal is to make it easier to maintain.

### DIFF
--- a/test/reporter-html/diff.js
+++ b/test/reporter-html/diff.js
@@ -1,0 +1,163 @@
+QUnit.module( "diff" );
+
+QUnit.test( "throws if arguments are not strings", function( assert ) {
+	assert.throws(function() { QUnit.diff( {}, "" ); });
+	assert.throws(function() { QUnit.diff( "", {} ); });
+
+	assert.throws(function() { QUnit.diff( [], "" ); });
+	assert.throws(function() { QUnit.diff( "", [] ); });
+
+	assert.throws(function() { QUnit.diff( undefined, "" ); });
+	assert.throws(function() { QUnit.diff( "", undefined ); });
+
+	assert.throws(function() { QUnit.diff( null, "" ); });
+	assert.throws(function() { QUnit.diff( "", null ); });
+
+	assert.throws(function() { QUnit.diff( 42, "" ); });
+	assert.throws(function() { QUnit.diff( "", 42 ); });
+
+	assert.throws(function() { QUnit.diff( true, "" ); });
+	assert.throws(function() { QUnit.diff( "", true ); });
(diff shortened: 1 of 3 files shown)


test/reporter-html/diff.js
```diff
@@ -0,0 +1,163 @@
+QUnit.module( "diff" );
```
jzaefferer: Why is this in the `reporter-html` folder? The relevant code isn't attached to the reporter.

test/reporter-html/diff.js
```diff
@@ -0,0 +1,163 @@
...
+
+QUnit.test( "throws if arguments are not strings", function( assert ) {
+	assert.throws(function() { QUnit.diff( {}, "" ); });
+	assert.throws(function() { QUnit.diff( "", {} ); });
```
jzaefferer: I think these two are enough to verify that there is some argument checking in place, no need for further tests.

test/reporter-html/diff.js
```diff
@@ -0,0 +1,163 @@
...
+});
+
+QUnit.test( "different strings", function( assert ) {
+	var a = "abcd";
+	var b = "xkcd";
+
+	assert.equal(
+		QUnit.diff( a, b ),
+		"<del>ab</del><ins>xk</ins><span>cd</span>",
+		"QUnit.diff( a, b )"
```
jzaefferer: This doesn't seem like a particularly useful message. Also applies to the others.

jzaefferer:
I've reviewed all commits individually (the only sane option here, I think). A few minor notes, looking very good overall.

leobalter:
Thanks, I'll fix it when I get at home

test/reporter-html/diff.js
leobalter: Our current diff is only relevant to the HTML reporter as its output is HTML. These tests intend to progress the coverage on the diff functions, but they are meant for what's expected on the reporter, meaning we could change it for any other library in a future.

test/reporter-html/diff.js
leobalter: A practice I got after test262 where all cases were relevant, but that's about a language test. :)

leobalter:
Waiting feedback to decide on https://github.com/jquery/qunit/pull/850#discussion-diff-37100759 but everything is fixed

test/reporter-html/diff.js
jzaefferer: Okay, we can leave it here.

jzaefferer:
Good to land.

PR #852, opened 2015-08-20
Readme: Replace Jenkins badge with Travis

Krinkle:
Replace badge from jenkins.jquery.org with badge from Travis CI

Follows-up d1564e1a47 and pull #849, which remove TestSwarm from the Jenkins run.
However that means the primary badge isn't very useful anymore and doesn't
include cross-browser results. It also seems that the Jenkins badge plugin doesn't
handle proxy caching properly, causing it to show as "Unstable" on GitHub long
after the build was fixed.

Update both badges to use SVG so that it doesn't look pixelated on high density screens.


leobalter:
LGTM, just want to confirm with @jzaefferer if "readme" is the proper components to use on the commit message.

From the valid names listed at package.json, looks like `All` would be the best option, but it seems weird for a change on the readme file.

Krinkle:
Previous commits to the readme file also used that prefix (see [history](https://github.com/jquery/qunit/commits/master/README.md)). I just copied the prefix from there.

I guess those commits are from before the entry in package.json. I've updated it accordingly.

jzaefferer:
Looks good to me.

PR #853, opened 2015-08-24
Build: Enable IRC notifications for Travis CI

Krinkle:
Follows-up 0d52b8d, and d1564e1.

Also move 'sudo' to be on top because it controls which container is used, more hierarchical in order.


jzaefferer:
Looks good to me.

PR #856, opened 2015-09-07
Core: Equivalency for desdendants of null constructors

### LINKED ISSUE #851: deepEqual fails to work with descendants of something with a null prototype
the following works as expected.

```js
var a = Object.create(null);
a.foo = 1;
var b = { foo: 1 }

deepEqual(a, b);
```

Unfortunately, in practice `Object.create(null)` is fairly costly. To mitigate this cost a specialized `NullObject` can be used. This Object has similar safe characteristics, but with dramatically reduced allocation costs.

#### Reference implementation:

```
function NullObject() { }
NullObject.prototype = Object.create(null, {
  constructor: {
    value: undefined,
    enumerable: false,
    writable: true
  }
});
```

### This leads us to the problem:

```js
var a = new EmptyObject();
a.foo = 1;
var b = { foo: 1 }

deepEqual(a, b); // fails
```

This fails because, `a.constructor !== b.constructor` and `a.prototype !== null`

relevant code in [QUnit](https://github.com/jquery/qunit/blob/d359314e0ad6a21d998bed450f6fc289c5be3490/src/equiv.js#L183-L188):
```js
if ( a.constructor !== b.constructor ) {

  // Allow objects with no prototype to be equivalent to
  // objects with Object as their constructor.
  if ( !( ( getProto( a ) === null && getProto( b ) === Object.prototype ) ||
    ( getProto( b ) === null && getProto( a ) === Object.prototype ) ) ) {
    return false;
  }
}
```

As QUnit already special cases `null` prototyped object with projo deepEquality, could we explore another special case to handle this scenario? Is there an alternative I am missing?

### One solution,
would be to also detect a `null` constructor.

change to the NullObject constructor would be as follows:

```
function NullObject() { }
NullObject.prototype = Object.create(null, {
  constructor: {
    value: null, // <-- small change
    enumerable: false,
    writable: true
  }
});
```

Change to QUnit would require also checking for a null constructor.

````js
if ( a.constructor !== b.constructor ) {

  // Allow objects with no prototype to be equivalent to
  // objects with Object as their constructor.
  if ( !( ( ( getProto( a ) === null || getConstructor( a ) === null) && getProto( b ) === Object.prototype ) ||
    ( ( getProto( b ) === null || getConstructor( b ) === null) && getProto( a ) === Object.prototype ) ) ) {
    return false;
  }
}
```

*note: this protocol does appear funky, but would be quite handy in practice*

leobalter:
Closes #851

@stefanpenner Does it fixes the issue on #851?

I made it in several commits in order to allow a step by step review.

### DIFF
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -148,12 +148,13 @@ QUnit.test( "Objects basics", function( assert ) {
 
 });
 
-QUnit.test( "Objects with and without prototypes", function( assert ) {
-	var nonEmptyWithNoProto;
+QUnit[ typeof Object.create === "function" ? "test" : "skip" ]( "Objects with null prototypes",
+	function( assert ) {
+		var nonEmptyWithNoProto;
 
-	// Objects with no prototype, created via Object.create(null), are used e.g. as dictionaries.
-	// Being able to test equivalence against object literals is quite useful.
-	if ( typeof Object.create === "function" ) {
+		// Objects with no prototype, created via Object.create(null), are used
+		// e.g. as dictionaries.
+		// Being able to test equivalence against object literals is quite useful.
 		assert.equal(
 			QUnit.equiv( Object.create( null ), {} ),
 			true,
@@ -169,7 +170,7 @@ QUnit.test( "Objects with and without prototypes", function( assert ) {
 			"object without prototype VS object"
 		);
 	}
-});
+);
 
 QUnit.test( "Arrays basics", function( assert ) {
 
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -161,6 +161,12 @@ QUnit[ typeof Object.create === "function" ? "test" : "skip" ]( "Objects with nu
 			"empty object without prototype VS empty object"
 		);
 
+		assert.equal(
+			QUnit.equiv( {}, Object.create( null ) ),
+			true,
+			"empty object VS empty object without prototype"
+		);
+
 		nonEmptyWithNoProto = Object.create( null );
 		nonEmptyWithNoProto.foo = "bar";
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
> @stefanpenner Does it fixes the issue on #851?

i believe so, will test for you this afternoon.

Question:

* should we actually walk the proto chain all the way to detect the `null` or is 1 step sufficient.

leobalter:
> should we actually walk the proto chain all the way to detect the null or is 1 step sufficient.

I thought about this, but I was being very conservative on this change. 

I can make this change if you believe this might be better.

test/main/deepEqual.js
```diff
@@ -146,16 +146,66 @@ QUnit.test( "Objects basics", function( assert ) {
...
+			value: null
+		}
+	});
+
+	var a = new NullObject();
+	a.foo = 1;
+	var b = { foo: 1 };
+
+	assert.ok( QUnit.equiv( a, b ) );
+	assert.ok( QUnit.equiv( b, a ) );
```
leobalter: @stefanpenner 

Also:

var c = Object.create( a );

c.constructor === null, so maybe the native chain walk also answers your question, right?

jzaefferer:
> I can make this change if you believe this might be better.
> c.constructor === null, so maybe the native chain walk also answers your question, right?

What's the status of this? @stefanpenner could you take another look?

stefanpenner:
> c.constructor === null, so maybe the native chain walk also answers your question, right?

im also not totally sure what is correct here. That being said, i'm inclined to say the chain walk is the better option

leobalter:
I rebased the branch to have a updated CI run.

@stefanpenner I'm willing to do a release on this before #859, I'll squash and land after your ok.

jzaefferer:
Just went through each commit. Both refactorings and the actual change look good to me.

Btw. there was a failure on one of the browserstack-runner runs on Travis, but somehow that didn't cause the build to fail.

stefanpenner:
LGTM

leobalter:
merged via @b8f1926.

Thanks, @jzaefferer and @stefanpenner!

Issue #855, opened 2015-09-04
Safari 7.0 isn't tested


mzgol:
BrowserStack has removed Safari 7.0 from the API, I notified them that it's a problem for us. However, because of [these lines of browserstack-all.json](https://github.com/jquery/qunit/blob/master/browserstack-all.json#L21-L22) you now wouldn't even try to this version; only Safari 5.1, 6.2, 8.0 & 9.0 are included. This means Safari 6.0 & 7.0 is gone (6.2 has an identical engine as 8.0).

Due to the fact of how messed up Safari versioning is:

1. Safari 8.0, 7.1 & 6.2 share the WebKit version
2. Safari 7.0 & 6.1 share the WebKit version

I think it's dangerous to just rely on aliases; it'd be better to explicitly list all Safari versions supported; and do it without shortcuts like `safari_6` but use the full `safari_6_0` name.

leobalter:
Thanks for letting me know.

Does safari_6_2 works? otherwise we'll need to specify each browser without aliases.

mzgol:
I'm not sure but you specify `safari_5_1` already so I assume it should. It's best to include every Safari version since `6.0` with the `_0` suffix as minor updates are basically the same as major ones in this land.

leobalter:
I'll have to test it, browserstack does not document the aliases and some key names does not work.

jzaefferer:
Btw. Safari 7 is back in the API (I think), apparently it won't come back in Live though, so don't expect to see it there.

mzgol:
Yup, I've restored it for projects using TestSwarm.

mzgol:
Please see https://github.com/jquery/jquery/issues/2609. It seems the new Safari 7.0 is a fake one and under the hood it has a 7.1/8.0 engine so don't rely on its results for now. :/ I notified BrowserStack about the issue.

jzaefferer:
Core is not going to test against Safari 7.0 anymore, so we don't have to bother either.

PR #857, opened 2015-09-07
Build: Multiple runs of browserstack on TravisCI

### LINKED ISSUE #854: Split browserstack run on CI
Currently, browserstack-runner is returning timeouts. This is due to the number of different browsers and files.

We can try to split the runs to avoid timeouts. I am thinking about:

1. run the tests only on the current browsers
2. run the tests on the older browsers

Another split if we need to go further:

1. **main** tests only on the current browsers
2. **main** tests on the other browsers
3. additional tests on the current browsers
4. additional tests on the other browsers

leobalter:
~~WIP, this should make multiple runs on browserstack to avoid timeout errors~~

Closes #854 
Ref #855 (Safari 7.0 depends on browserstack support)


browserstack.json
```diff
@@ -4,15 +4,31 @@
...
+		"opera_previous",
 		"opera_current",
+		"safari_5_1",
+		{
+			"browser": "safari",
+			"browser_version": "6.0"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "6.2"
```
mzgol: Safari 6.2, 7.1 & 8.0 have indentical WebKit versions. You'll now test all 3 of them. You might still want that to be sure (Sizzle has this strategy, jQuery Core doesn't), just making sure that's intended. :)

mzgol: Also, note that you're not testing Safari 7.0. This is temporarily impossible due to a mishap on BrowserStack's part but I hope it will be resolved soon.

jzaefferer:
The newly add `BrowserStackLocal` binary shouldn't be there.

browserstack.json
leobalter: Thanks, I was just testing I'll still refine this PR. 

I needed to push some commits to check Travis-CI

leobalter:
> The newly add BrowserStackLocal binary shouldn't be there. 

my bad, now fixed

browserstack.json
```diff
@@ -4,15 +4,31 @@
...
+		{
+			"browser": "safari",
+			"browser_version": "7.0"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "7.1"
+		},
 		"safari_previous",
 		"safari_current",
```
leobalter: @mzgol 
This list run tests on Safari:

- 9 (current)
- 8 (previous)
- 7.1
- 7.0 (same Webkit as 6.2)
- 6.0
- 5.1

review, plz

leobalter:
Now the tests are ordered by a `build/run-browserstack.sh`, ignoring the `./browserstack.json` which is there only for local runs with all the browsers.

Should we delete this root file and run browsertack tests only from the `./build/` files?

browserstack.json
mzgol: 7.0 is the same as 6.1, 8.0 is the same as 6.2. Unfortunately, BrowserStack removed both 7.0 & 6.1, I wrote to them we need those versions (or at least one of them).

leobalter: Browserstack fail with Safari 7.0, so I changed it to Safari 6.2

mzgol: Basically, Safari major number corresponds to the OS X release, minor is bumped when engine is updated. Thus, 6.x is Mountain Lion, 7.x is Mavericks, 8.x is Yosemite, 9.x is El Capitan. 6.2 on Mountain Lion may have a different UI than 8.0 on Yosemite but WebKit used is the same.

mzgol: > Browserstack fail with Safari 7.0, so I changed it to Safari 6.2

They're not equivalent, unfortunately. See my above comments.

mzgol: (also, 7.1 is the same as 8.0 engine-wise)

browserstack.json
```diff
@@ -4,15 +4,31 @@
...
+		},
+		{
+			"browser": "safari",
+			"browser_version": "6.2"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "7.1"
+		},
 		"safari_previous",
```
mzgol: I'd remove `safari_previous`; it will most likely soon match `8.1` when it gets released on Yosemite. Safari versioning scheme makes `safari_previous` useless.

browserstack.json
leobalter: Thanks.

From here is there anything else you recommend to improve the list?

browserstack.json
leobalter: I didn't see this msg before. 

> Safari versioning scheme makes safari_previous useless.

yeap.

leobalter:
> Should we delete this root file and run browsertack tests only from the ./build/ files?

Confirmed with @jzaefferer I'll delete browserstack.json on the root and set a npm script to call the build/run-browsertack.sh

browserstack.json
mzgol: > From here is there anything else you recommend to improve the list?

I need to know your planned strategy first. :) Do you plan to test on various equivalent (WebKit-wise) releases or only on one line for each WebKit version? If the latter then you have a duplication: 6.2, 7.1 & 8.0 (covered currently by `safari_previous`) are the same.

Also, you didn't address my comment: https://github.com/jquery/qunit/pull/857/files#r38875240.

build/browserstack-legacy-1.json
```diff
@@ -0,0 +1,40 @@
...
+			"browser": "safari",
+			"browser_version": "6.0"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "6.2"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "8.0"
```
leobalter: @mzgol I hope I got it now.

I removed Safari `previous` and changed 7.1 to 8.0.

browserstack.json
mzgol: OK, I was commenting an outdated diff. :D Ignore.

build/browserstack-legacy-1.json
```diff
@@ -0,0 +1,40 @@
...
+		"firefox_previous",
+		"opera_previous",
+		"safari_5_1",
+		{
+			"browser": "safari",
+			"browser_version": "6.0"
+		},
+		{
+			"browser": "safari",
+			"browser_version": "6.2"
```
mzgol: This is equivalent to 8.0.

leobalter: :tableflip: 

Thanks :)

jzaefferer:
Looks good to land.

Issue #858, opened 2015-09-10
Nested modules improvements


leobalter:
With #800 done, @gibson042 brought some important topics to improve:

>
* executing all of a module's own tests before entering its submodules (regardless of test/submodule definition order)
* always minting a new environment definition for every submodule, even if its `hooks` argument is `undefined` or not even specified

> If they are intentional, there should be assertions covering them. If they are unintentional and undesired, there should be assertions covering the desired behavior.

Also: 

> the one regarding [explicit context sensitivity of beforeEach/afterEach/test/etc][1].
via arguments or context, which should be covered by your above followup).

[1]: https://github.com/jquery/qunit/pull/670#discussion_r26321413

PR #859, opened 2015-09-10
Nested modules

### LINKED ISSUE #543: Allow nested suites (modules)?

leobalter:
Replaces/Closes #800 
Closes #543

This PR closes #800 with extra additions to the PR to assure a good landing of nested modules.

We might want to observe the issues listed at #858 and add extra commits

cc @gibson042

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -41,7 +41,7 @@ extend( QUnit, {
 		if ( executeNow instanceof Function ) {
 			config.moduleStack.push( module );
 			setCurrentModule( module );
-			executeNow( moduleFns );
+			executeNow.call( module.testEnvironment, moduleFns );
 			config.moduleStack.pop();
 			module = module.parentModule || currentModule;
 		}
@@ -56,9 +56,18 @@ extend( QUnit, {
 			var module = {
 				name: moduleName,
 				parentModule: parentModule,
-				testEnvironment: testEnvironment,
 				tests: []
 			};
+
+			var env = {};
+			if ( parentModule ) {
+				extend( env, parentModule.testEnvironment );
+				delete env.beforeEach;
+				delete env.afterEach;
+			}
+			extend( env, testEnvironment );
+			module.testEnvironment = env;
+
 			config.modules.push( module );
 			return module;
 		}
--- a/src/core.js
+++ b/src/core.js
@@ -10,8 +10,9 @@ extend( QUnit, {
 
 	// call on start of module test to prepend name to all tests
 	module: function( name, testEnvironment, executeNow ) {
+		var module, moduleFns;
 		var currentModule = config.currentModule;
-		var module;
+
 		if ( arguments.length === 2 ) {
 			if ( testEnvironment instanceof Function ) {
 				executeNow = testEnvironment;
@@ -32,10 +33,15 @@ extend( QUnit, {
 
 		module = createModule();
 
+		moduleFns = {
+			beforeEach: setHook( module, "beforeEach" ),
+			afterEach: setHook( module, "afterEach" )
+		};
+
 		if ( executeNow instanceof Function ) {
 			config.moduleStack.push( module );
 			setCurrentModule( module );
-			executeNow();
+			executeNow( moduleFns );
 			config.moduleStack.pop();
 			module = module.parentModule || currentModule;
 		}
@@ -293,4 +299,14 @@ function done() {
 		total: config.stats.all,
 		runtime: runtime
 	});
+}
+
+function setHook( module, hookName ) {
+	if ( module.testEnvironment === undefined ) {
+		module.testEnvironment = {};
+	}
+
+	return function( callback ) {
+		module.testEnvironment[ hookName ] = callback;
+	};
 }
--- a/src/core.js
+++ b/src/core.js
@@ -9,12 +9,15 @@ QUnit.version = "@VERSION";
 extend( QUnit, {
 
 	// call on start of module test to prepend name to all tests
-	module: function( name, testEnvironment ) {
-		var currentModule = {
-			name: name,
-			testEnvironment: testEnvironment,
-			tests: []
-		};
+	module: function( name, testEnvironment, executeNow ) {
+		var currentModule = config.currentModule;
+		var module;
+		if ( arguments.length === 2 ) {
+			if ( testEnvironment instanceof Function ) {
+				executeNow = testEnvironment;
+				testEnvironment = undefined;
+			}
+		}
 
 		// DEPRECATED: handles setup/teardown functions,
 		// beforeEach and afterEach should be used instead
@@ -27,8 +30,37 @@ extend( QUnit, {
 			delete testEnvironment.teardown;
 		}
 
-		config.modules.push( currentModule );
-		config.currentModule = currentModule;
+		module = createModule();
+
+		if ( executeNow instanceof Function ) {
+			config.moduleStack.push( module );
+			setCurrentModule( module );
+			executeNow();
+			config.moduleStack.pop();
+			module = module.parentModule || currentModule;
+		}
+
+		setCurrentModule( module );
+
+		function createModule() {
+			var parentModule = config.moduleStack.length ?
+				config.moduleStack.slice( -1 )[ 0 ] : null;
+			var moduleName = parentModule !== null ?
+				[ parentModule.name, name ].join( " > " ) : name;
+			var module = {
+				name: moduleName,
+				parentModule: parentModule,
+				testEnvironment: testEnvironment,
+				tests: []
+			};
+			config.modules.push( module );
+			return module;
+		}
+
+		function setCurrentModule( module ) {
+			config.currentModule = module;
+		}
+
 	},
 
 	// DEPRECATED: QUnit.asyncTest() will be removed in QUnit 2.0.
--- a/test/main/modules.js
+++ b/test/main/modules.js
@@ -242,3 +242,55 @@ QUnit.module( "contained suite arguments", function( hooks ) {
 		} );
 	} );
 } );
+
+QUnit.module( "contained suite `this`", function( hooks ) {
+	this.outer = 1;
+
+	hooks.beforeEach( function() {
+		this.outer++;
+	} );
+
+	hooks.afterEach( function( assert ) {
+		assert.equal(
+			this.outer, 42,
+			"in-test environment modifications are visible by afterEach callbacks"
+		);
+	} );
+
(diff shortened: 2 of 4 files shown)


test/main/modules.js
```diff
@@ -147,3 +147,57 @@ QUnit.module( "Deprecated setup/teardown", {
...
+	QUnit.test( "test after all nesting modules processed and new module declared",
+		function( assert ) {
+			var module = assert.test.module;
+			assert.equal( module.name, "nested modules > second outer" );
+		});
+});
+
+QUnit.test( "modules with nested functions does not spread beyond", function( assert ) {
+	assert.equal( assert.test.module.name, "pre-nested modules" );
+});
```
leobalter: @jzaefferer @gibson042 I need opinions on this test.

gibson042: Seems ok to me. The interaction itself sucks, but the test captures it well.

leobalter:
@gibson042 I made two commits to include arguments on the contained modules callback. I appreciate if you review and criticize.

cc @jzaefferer

test/main/modules.js
```diff
@@ -147,3 +147,102 @@ QUnit.module( "Deprecated setup/teardown", {
...
+		assert.strictEqual( typeof hooks.afterEach, "function" );
+	} );
+
+	QUnit.module( "contained suite argument hooks", function( test, hooks ) {
+		var beforeEach = hooks.beforeEach;
+		var afterEach = hooks.afterEach;
+
+		beforeEach( function( assert ) {
+			assert.ok( true, "beforeEach called" );
+		} );
```
leobalter: This is even better on ES6:

```js
QUnit.module( "foo", ( test, { beforeEach, afterEach } ) => {
  beforeEach: () => {
    ...
  };

  ...
});
```

Also, a funny fact:

```js
const describe = QUnit.module;

describe( "wat", it => {
  it( "works", ( { ok } ) => {
    ok( true, "omg" );
  });
});
```

src/core.js_
```diff
@@ -0,0 +1,312 @@
+QUnit.urlParams = urlParams;
```
gibson042: I'm assuming these `.js_` files are accidental additions.

leobalter: caught in a git add .

leobalter: 57527a3

gibson042:
> I made two commits to include arguments on the contained modules callback. I appreciate if you review and criticize.

I'm not sure about these particular parameters. As I've noted [elsewhere](https://github.com/jquery/qunit/pull/670#discussion_r26321413), I'd like `module` and `test` callback signatures to look as similar as possible. But further than that, I'd like `module` callbacks to be straightforward. It seems to me that data passed in to them—including as `this`—should be entirely sufficient to define submodules/tests/etc., but hopefully not exceed that.

In my opinion, it's better to pass nothing at all now than it is to guess wrong about the signature. Since this iteration still requires access to the global `QUnit` for some operations (e.g., submodules), I'd actually prefer an argument-free `executeNow()` for this PR, saving context/arguments for later [as discussed](https://github.com/jquery/qunit/pull/800#issuecomment-139336531). For example, a possible future that would be precluded by passing `test` now:
```js
QUnit.module("outer", function( outerModule ) {
	// environment (prototype for test/submodule context)
	this.outer = true;
	this.outerBefore = 0;

	// test before/after
	outerModule.test.beforeEach(function() {
		this.outerBefore++;
	});

	// tests
	outerModule.test("foo", function( assert ) {
		assert.equal(this.outerBefore, 1, "beforeEach updated environment");
	});
	outerModule.test("bar", function( assert ) {
		assert.equal(this.outerBefore, 1, "sibling tests don't share environments");
	});

	// submodules
	outerModule.module("inner", function( innerModule ) {
		// environment (inherits from parent)
		this.inner = true;

		function assertEnvironment( assert, before ) {
			assert.ok(this.outer, "outer module initialized environment");
			assert.ok("outerBefore" in this, "outer module beforeEach updated environment");
			assert.equal(this.outerBefore, 1, "submodules don't share environments");
			assert.ok(this.inner, "inner module initialized environment");
			if ( before ) {
				return;
			}
			assert.ok(this.innerBefore, "inner module beforeEach updated environment");
			assert.ok(this.tested, "test updated environment");
		}

		// test before/after
		innerModule.test.beforeEach(function( assert ) {
			this.innerBefore = true;
			assertEnvironment( assert, true );
		});
		innerModule.test.afterEach(function( assert ) {
			assertEnvironment( assert );
		});

		// tests
		innerModule.test("bar", function( assert ) {
			this.tested = true;
			assertEnvironment( assert );
		});
	});
});
```

leobalter:
```js
QUnit.module("outer", function( outerModule ) {
    // environment (prototype for test/submodule context)
    this.outer = true;
    this.outerBefore = 0;

    // test before/after
    outerModule.test.beforeEach(function() {
        this.outerBefore++;
    });

    // tests
    outerModule.test("foo", function( assert ) {
        assert.equal(this.outerBefore, 1, "beforeEach updated environment");
    });
    outerModule.test("bar", function( assert ) {
        assert.equal(this.outerBefore, 1, "sibling tests don't share environments");
    });

    ...
```

These `this.outerBefore` tests seems very confusing to me. I can understand the point here but I'm uncomfortable with it so far.

---

> I'm not sure about these particular parameters.

And reviewing my code I'll indeed remove the first `test` argument.

And following your own example, I would stick with the first argument being an object with the hooks methods and it can later being extended. This way:

```js
QUnit.module("outer", function( outerModule ) {
    ...
    outerModule.beforeEach(function() {
        ...
    });

    outerModule.afterEach(function() {
        ...
    });
```

but also, later in further additions:

```js
QUnit.module("outer", function( t ) {
    t.beforeEach(function() {
        ...
    });
    t.test( "the test method", function( assert ) {
        ...
    });
});
```

and using ES6:

```js
QUnit.module("outer", ( { beforeEach } ) => {
    beforeEach(function() {
        ...
    });
    QUnit.test( "the test method", function( assert ) {
        ...
    });
});
```

The reason to start with `beforeEach` and `afterEach` methods is to avoid calling modules with three arguments, where the hooks object is already big enough:

```js
QUnit.module( "outer", {
        beforeEach: function( assert ) {
            ...
        },
        afterEach: function( assert ) {
            ...
        }
    },
    function( t ) {
        QUnit.test( "the test method", function( assert ) {
            ...
        });
    }
);
```

leobalter:
@gibson042: both 29f5b84 and 53778e3 has the shared `this` based on your example.

The `this` object is only propagated from the outer to the inner scopes without duplicating the test hooks.

I also removed the test argument as you can see on the fixup commits.

We may then extend the module callback argument to include `test` later. I want to at least bring the shared `this` and the hooks when landing the nested modules.

src/core.js
```diff
@@ -27,8 +31,51 @@ extend( QUnit, {
...
+				[ parentModule.name, name ].join( " > " ) : name;
+			var module = {
+				name: moduleName,
+				parentModule: parentModule,
+				tests: []
+			};
+
+			var env = {};
+			if ( parentModule ) {
+				extend( env, parentModule.testEnvironment );
```
gibson042: I think `Object.create`-style inheritance would be better, but am willing to shelve that for now.

test/main/modules.js
```diff
@@ -147,3 +147,141 @@ QUnit.module( "Deprecated setup/teardown", {
...
+
+QUnit.module( "contained suite `this`", function( hooks ) {
+	this.outer = 1;
+
+	hooks.beforeEach( function() {
+		this.outer++;
+	} );
+
+	hooks.afterEach( function( assert ) {
+		assert.equal( this.outer, 2 );
```
gibson042: The increment-based tests in https://github.com/leobalter/qunit/commit/29f5b842e61e2b735b0cc9046034c26409375271#diff-3ff526461245427b5e9714efeb6ca4ed failed, right? It is unfortunate, but IIRC tests within a module _do_ currently share environment.

gibson042:
:+1: The single callback argument is a decent (if limited) module stand-in and has good room for growth, and you are absolutely right about this:
> The reason to start with `beforeEach` and `afterEach` methods is to avoid calling modules with three arguments, where the hooks object is already big enough

Let's hope that `module` and `test` follow suit, so that we can eventually get to a point where `module` callbacks need nothing other than what's passed directly to them, just like current `test` callbacks.

src/core.js
leobalter: I would love to do this. We still support ES3 until QUnit 2.0.0

test/main/modules.js
leobalter: this commit will be merged with 53778e3.

It is passing here. I just wanted to set the steps on different commits during the review stage.

> It is unfortunate, but IIRC tests within a module do currently share environment.

I've done it this way getting from this assertion below from your example.

```js
assert.ok(this.outer, "outer module initialized environment");
```

Should it be different?

gibson042: Turns out the environments _are_ already distinct (:raised_hands:), so there's just some missing code here to show that in-test environment modifications are visible by `afterEach` callbacks but not by subsequent tests (as demonstrated in https://github.com/leobalter/qunit/commit/29f5b842e61e2b735b0cc9046034c26409375271#diff-3ff526461245427b5e9714efeb6ca4ed ).

src/core.js
gibson042: Arbitrary inheritance is possible in ES3, too:
```js
var createObject = (function( Base ) {
	return function( prototype ) {
		Base.prototype = prototype;
		return new Base();
	};
})( Function() );
…
env = createObject( parentModule.testEnvironment );
```

The above is similar to a stripped-down version of lodash's [`baseCreate`](https://github.com/lodash/lodash/blob/d407d8aa08a6f739441a82e3b7c9d8a5cf66ba16/lodash.js#L1987-L2005), but you could easily go all the way to [`_.create`](https://lodash.com/docs#create) if you want this part to look like `module.testEnvironment = createObject( parentModule.testEnvironment, testEnvironment )` (removing the need for a separate `extend`).

test/main/modules.js
leobalter: Good catch. I got it on ce70ce9

src/core.js
leobalter: This is cool, but I would leave this improvement for later as this requires a whole new review. Does it works for you?

gibson042: Yep.

gibson042:
I think this is good to go; [module environment inheritance](https://github.com/jquery/qunit/pull/859#discussion_r40965697) can be addressed with a subsequent issue & PR.

leobalter:
Opened #869 to follow up with this.

PR #861, opened 2015-09-17
browserstack improvements

### LINKED ISSUE #860: More browserstack-runner improvements
As commented here, the current build script still seems to have issues: https://github.com/jquery/qunit/pull/856#issuecomment-141020718
Failures should propagate to having the build fail.

That was iOS timing out, which we should try to avoid by sticking with simulators, like this: 
https://github.com/shirish87/sizzle/commit/577b55b98f17c1bc58915548a9d6b6a2bda1b0af

/cc @shirish87 are there other improvements we could implement? See 39d882110364d516f15bfa0e59cc141bea9549bd for the recent changes @leobalter landed.

leobalter:
Closes #860 

I'm not sure it 1535b96 really works. The screenshots already show a simulator.

### DIFF
--- a/build/browserstack-current-1.json
+++ b/build/browserstack-current-1.json
@@ -15,7 +15,8 @@
 		"edge_latest",
 		{
 			"os": "iOS",
-			"os_version": "8.3"
+			"os_version": "8.3",
+			"real_mobile": false
 		}
 	]
 }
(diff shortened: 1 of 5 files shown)


jzaefferer:
Looks good to me. Even if we're currently getting simulators, for speed purposes it makes sense to stick with them.

Issue #843, opened 2015-08-05
Extend assert.async to accept a count of asynchronous operations


gibson042:
Currently, `assert.async` returns an invoke-once callback, which can lead to overly-cumbersome patterns (e.g., [manual counter tracking](https://github.com/jquery/jquery/blob/acda1d633fd6c8a0a7dc8b8236780daa3fe44cc1/test/unit/basic.js#L9-L22) or [collect and pop](https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/deferred.js#L411-L421)). It would be more convenient to specify the expected count of invocations explicitly by argument:
```js
var done = assert.async( 4 );

asyncOperation1( function( result ) {
	assert.equal(…);
	done();
} );
asyncOperation2( function( result ) {
	assert.equal(…);
	done();
} );
…
```

mzgol:
This would be great and seems a common enough pattern.

mixed:
+1 for this feature.
@jzaefferer Could I work on this issue?

PR #863, opened 2015-09-21
Assert: Add accept a call count parameter of callback of assert.async.

mixed:
Ref #843

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -32,18 +32,21 @@ QUnit.assert = Assert.prototype = {
 
 		return function done() {
 			acceptCallCount -= 1;
-			if ( acceptCallCount <= 0 ) {
-				if ( !popped ) {
-					test.semaphore -= 1;
-					popped = true;
-					resumeProcessing();
-				} else {
-					test.pushFailure( "Too many called the callback returned from `assert.async`",
-						sourceFromStacktrace( 2 ) );
-				}
-			} else {
+
+			if ( acceptCallCount > 0 ) {
 				pauseProcessing();
+				return;
 			}
+
+			if ( popped ) {
+				test.pushFailure( "Too many calls to the `assert.async` callback",
+					sourceFromStacktrace( 2 ) );
+				return;
+			}
+
+			test.semaphore -= 1;
+			popped = true;
+			resumeProcessing();
 		};
 	},
 
--- a/src/assert.js
+++ b/src/assert.js
@@ -31,16 +31,14 @@ QUnit.assert = Assert.prototype = {
 		pauseProcessing();
 
 		return function done() {
-			acceptCallCount -= 1;
-
-			if ( acceptCallCount > 0 ) {
-				pauseProcessing();
-				return;
-			}
 
 			if ( popped ) {
 				test.pushFailure( "Too many calls to the `assert.async` callback",
 					sourceFromStacktrace( 2 ) );
+				return;
+			}
+			acceptCallCount -= 1;
+			if ( acceptCallCount > 0 ) {
 				return;
 			}
 
--- a/src/assert.js
+++ b/src/assert.js
@@ -15,24 +15,34 @@ QUnit.assert = Assert.prototype = {
 		}
 	},
 
-	// Increment this Test's semaphore counter, then return a single-use function that
+	// Increment this Test's semaphore counter, then return a function that
 	// decrements that counter a maximum of once.
-	async: function() {
+	async: function( count ) {
 		var test = this.test,
-			popped = false;
+			popped = false,
+			acceptCallCount = count;
+
+		if ( typeof acceptCallCount === "undefined" ) {
+			acceptCallCount = 1;
+		}
 
 		test.semaphore += 1;
 		test.usedAsync = true;
 		pauseProcessing();
 
 		return function done() {
-			if ( !popped ) {
-				test.semaphore -= 1;
-				popped = true;
-				resumeProcessing();
+			acceptCallCount -= 1;
+			if ( acceptCallCount <= 0 ) {
+				if ( !popped ) {
+					test.semaphore -= 1;
+					popped = true;
+					resumeProcessing();
+				} else {
+					test.pushFailure( "Too many called the callback returned from `assert.async`",
+						sourceFromStacktrace( 2 ) );
+				}
 			} else {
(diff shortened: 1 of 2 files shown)


src/assert.js
```diff
@@ -15,24 +15,34 @@ QUnit.assert = Assert.prototype = {
...
-				popped = true;
-				resumeProcessing();
+			acceptCallCount -= 1;
+			if ( acceptCallCount <= 0 ) {
+				if ( !popped ) {
+					test.semaphore -= 1;
+					popped = true;
+					resumeProcessing();
+				} else {
+					test.pushFailure( "Too many called the callback returned from `assert.async`",
```
jzaefferer: How about
```js
"Too many calls to the `assert.async` callback"
```

src/assert.js
```diff
@@ -15,24 +15,34 @@ QUnit.assert = Assert.prototype = {
...
 		test.usedAsync = true;
 		pauseProcessing();
 
 		return function done() {
-			if ( !popped ) {
-				test.semaphore -= 1;
-				popped = true;
-				resumeProcessing();
+			acceptCallCount -= 1;
+			if ( acceptCallCount <= 0 ) {
```
jzaefferer: Let's refactor the `else` conditions into  guard clauses:
```js
if ( acceptCallCount > 0 ) {
  pauseProcessing();
  return;
}
if ( popped ) {
  test.pushFailure(...);
  return;
}
[...]

**Comment by jzaefferer** [comment]
Thanks for the contribution! Overall this is looking good, just a few details to address.

File: src/assert.js (hunk 2 above)
Comment (mixed) [inline src/assert.js]: @jzaefferer 
OK. Thank you.
I will update your review.

File: src/assert.js (hunk 1 above)
Comment (mixed) [inline src/assert.js]: @jzaefferer 
Thank you.

File: src/assert.js [hunk 3]
```diff
@@ -15,25 +15,38 @@ QUnit.assert = Assert.prototype = {
...
-			if ( !popped ) {
-				test.semaphore -= 1;
-				popped = true;
-				resumeProcessing();
-			} else {
-				test.pushFailure( "Called the callback returned from `assert.async` more than once",
+			acceptCallCount -= 1;
+
+			if ( acceptCallCount > 0 ) {
+				pauseProcessing();
```
jzaefferer: Looking at this again, this shouldn't be needed. As long as `resumeProcessing()` doesn't run, there's no point in calling `pauseProcessing()` again, right? Should be able to drop that and only keep the return.

jzaefferer:
Apart from that one extra call this looks good to me. @leobalter what do you think?

src/assert.js
gibson042: Assuming that's the case (which it seems to be), the function might read better with a little more rearranging:
```js
if ( popped ) {
	// push failure and return
}

acceptCallCount -= 1;
if ( acceptCallCount === 0 ) {
	// set popped and resume processing
}
```

mixed: @jzaefferer 
If `config.testTimeout` set 1000. If `pauseProcessing()` do not called that delay aggregate time is `testTimeout` time when all `done()` finished. The `testTimeout` is too short for all `done` called. So When `done()` called, reset `setTimeout`.

What do you think it?

mixed: @gibson042 
IMO, Looks Good.

gibson042: > If `pauseProcessing()` do not called that delay aggregate time is `testTimeout` time when all `done()` finished. The `testTimeout` is too short for all `done` called. So When `done()` called, reset `setTimeout`.

`config.testTimeout` always counts from the start of each test; see http://api.qunitjs.com/QUnit.config/ ("a global timeout in milliseconds after which all tests will fail… to prevent the testrunner getting stuck").

mixed: @gibson042 
OK. Thank you.
I got it.^^

leobalter: @mixed, isn't it already paused before the done function?

https://github.com/jquery/qunit/pull/863/files#diff-f8f3ba9b60ca5c3a0a1dd233f0d34122L26

If `config.testTimeout` is not enough, it should be modified by the user.

---

@mixed I agree that @gibson042's suggestion helps in the function readability. It's like saying to check for fails before running the rest of the code.

In another PoV, the way it is, give a very slight improvement to the performance as popped is checked if the function has not returned with the `pauseProcessing();`

---

If anything changes in the code and for future improvements with the tail call optimization, you can blend the return statements with the last call:

```js
...
return pauseProcessing();

...

return test.pushFailure( ... );

...

return resumeProcessing();
```

test/main/async.js
```diff
@@ -164,9 +187,34 @@ QUnit.test( "fails if callback is called more than once in test", function( asse
...
+		throw new Error( msg );
+	};
+
+	assert.throws(function() {
+		var overDone = assert.async( 3 );
+		overDone();
+		overDone();
+		overDone();
+		overDone();
+	}, new RegExp( "Too many calls to the `assert.async` callback" ) );
```
leobalter: only the last overDone should be inside this function.

```js
	var overDone = assert.async( 3 );
	overDone();
	overDone();
	overDone();
	assert.throws(function() {		
		overDone();
	}, new RegExp( "Too many calls to the `assert.async` callback" ) );
```

This reveals a similar issue on a previous existing assertion: https://github.com/jquery/qunit/pull/863/files#diff-fbe33c9a2379400e62e9f4120ec9c024R190

src/assert.js
leobalter: I would have used "Exceeded calls to the `assert.async` callback" instead. But that works too.

leobalter:
Just one last suggestion on the `assert.throws` commands and this will be good to land.

We should also update the docs at https://github.com/jquery/api.qunitjs.com

src/assert.js
mixed: @gibson042 @leobalter 
Thank you for explaining in detail.

mixed:
@leobalter 
OK. I updated testcase and I am going to update document today.

leobalter:
Thank you, @mixed!

mixed:
@leobalter 
I updated docs.
jquery/api.qunitjs.com#110

leobalter:
Thanks, @mixed!

I rebased and squashed the commits on the merged 41d5ee1.

The docs will be updated within the next version release.

PR #866, opened 2015-09-29
Test: Allow async test defined from test callback parameters

leobalter:
As requested on twitter by @fnando, this proposal brings a shortcut to define async tests through the test callback parameters.

This is based on the test callback function length and it is worth to investigate if it doesn't conflict anything on extensions like ember-qunit.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -89,7 +89,7 @@ Test.prototype = {
 	},
 
 	run: function() {
-		var promise;
+		var promise, done;
 
 		config.current = this;
 
@@ -97,16 +97,20 @@ Test.prototype = {
 			QUnit.stop();
 		}
 
+		if ( this.callback.length > 1 ) {
+			done = this.assert.async.call( this.assert );
+		}
+
 		this.callbackStarted = now();
 
 		if ( config.notrycatch ) {
-			promise = this.callback.call( this.testEnvironment, this.assert );
+			promise = this.callback.call( this.testEnvironment, this.assert, done );
 			this.resolvePromise( promise );
 			return;
 		}
 
 		try {
-			promise = this.callback.call( this.testEnvironment, this.assert );
+			promise = this.callback.call( this.testEnvironment, this.assert, done );
 			this.resolvePromise( promise );
 		} catch ( e ) {
 			this.pushFailure( "Died on test #" + ( this.assertions.length + 1 ) + " " +
--- a/test/main/async.js
+++ b/test/main/async.js
@@ -169,6 +169,15 @@ QUnit.test( "fails if callback is called more than once in test", function( asse
 	done();
 });
 
+QUnit.module( "async test defined by test callback" );
+
+QUnit.test( "defined done", function( assert, done ) {
+	setTimeout(function() {
+		assert.ok( true );
+		done();
+	});
+});
+
 QUnit.module( "assert.async fails if callback is called more than once in", {
 	beforeEach: function( assert ) {
 


fnando:
1MM :heart: :joy:

leobalter:
cc @rwjblue and @stefanpenner

rwjblue:
Shouldn't cause issues with ember-qunit.

To allow folks to use this in ember-qunit I'll have to double check to confirm we pass the additional arguments through (and that the arity of the callback function is correct), but that shouldn't be considered a blocker for this...

src/test.js
```diff
@@ -97,16 +97,20 @@ Test.prototype = {
 			QUnit.stop();
 		}
 
+		if ( this.callback.length > 1 ) {
+			done = this.assert.async.call( this.assert );
```
mzgol: Why not `done = this.assert.async();`?

leobalter: sure! I was testing when I wrote this and I forgot to clean it up.

jzaefferer:
It looks like a nice addition to make async testing quicker to set up. The implicit nature might also become a problem. What happens when some (private/custom) extension introduces that 2nd parameter already? Will the test time out eventually? Do we even have a default timeout?

Can we make the detection more precise? For example, make sure that the 2nd argument is actually called "done"?

fnando:
> Can we make the detection more precise? For example, make sure that the 2nd argument is actually called "done"?

@jzaefferer I've been in projects that used `done` and `next` for this. Not sure if relying on the argument name is a good idea. But I think it is possible with `func.toString()` (works in node, not sure about browsers).

rwjblue:
> make sure that the 2nd argument is actually called "done"

Relying on the argument name will roughly mean that testing minified (I often like to test minified production ready code to confirm just before a deployment) will fail...

leobalter:
> What happens when some (private/custom) extension introduces that 2nd parameter already?

That was my first concern and why I confirmed it is an ok for ember-qunit. I don't know about other QUnit usages that could brake with this option.

gibson042:
To be honest, I'm uncomfortable with automatically opening an asynchronous context by function arity. It makes future signature changes (including those that might further [`module`/`test` convergence](https://github.com/jquery/qunit/pull/670#discussion_r26321413)) awkward.

If we're comfortable with magic, though, maybe it could be slipped into `asyncTest`?

Krinkle:
The existing API is fine. I don't see why we need a way to do it in one line. It adds additional complexity, things that can go wrong, cause bugs etc., for a marginal benefit.

This looks like a case of brain/gut conflict where this will give the illusion of saving time because you may have to write less code. But that's a misconception. Developers rarely (if ever) spend the majority of their time writing the actual code.

I'm not principally opposed to having async-test-in-one-line, but I would recommend against this particular implementation. Per @gibson042, if we really want this, let's bring back `asyncTest`.

https://api.qunitjs.com/QUnit.asyncTest/
> This method is deprecated and will be removed in QUnit 2.0. Use QUnit.test() in conjunction with assert.async() instead.

leobalter:
this example also solves the same issue, and it can be an external extension:

```js
asyncTest = function( name, callback ) {
  return QUnit.test( name, function( assert ) {
    var done = assert.async();
    return callback.call( this, assert, done );
  });
};

the returns are there to keep the tests promiseable, and maybe the second one is not necessary.

**Comment by jzaefferer** [comment]
Old:
```js
asyncTest( name, function() {
  setTimeout( function() {
    start();
  } );
} );
```
New:
```js
asyncTest( name, function( assert, start ) {
  setTimeout( function() {
    done();
  } );
} );
```
Certainly makes migration a lot easier, just add one argument (or two, if you're not using the `assert`).

If that's what we go for, we should mention that on https://github.com/apsdehal/qunit-migrate/issues/4

Could also use the opportunity to rename the method to `testAsync`, makes it easier to spot invalid usage by throwing an error from `asyncTest`.

leobalter:
My example was referring to a custom asyncTest function, made externally. I prefer to leave asyncTest deprecated as it is and not keep a test method other than `QUnit.test`.

> It makes future signature changes (including those that might further module/test convergence) awkward.

This is a good argument.

---

@fnando, as you suggested this, would bring some ideas and/or suggestions?

fnando:
@leobalter To be fair, using the arity as a way of determining if the test is async is not that magical and is a known technic from other libs (e.g. mocha.js).

I never really liked having two different functions like `test` and `asyncTest`, so I'd run away from that. If using arity is a problem, I would stick to `assert.async()`, even though I find it a boring task (small task, but a boring one).

tl;dr: I like this convenience, but I won't die if I can't have it. :)

leobalter:
@JamesMGreene, any addition?

leobalter:
I'm closing this. So far it seems better to wait for more discussion and arguments if we want to move this forward.

PR #867, opened 2015-10-01
HTML Reporter: Collapse multiple failing tests

### LINKED ISSUE #664: Collapse failed tests in HTML reporter

leobalter:
Replaces #769 
Fixes #664

@jzaefferer this is rebased and I squashed @gauravmittal1995 commits to clarify the changes I applied after.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -63,6 +63,7 @@ QUnit.init = function() {
 };
 
 var config = QUnit.config,
+	collapseNext = false,
 	hasOwn = Object.prototype.hasOwnProperty,
 	defined = {
 		document: window.document !== undefined,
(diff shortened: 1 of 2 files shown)


jzaefferer:
Mixing this change with onevar cleanup makes it harder to review than necessary. Despite that, looks good. Much simpler logic is certainly preferable.

leobalter:
If you prefer I can keep this onevar cleanup separated.

jzaefferer:
Yeah, let's keep that separate. If this change ends up causing a regression, it'll be easier to track down.

PR #873, opened 2015-10-08
Test: Fix regression when a failing test canceled the module hooks

### LINKED ISSUE #870: beforeEach, afterEach hooks not called if run second time after previous at least one test failed
I have next suite:
```javascript
(function(module, test)
{
    var beforeEachHook = function(assert)
    {
        console.log("beforeEach: " + assert.test.testName);
    };

    var afterEachHook = function(assert)
    {
        console.log("afterEach: " + assert.test.testName);
    };

    module("Suite with one failed test",
        {
            beforeEach: beforeEachHook,
            afterEach: afterEachHook
        }
    );

    test("First unit test", function(assert)
    {
        assert.ok(true);
    });

    test("Second unit test", function(assert)
    {
        assert.ok(false);
    });

    test("Third unit test", function(assert)
    {
        assert.ok(true);
    });

}(QUnit.module, QUnit.test));
```

When I run it first time I received in console next messages:
```
beforeEach: First unit test
afterEach: First unit test
beforeEach: Second unit test
afterEach: Second unit test
beforeEach: Third unit test
afterEach: Third unit test
```
Result of execution:
```
First unit test - okay
Second unit test - failed
Third unit test - okay
```

When I run it second time I received only next messages:
```
beforeEach: Second unit test
afterEach: Second unit test
beforeEach: First unit test
afterEach: First unit test
```
Result of execution:
```
First unit test - okay
Second unit test - failed
Third unit test - okay
```

As you can see in second run `before each` and `after each` of `Third unit test` is not called.

```javascript
QUnit.version = "1.19.0";
```

As I noticed `Test#hooks` not add `beforeEach`, `afterEach` hooks as it not pass next condition:
```javascript
		if ( this.module.testEnvironment &&
				QUnit.objectType( this.module.testEnvironment[ handler ] ) === "function" ) {
			hooks.push( this.queueHook( this.module.testEnvironment[ handler ], handler ) );
		}
```
```javascript
QUnit.objectType( this.module.testEnvironment[ handler ] )
```
equal to `"undefined"`

As it possibly cleared in ran `Test#before` (I notice that "execution" of test starts before all `Test#hook` executed) in next lines:
```javascript
		if ( this.module.testEnvironment ) {
			delete this.module.testEnvironment.beforeEach;
			delete this.module.testEnvironment.afterEach;
		}
```

leobalter:
Closes #870

This fixes a regression bug that was preventing the module hooks from
being assigned on the following tests of the same module.

To make it work without reverting the changes on fdf00a4, it improves
the reordered queue for tests that previously failed in the same
browser session.

---

This little bug took me 3 hours to fix, at least the queue is a little bit faster now.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -230,7 +230,7 @@ Test.prototype = {
 	},
 
 	queue: function() {
-		var bad,
+		var priority,
 			test = this;
 
 		if ( !this.valid() ) {
@@ -262,16 +262,11 @@ Test.prototype = {
 			]);
 		}
 
-		// `bad` initialized at top of scope
-		// defer when previous test run passed, if storage is available
-		bad = QUnit.config.reorder && defined.sessionStorage &&
+		// Prioritize previously failed tests, detected from sessionStorage
+		priority = QUnit.config.reorder && defined.sessionStorage &&
 				+sessionStorage.getItem( "qunit-test-" + this.module.name + "-" + this.testName );
 
-		if ( bad ) {
-			run();
-		} else {
-			synchronize( run, true );
-		}
+		return synchronize( run, priority );
 	},
 
 	push: function( result, actual, expected, message, negative ) {
@@ -466,19 +461,42 @@ function generateHash( module, testName ) {
 	return hex.slice( -8 );
 }
 
-function synchronize( callback, last ) {
+function synchronize( callback, priority ) {
+	var last = !priority;
+
 	if ( QUnit.objectType( callback ) === "array" ) {
 		while ( callback.length ) {
 			synchronize( callback.shift() );
 		}
 		return;
 	}
-	config.queue.push( callback );
+
+	if ( priority ) {
+		priorityFill( callback );
+	} else {
+		config.queue.push( callback );
+	}
 
 	if ( config.autorun && !config.blocking ) {
 		process( last );
 	}
 }
+
+// Place previously failed tests on a queue priority line, respecting the order they get assigned.
+function priorityFill( callback ) {
+	var queue, prioritizedQueue;
+
+	queue = config.queue.slice( priorityFill.pos );
+	prioritizedQueue = config.queue.slice( 0, -config.queue.length + priorityFill.pos );
+
+	queue.unshift( callback );
+	queue.unshift.apply( queue, prioritizedQueue );
+
+	config.queue = queue;
+
+	priorityFill.pos += 1;
+}
+priorityFill.pos = 0;
 
 function saveGlobal() {
 	config.pollution = [];


leobalter:
I'll release 1.20.0 after merging this one.

src/test.js
```diff
@@ -466,20 +461,43 @@ function generateHash( module, testName ) {
...
+		config.queue.push( callback );
+	}
 
 	if ( config.autorun && !config.blocking ) {
 		process( last );
 	}
 }
 
+// Place previously failed tests on a queue priority line, respecting the order they get assigned.
+function priorityFill( callback ) {
```
jzaefferer: I don't understand what this method is doing.

src/test.js
```diff
@@ -466,20 +461,43 @@ function generateHash( module, testName ) {
...
+
+	queue = config.queue.slice( priorityFill.pos );
+	prioritizedQueue = config.queue.slice( 0, -config.queue.length + priorityFill.pos );
+
+	queue.unshift( callback );
+	queue.unshift.apply( queue, prioritizedQueue );
+
+	config.queue = queue;
+
+	return priorityFill.pos++;
```
jzaefferer: Why is this a return, when the return value is ignored?

I also suggest switching to `+= 1` instead of `++`.

src/test.js
leobalter: without this *priority line* for failed tests, I would add successive tests that previously failed on the very beggining of the queue.

Let's say with the following tests:

```js
QUnit.test('test 1', function( assert ) {
  assert.ok( true );
});

QUnit.test('test 2', function( assert ) {
  assert.ok( false );
});

QUnit.test('test 3', function( assert ) {
  assert.ok( false );
});
```

The queue would regularly be on the first run:

- `test 1` (ok)
- `test 2` (failed)
- `test 3` (failed)

On the second run, it should be on the following order:

- `test 2` (failed)
- `test 3` (failed)
- `test 1` (ok)

And without this *priority queue* I would have instead:

- `test 3` (failed)
- `test 2` (failed)
- `test 1` (ok)

src/test.js
leobalter: As you noticed, this return does not improve anything, I'll remove it. Even though, I'm getting paranoid to add `return` on every function because of Tail Call Optimization, spec'ed on ES2015 and which will affect JS engines silently.

The TCO avoids memory leaks and brings a slight performance optimization when we use it.

With that said, it **only** affects function returns of other function calls (the tail call), like `return anything();`. The subroutine function will reuse the current stack frame (the scope stack) without creating a new one.

I know I explained more than the necessary, I just wanted to share this. It is awesome, right? We should make TCO calls a recommendation on our style guide.

leobalter: Fixed at 05bdc35

src/test.js
jzaefferer: Gotcha, thanks for the explanation. Should be good to land then.

PR #874, opened 2015-10-12
Core: Support Symbol types on QUnit.equiv

leobalter:
Also adds an extra code golf (with step-by-step commits) based on coverage.

As mentioned on 5e4ba82, `QUnit.objectType` already covers all JS types except the new comming Symbol on ES2015. For other object types it will return `'object'` which is covered too.

### DIFF
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -99,6 +99,7 @@ function objectType( obj ) {
 		case "Date":
 		case "RegExp":
 		case "Function":
+		case "Symbol":
 			return type.toLowerCase();
 	}
 	if ( typeof obj === "object" ) {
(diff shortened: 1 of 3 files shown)


src/core/utilities.js
```diff
@@ -99,12 +99,12 @@ function objectType( obj ) {
 		case "Date":
 		case "RegExp":
 		case "Function":
+		case "Symbol":
 			return type.toLowerCase();
 	}
 	if ( typeof obj === "object" ) {
 		return "object";
 	}
```
leobalter: I can't figure any other type than "object" here, maybe I should remove this if and return `"object"` at this part as well.

This is remaining as a fail safe only.

jzaefferer: No idea.

jzaefferer:
Looks good to me.

PR #875, opened 2015-10-16
Qunit only

### LINKED ISSUE #496: [QUnit.only] Add support for source-based test filtering (aka exclusive tests)
I think it would be handy to be able to filter tests right in the source code. This feature is essential when [Karma](http://karma-runner.github.io) is used and the test suite is large. 

Mocha and Jasmine already support this feature via '.only' and 'iit' respectively. 

What do you guys think?

ebenoist:
Included is a first pass that does not include any changes to the reporter/html.js. I'd love some feedback on how the reporter is normally tested and what patterns I can follow (if any).

Fixes #496

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ dist
 node_modules
 build/report
 browserstack-run.pid
+temp/
--- a/src/core.js
+++ b/src/core.js
@@ -85,6 +85,8 @@ extend( QUnit, {
 
 	skip: skip,
 
+	only: only,
+
 	// DEPRECATED: The functionality of QUnit.start() will be altered in QUnit 2.0.
 	// In QUnit 2.0, invoking it will ONLY affect the `QUnit.config.autostart` blocking behavior.
 	start: function( count ) {
--- a/src/test.js
+++ b/src/test.js
@@ -1,3 +1,5 @@
+var focused = false;
+
 function Test( settings ) {
 	var i, l;
 
@@ -246,7 +248,6 @@ Test.prototype = {
 				},
 
 				test.hooks( "beforeEach" ),
-
 				function() {
 					test.run();
 				},
@@ -545,6 +546,8 @@ function asyncTest( testName, expected, callback ) {
 
 // Will be exposed as QUnit.test
 function test( testName, expected, callback, async ) {
+	if ( focused )  { return; }
+
 	var newTest;
 
 	if ( arguments.length === 2 ) {
@@ -564,10 +567,36 @@ function test( testName, expected, callback, async ) {
 
 // Will be exposed as QUnit.skip
 function skip( testName ) {
+	if ( focused )  { return; }
+
 	var test = new Test({
 		testName: testName,
 		skip: true
 	});
 
 	test.queue();
+}
+
+// Will be exposed as QUnit.only
+function only( testName, expected, callback, async ) {
+	var newTest;
+
+	if ( focused )  { return; }
+
+	QUnit.config.queue.length = 0;
+	focused = true;
+
+	if ( arguments.length === 2 ) {
+		callback = expected;
+		expected = null;
+	}
+
+	newTest = new Test({
+		testName: testName,
+		expected: expected,
+		async: async,
+		callback: callback
+	});
+
+	newTest.queue();
 }
--- a/test/only.html
+++ b/test/only.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>QUnit Only Test Suite</title>
+		<link rel="stylesheet" href="../dist/qunit.css">
+		<script src="../dist/qunit.js"></script>
+		<script src="only.js"></script>
+	</head>
+	<body>
+		<div id="qunit"></div>
+	</body>
+</html>
--- a/test/only.js
+++ b/test/only.js
@@ -1,0 +1,17 @@
+QUnit.module( "QUnit.only" );
+
+QUnit.test( "implicitly skipped test", function( assert ) {
+	assert.ok( false, "test should be skipped" );
+});
+
+QUnit.only( "only run this test", function( assert ) {
+	assert.ok( true, "only this test should run" );
+});
+
+QUnit.test( "another implicitly skipped test", function( assert ) {
+	assert.ok( false, "test should be skipped" );
+});
+
+QUnit.only( "ignore the subsequent calls to only", function( assert ) {
+	assert.ok( false, "this test should be skipped" );
+});
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -106,7 +106,8 @@ grunt.initConfig({
 			"test/reporter-html/index.html",
 			"test/reporter-html/legacy-markup.html",
 			"test/reporter-html/no-qunit-element.html",
-			"test/reporter-html/single-testid.html"
+			"test/reporter-html/single-testid.html",
+			"test/only.html"
 		]
 	},
 	coveralls: {


.gitignore
```diff
@@ -2,3 +2,4 @@ dist
 node_modules
 build/report
 browserstack-run.pid
+temp/
```
leobalter: @ebenoist, it was discussed before about adding this temp folder on the gitignore. As it is only created during on grunt failures and it is not persistent, it shouldn't be here.

Although, I don't mind with it being listed here. If the team agrees, I would just leave this folder, better than having further PRs with the folder contents.

test/only.html
```diff
@@ -0,0 +1,14 @@
...
+	<head>
+		<meta charset="UTF-8">
+		<title>QUnit Only Test Suite</title>
+		<link rel="stylesheet" href="../dist/qunit.css">
+		<script src="../dist/qunit.js"></script>
+		<script src="only.js"></script>
+	</head>
+	<body>
+		<div id="qunit"></div>
+		<div id="qunit-fixture">test markup</div>
```
leobalter: this line is not necessary. We're not using the fixture.

test/only.js
```diff
@@ -0,0 +1,13 @@
...
+	assert.ok( false, "test should be skipped" );
+});
+
+QUnit.only( "only run this test", function( assert ) {
+	assert.ok( true, "only this test should run" );
+});
+
+QUnit.test( "another implicitly skipped test", function( assert ) {
+	assert.ok( false, "test should be skipped" );
+});
```
leobalter: it would be great to add a second `QUnit.only` with a failing assertion. This way we can be sure it only runs the first assigned.

src/core.js
```diff
@@ -303,6 +305,7 @@ function done() {
 	passed = config.stats.all - config.stats.bad;
 
 	runLoggingCallbacks( "done", {
+		skipped: Test.count - config.stats.all,
```
leobalter: This deserves some tests. And the tests might be the most complex to implement this single line. Maybe we should split this part to address it in another PR.

ebenoist: I couldn't agree more. I noticed that some of these changes were really for the html reporter. I'll pill this and some of the other attributes out of this PR and address them in the next.

test/only.js
ebenoist: The current behavior is the last only wins. I'm happy to add a spec for that behavior. Making the first only win mean setting some state that I could interrogate later when the next only is encountered.

test/only.html
ebenoist: :+1:

test/only.js
leobalter: It seems fine to keep this behavior, last assigned wins. We just need to document this.

On these tests we may now add a failing `.only` before the one that should pass.

ebenoist: Sounds good to me. I'll try and get something up with this feedback tonight or tomorrow. Thanks for the feedback.

leobalter: Thank you! Regarding the changes it looks great! If we get docs for it at https://github.com/jquery/api.qunitjs.com I'll be able to release it soon!

jzaefferer: We should keep it consistent with `.only` in Mocha, where, I think, the first one wins.

ebenoist: I'm game for that change as well. Jasmine takes another approach and makes `only` cumulative.

Either one of the above approaches will incur more complexity than the given solution. Let me know what you think.

leobalter: The consistency is a very reasonable argument. That might be simple if we register the first `only` assignment and check it on every only call. If another `only` is called, it will be registered as a regular test.

ebenoist: :+1: works for me.

jzaefferer: The method could override itself with a reference to QUnit.test when called. Then it won't run again.

ebenoist: That would work for me as well. It feels a little clever though and would limit something like a change to cumulative onlys. I'm good either way. I think I'd personally prefer the config over the ref swap, but I'm game for whichever you two think would be more appropriate.

jzaefferer: > would limit something like a change to cumulative onlys

How so? If you want to change the API, change it, the actual implementation shouldn't block anything.

ebenoist:
Updated with PR comments addressed. Let me know what you folks think.

src/test.js
```diff
@@ -571,3 +571,27 @@ function skip( testName ) {
...
+		async: async,
+		callback: callback
+	});
+
+	QUnit.config.queue.length = 0;
+	QUnit.config.filter = newTest.module.name + ": " + newTest.testName;
+
+	newTest.queue();
+
+	QUnit.only = QUnit.test; // First only wins, ignore subsequent onlys
```
leobalter: [Code style](http://contribute.jquery.org/style-guide/js/#comments) issue:

- Comments are always preceded by a blank line
- Single line comments go over the line they refer to

leobalter:
Note that [it won't prevent](https://github.com/jquery/qunit/pull/875#discussion-diff-42653985R596) a `var only = QUnit.only;` or a ES6's: `import {test, only} from 'qunitjs';`.

Maybe a check on [`QUnit.config.filter`](https://github.com/jquery/qunit/pull/875#discussion-diff-42653985R592) to assign it only if it's not assigned yet. That way an `only` call won't win against an URL parameter, but that does not sound as a problem.

Everything else looks great.

ebenoist:
@leobalter I'd prefer not to do the ref swapping either. What is the buy in on this? I'm happy to add some state to `QUnit.config` that I can interrogate similarly.

ebenoist:
@leobalter I thought about that again, is this what you're thinking?

```Javascript
	// First only wins, ignore subsequent onlys
	if ( !QUnit.config.filter ) {
		QUnit.config.queue.length = 0;
		QUnit.config.filter = newTest.module.name + ": " + newTest.testName;
		newTest.queue();
	}
```

I just updated the PR with that. I like it.

ebenoist:
It has a nice side effect of ensuring that filter is respected first.

leobalter:
It looks great!

ebenoist:
Cool!

src/test.js
```diff
@@ -571,3 +571,27 @@ function skip( testName ) {
...
+		expected: expected,
+		async: async,
+		callback: callback
+	});
+
+	// First only wins, ignore subsequent onlys
+	if ( !QUnit.config.filter ) {
+		QUnit.config.queue.length = 0;
+		QUnit.config.filter = newTest.module.name + ": " + newTest.testName;
+		newTest.queue();
```
leobalter: it sounds weird, but this line should be placed after and outside the if block. 

If the filter is already set to run this `only` test (let's say I click on the rerun button on the reporter), I still need to call `newTest.queue()`.

ebenoist:
Updated!

leobalter:
Great! I'll merge it after the CI checks.

Would you help on the docs @ http://github.com/jquery/api.qunitjs.com ?

ebenoist:
I'd be happy to doc it as well.

ebenoist:
https://github.com/jquery/api.qunitjs.com/pull/114

Does this work?

ebenoist:
Build flake?

jzaefferer:
I guess we currently mix lots of internal state with actual configuration. I'd like to have that improve, instead of adding even more state to that config object.

ebenoist:
@jzaefferer makes sense. The above approach avoids either pitfall. The build still seems to be flaking out though :(

leobalter:
@ebenoist I'm running the CI build again, the error was due to a browserstack timeout. It's passing on my machine, thou.

leobalter:
> I guess we currently mix lots of internal state with actual configuration. I'd like to have that improve, instead of adding even more state to that config object. 

I mistook `config.filter` by `config.testId`. 

Regarding that, an internal state might be better instead of exposing it. A variable would do this work.

@ebenoist, I'm sorry for bringing this extra work, would you mind to replace using QUnit.config.filter to use some local variable?

ebenoist:
By local variable do you mean some state on `QUnit.config` such as `QUnit.config.focused`? I can then interrogate that variable when deciding whether to add additional tests to the queue?

leobalter:
I mean not using `QUnit.config` at all, but instead using a local variable at `src/test.js`

src/test.js
```diff
@@ -571,3 +571,28 @@ function skip( testName ) {
...
+		expected: expected,
+		async: async,
+		callback: callback
+	});
+
+	// First only wins, ignore subsequent onlys
+	if ( !QUnit.config.filter ) {
+		QUnit.config.queue.length = 0;
+		QUnit.config.filter = newTest.module.name + ": " + newTest.testName;
+	}
```
leobalter: Using the variable would be like: 

```js
var hasOnly = false

function only( testName, expected, callback, async ) {

	...

	if ( !hasOnly ) {
		QUnit.config.queue.length = 0;
		hasOnly = true;
	}

	...
```

ebenoist: Understood, but I'd still need to interrogate that state when adding additional tests. I can look up that variable on `QUnit.test` and `QUnit.skip` requests as well. The filter was effectively doing this for me previously.

ebenoist:
```Diff
diff --git a/src/test.js b/src/test.js
index 6078d74..2c619dc 100644
--- a/src/test.js
+++ b/src/test.js
@@ -246,7 +246,6 @@ Test.prototype = {
 				},
 
 				test.hooks( "beforeEach" ),
-
 				function() {
 					test.run();
 				},
@@ -545,6 +544,8 @@ function asyncTest( testName, expected, callback ) {
 
 // Will be exposed as QUnit.test
 function test( testName, expected, callback, async ) {
+	if ( focused )  { return; }
+
 	var newTest;
 
@@ -564,6 +565,8 @@ function test( testName, expected, callback, async ) {
 
 // Will be exposed as QUnit.skip
 function skip( testName ) {
+	if ( focused )  { return; }
+
 	var test = new Test({
 		testName: testName,
 		skip: true
@@ -572,10 +575,17 @@ function skip( testName ) {
 	test.queue();
 }
 
+var focused = false;
+
 // Will be exposed as QUnit.only
 function only( testName, expected, callback, async ) {
 	var newTest;
 
+	if ( focused )  { return; }
+
+	QUnit.config.queue.length = 0;
+	focused = true;
+
 		callback = expected;
 		expected = null;
@@ -588,11 +598,5 @@ function only( testName, expected, callback, async ) {
 		callback: callback
 	});
 
-	// First only wins, ignore subsequent onlys
-	if ( !QUnit.config.filter ) {
-		QUnit.config.queue.length = 0;
-		QUnit.config.filter = newTest.module.name + ": " + newTest.testName;
-	}
-
 	newTest.queue();
 }
```

leobalter: works for me, but the `var focused = false;` should go on the top.

ebenoist:
Updated with the diff from the PR and I hoisted the `focused` var.

leobalter:
merged at efe0303. Thanks, @ebenoist!

ebenoist:
Hooray!

PR #876, opened 2015-10-17
HTML Reporter: Added indicator for filtered test

### LINKED ISSUE #528: There is no visual indication when you're focused on a single test
When you double-click on a test to run only that test, there is no indication that you are only running a single test except for the URL. As you can see from the popularity of [this StackOverflow ticket](http://stackoverflow.com/questions/13752715/qunit-is-only-running-the-first-test/13753426#13753426), this is confusing for more people than just myself. 

I propose that the page offer some indication that you're running a single test and show a link to go back to the full suite.

arghgr:
Closes #528

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -460,6 +460,17 @@ function storeFixture() {
 	}
 }
 
+function appendFilteredTest() {
+	var testId = QUnit.config.testId;
+	if ( !testId || testId.length <= 0 ) {
+		return "";
+	}
+	return "<div id='qunit-filteredTest'>Rerunning selected tests: " + testId.join(", ") +
+		" <a id='qunit-clearFilter' href='" +
+		setUrl({ filter: undefined, module: undefined, testId: undefined }) +
+		"'>" + "Run all tests" + "</a></div>";
+}
+
 function appendUserAgent() {
 	var userAgent = id( "qunit-userAgent" );
 
@@ -467,7 +478,7 @@ function appendUserAgent() {
 		userAgent.innerHTML = "";
 		userAgent.appendChild(
 			document.createTextNode(
-				"QUnit " + QUnit.version  + "; " + navigator.userAgent
+				"QUnit " + QUnit.version + "; " + navigator.userAgent
 			)
 		);
 	}
@@ -531,6 +542,7 @@ QUnit.begin(function( details ) {
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
 			"<h2 id='qunit-banner'></h2>" +
 			"<div id='qunit-testrunner-toolbar'></div>" +
+			appendFilteredTest() +
 			"<h2 id='qunit-userAgent'></h2>" +
 			"<ol id='qunit-tests'></ol>";
 	}
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -11,17 +11,17 @@
 
 /** Font Family and Sizes */
 
-#qunit-tests, #qunit-header, #qunit-banner, #qunit-testrunner-toolbar, #qunit-userAgent, #qunit-testresult {
+#qunit-tests, #qunit-header, #qunit-banner, #qunit-testrunner-toolbar, #qunit-filteredTest, #qunit-userAgent, #qunit-testresult {
 	font-family: "Helvetica Neue Light", "HelveticaNeue-Light", "Helvetica Neue", Calibri, Helvetica, Arial, sans-serif;
 }
 
-#qunit-testrunner-toolbar, #qunit-userAgent, #qunit-testresult, #qunit-tests li { font-size: small; }
+#qunit-testrunner-toolbar, #qunit-filteredTest, #qunit-userAgent, #qunit-testresult, #qunit-tests li { font-size: small; }
 #qunit-tests { font-size: smaller; }
 
 
 /** Resets */
 
-#qunit-tests, #qunit-header, #qunit-banner, #qunit-userAgent, #qunit-testresult, #qunit-modulefilter {
+#qunit-tests, #qunit-header, #qunit-banner, #qunit-filteredTest, #qunit-userAgent, #qunit-testresult, #qunit-modulefilter {
 	margin: 0;
 	padding: 0;
 }
@@ -66,6 +66,12 @@
 	color: #5E740B;
 	background-color: #EEE;
 	overflow: hidden;
+}
+
+#qunit-filteredTest {
+	padding: 0.5em 1em 0.5em 1em;
+	background-color: #F4FF77;
+	color: #366097;
 }
 
 #qunit-userAgent {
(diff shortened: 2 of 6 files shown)


arghgr:
Also added link to clear filter if the tests are filtered.

reporter/html.js
```diff
@@ -467,7 +479,7 @@ function appendUserAgent() {
 		userAgent.innerHTML = "";
 		userAgent.appendChild(
 			document.createTextNode(
-				"QUnit " + QUnit.version  + "; " + navigator.userAgent
+				"QUnit " + QUnit.version	+ "; " + navigator.userAgent
```
jzaefferer: What is this change? GitHub displays the green "added" diff as unselectable, weird.

reporter/html.js
```diff
@@ -460,6 +460,18 @@ function storeFixture() {
 	}
 }
 
+function appendFilteredTest() {
+	var testId = QUnit.config.testId;
+	if ( testId && testId.length > 0 ) {
+		return "<div id='qunit-filteredTest'>Running filtered test: " + testId.join(", ") +
```
jzaefferer: I wonder how useful it really is to show the IDs here. They're also in the URL, and rather useless by themselves.

Duplicating the test name won't be useful, since those are already shown below anyway. How about this? "Rerunning selected tests"

reporter/html.js
```diff
@@ -460,6 +460,18 @@ function storeFixture() {
...
+			" <a id='qunit-clearFilter' href='" +
+			setUrl({ filter: undefined, module: undefined, testId: undefined }) +
+			"'>" + "Clear filter" + "</a></div>";
```
jzaefferer: How about this? "Rull all tests"

reporter/html.js
```diff
@@ -460,6 +460,18 @@ function storeFixture() {
...
+	} else {
```
jzaefferer: You don't need an else after a return.

src/qunit.css
```diff
@@ -68,6 +68,12 @@
 	overflow: hidden;
 }
 
+#qunit-filteredTest {
+	padding: 0.5em 1em 0.5em 1em;
+	background-color: #F4FF77;
+	color: #366097;
```
jzaefferer: This looks like the color we also use for the "skipped" badge, which seems like a good idea, but I don't like what it looks like when it goes over the whole page. I tried some other colors but couldn't find anything that looks good. Reusing existing background colors like `#D2E0E6` works, but makes this too easy to miss.

test/reporter-html/single-testid.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Main Test Suite</title>
+	<link rel="stylesheet" href="../../dist/qunit.css">
+	<script src="../../dist/qunit.js"></script>
+  <script src="single-testid.js"></script>
```
jzaefferer: Bad indent, use tabs instead of spaces.

test/reporter-html/single-testid.js
```diff
@@ -0,0 +1,12 @@
+QUnit.config.testId = [ "2e48c6fa", "9ccf6855" ];
+
+QUnit.test( "Check for changed header after running filtered test", function( assert ) {
+  var html = document.getElementById( "qunit-filteredTest" ).innerHTML;
```
jzaefferer: Same here, indent with tabs.

PS: Why isn't jscs catching that?

reporter/html.js
arghgr: Not sure what happened there. Turning it back to a space.

reporter/html.js
arghgr: One of the reasons this issue was created was because many people who are starting out on this framework wouldn't think to check the URL, so in terms of UX, I think including the test IDs here wouldn't be redundant.

reporter/html.js
arghgr: Ok, changing that-- @leobalter and figured "Clear filter" made sense given the function's name, but "Run all tests" is more straightforward.

reporter/html.js
arghgr: We added this because if the function doesn't return anything, the div will show up as "undefined".

src/qunit.css
arghgr: Not sure what else might look good either, but I'd be happy to take suggestions.

test/reporter-html/single-testid.html
arghgr: Done!

test/reporter-html/single-testid.js
arghgr: Not sure either, changing it manually.

reporter/html.js
jzaefferer: Good point. Let's leave it then.

reporter/html.js
jzaefferer: Sure, but you can drop the `else`, rougly like this:
```js
if ( condition ) {
  return stuff;
}
return "";
```
I usually prefer the early return, or "guard clause":
```js
if ( !condition ) {
  return "";
}
/* lots of processing */
return stuff;
```

test/reporter-html/single-testid.html
jzaefferer: When you make a change based on a review, it helps if you immediately push that change, then the PR and diff are updated and I can see the result, instead of coming back to it later. That is, you say you're done, but I currently can't review your fix, since its still on your local machine (I assume).

test/reporter-html/single-testid.js
jzaefferer: Looks like our version of grunt-jscs is very much outdated. Could you look into updating that to latest? Seperate issue/PR though.

reporter/html.js
```diff
@@ -460,6 +460,18 @@ function storeFixture() {
...
 }
 
+function appendFilteredTest() {
+	var testId = QUnit.config.testId;
+	if ( testId && testId.length > 0 ) {
+		return "<div id='qunit-filteredTest'>Rerunning selected tests: " + testId.join(", ") +
+			" <a id='qunit-clearFilter' href='" +
+			setUrl({ filter: undefined, module: undefined, testId: undefined }) +
+			"'>" + "Run all tests" + "</a></div>";
+	} else {
```
jzaefferer: I'm not sure why my comment is marked as outdated. This else isn't needed. As I wrote before:

You can drop the `else`, roughly like this:
```js
if ( condition ) {
  return stuff;
}
return "";
```
I usually prefer the early return, or "guard clause":
```js
if ( !condition ) {
  return "";
}
/* lots of processing */
return stuff;
```

src/qunit.css
leobalter: Running filtered tests already makes the reporter look different, the badge color might be improved later as I believe the current is not that bad.

leobalter:
LGTM

jzaefferer:
:+1:

leobalter:
Merged! Thanks, @arghgr!

arghgr:
:ok_woman:

PR #877, opened 2015-10-18
Core: Issue 846 semaphore NaN

### LINKED ISSUE #846: Generate a failure when QUnit.start is passed a non-numeric argument
Ref #844

When `QUnit.start` is passed a truthy but non-numeric argument, its [test semaphore decrement](https://github.com/jquery/qunit/blob/d47da30d950ba4b1496e70879eb583e982cbb440/src/core.js#L76) sets `config.current.semaphore` to `NaN`. Such a condition should generate output analogous to the "Called start() while already started…" failure presented when `semaphore` dips below zero.

rdugue:
Fix for `config.current.semaphore` being set to `NaN` if truthy and non-numeric.  Still need to write
tests.

Closes #846

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -111,6 +111,17 @@ extend( QUnit, {
 			// If a test is running, adjust its semaphore
 			config.current.semaphore -= count || 1;
 
+			//If semaphore is non-numeric, throw error
+			if ( isNaN( config.current.semaphore ) ) {
+				config.current.semaphore = 0;
+
+				QUnit.pushFailure(
+					"Called start() with a NaN value",
+					sourceFromStacktrace( 2 )
+				);
+				return;
+			}
+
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -111,6 +111,17 @@ extend( QUnit, {
 			// If a test is running, adjust its semaphore
 			config.current.semaphore -= count || 1;
 
+			//If semaphore is non-numeric, throw error
+			if ( isNaN( config.current.semaphore ) ) {
+				config.current.semaphore = 0;
+
+				QUnit.pushFailure(
+					"Called start() with a NaN value",
```
gibson042: I think I'd use "non-numeric decrement" instead of "NaN value".

gibson042:
This is looking good. Please reach out if the tests give you any trouble.

PR #644, opened 2014-08-30
Events: QUnit logging system uses EventEmitter

JamesMGreene:
Fixes #422.

Even though this is earmarked for `v2.0`, it is 100% backward compatible.

Critique away.  :wink::wine_glass:

PR #882, opened 2015-10-22
Events: Use EventEmitter model for logging system

### LINKED ISSUE #422: Implement QUnit callbacks event listener style
_**Open for discussion!**_

I would like to change the QUnit core [logging callbacks](http://api.qunitjs.com/category/callbacks/) (e.g. `QUnit.done`, etc.) to utilize [theoretical] `on`, `off`, and `emit` methods.

Although we can leave the current callback functions around temporarily for deprecation's sake, the new preferred usage would be as follows (using suggested names):
- `QUnit.begin(fn)`
  - `QUnit.on('run.start', fn)`
- `QUnit.moduleStart(fn)`
  - `QUnit.on('module.start', fn)`
- `QUnit.testStart(fn)`
  - `QUnit.on('test.start', fn)`
- `QUnit.log(fn)`
  - `QUnit.on('assert', fn)`
- `QUnit.testDone(fn)`
  - `QUnit.on('test.done', fn)`
- `QUnit.moduleDone(fn)`
  - `QUnit.on('module.done', fn)`
- `QUnit.done(fn)`
  - `QUnit.on('run.done', fn)`

This change would bring with it all of the usual benefits of an EventEmitter-style setup:
- `on`:
  - Listeners can be added &mdash; current
  - Listeners for custom events can be added, e.g. for use in QUnit addons &mdash; new
- `off`:
  - Listeners can be removed &mdash; new
- `emit`:
  - Listeners can be manually triggered, e.g. for use in QUnit addons &mdash; new

One more important functionality I would prefer to add is the equivalent of a `stopPropagation` method (or flag, or return value) that would allow a listener to prevent subsequently added listeners from being triggered.  The use case here would again be primarily for QUnit addons.

For example, in QUnit Composite, I would like to do something like the following:

``` js
(function(QUnit) {

// Custom event for QUnit Composite
QUnit.on('suite.start', function(data) {
    // log data about a SUITE starting rather than confusing it with a test starting
});

// Custom event for QUnit Composite
QUnit.on('suite.done', function(data) {
    // log data about a SUITE ending rather than confusing it with a test ending
});

QUnit.on('test.start', function(data) {
    if (executingCompositeSuite) {
        QUnit.emit('suite.start', data);
        // Prevent all subsequently added 'test.start' listeners from being triggered
        return false;  // example of a "stopPropagation" functionality via return value
    }
});

QUnit.on('test.done', function(data) {
    if (executingCompositeSuite) {
        QUnit.emit('suite.done', data);
        // Prevent all subsequently added 'test.done' listeners from being triggered
        return false;  // example of a "stopPropagation" functionality via return value
    }
});

// For illustration only; this code already exists [in better form] in PR #408
var executingCompositeSuite = false;
QUnit.extend(QUnit, {
    testSuites: function(suites) {
        // When the first `test` in this set of `suites` starts (i.e. in `setup`), set:
        // executingCompositeSuite = true;

        // Run tests
        // ...

        // When the last `test` in this set of `suites` ends (i.e. in `teardown`), set:
        // executingCompositeSuite = false;
    }
});

})(QUnit);
```

leobalter:
Closes #422
Replaces #644

This is a rebased branch from @JamesMGreene's work on #644 to finish the EventEmitter.

---

[WIP]

### DIFF
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -1,23 +1,28 @@
+var listeners = {};
 var loggingCallbacks = {};
 
-// Register logging callbacks
-function registerLoggingCallbacks( obj ) {
-	var i, l, key,
-		callbackNames = [ "begin", "done", "log", "testStart", "testDone",
+// DEPRECATED: Register old logging callbacks
+// This will be removed on QUnit 2.0.0+
+// From now on use the QUnit.on( eventType, callback ) format
+function registerLoggingCallbacks( QUnit ) {
+	var i, l, key;
+	var callbackNames = [ "begin", "done", "log", "testStart", "testDone",
 			"moduleStart", "moduleDone" ];
+	var dictionary = {
+			"begin": "runStart",
+			"done": "runEnd",
+			"log": "assert",
+			"testStart": "testStart",
+			"testDone": "testEnd",
+			"moduleStart": "suiteStart",
+			"moduleDone": "suiteEnd"
+		};
 
 	function registerLoggingCallback( key ) {
 		var loggingCallback = function( callback ) {
-			if ( objectType( callback ) !== "function" ) {
-				throw new Error(
-					"QUnit logging methods require a callback function as their first parameters."
-				);
-			}
-
-			config.callbacks[ key ].push( callback );
+			return QUnit.on( dictionary[ key ], callback );
 		};
 
-		// DEPRECATED: This will be removed on QUnit 2.0.0+
 		// Stores the registered functions allowing restoring
 		// at verifyLoggingCallbacks() if modified
 		loggingCallbacks[ key ] = loggingCallback;
@@ -28,21 +33,7 @@ function registerLoggingCallbacks( obj ) {
 	for ( i = 0, l = callbackNames.length; i < l; i++ ) {
 		key = callbackNames[ i ];
 
-		// Initialize key collection of logging callback
-		if ( objectType( config.callbacks[ key ] ) === "undefined" ) {
-			config.callbacks[ key ] = [];
-		}
-
-		obj[ key ] = registerLoggingCallback( key );
-	}
-}
-
-function runLoggingCallbacks( key, args ) {
-	var i, l, callbacks;
-
-	callbacks = config.callbacks[ key ];
-	for ( i = 0, l = callbacks.length; i < l; i++ ) {
-		callbacks[ i ]( args );
+		QUnit[ key ] = registerLoggingCallback( key );
 	}
 }
 
@@ -73,3 +64,41 @@ function verifyLoggingCallbacks() {
 		}
 	}
 }
+
+function emit( type, data ) {
+	var i, len, callbacks;
+
+	// Validate
+	if ( QUnit.objectType( type ) !== "string" ) {
+		throw new Error( "Emitting QUnit events requires an event type" );
+	}
+
+	callbacks = listeners[ type ];
+	if ( callbacks ) {
+		for ( i = 0, len = callbacks.length; i < len; i++ ) {
+			callbacks[ i ]( data );
+		}
+	}
+}
+
+QUnit.on = function( type, listener ) {
+
+	// Validate
+	if ( QUnit.objectType( type ) !== "string" ) {
+		throw new Error( "Adding QUnit events requires an event type" );
+	}
+
+	if ( QUnit.objectType( listener ) !== "function" ) {
+		throw new Error( "Adding QUnit events requires a listener function" );
+	}
+
+	// Initialize collection of this logging callback
+	if ( !listeners[ type ] ) {
+		listeners[ type ] = [];
+	}
+
+	// Filter out duplicate listeners
+	if ( inArray( listener, listeners[ type ] ) < 0 ) {
+		listeners[ type ].push( listener );
+	}
+};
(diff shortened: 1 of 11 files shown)


leobalter:
I still need to figure out a way to run both logs and events tests on node.

leobalter:
I never involved myself too much on the js-reporters or the Event Emitter before.

Now that I'm trying to finish this pull request I have some thoughts and concerns:

From js-reporters perspective, we have one non-compliant event: the `assert` event. It's a great feature, but I sounds as a missing piece there.

*js-reporters* is a good project but in my own PoV, it's a great **test report adapter** to connect many test frameworks to many other test tools, like Karma, Grunt, Browserstack, etc. On the other side, I wouldn't like to use it for actual reporters (like html, console, tap, etc).

This PR does not adapt QUnit with the *event details* for js-reporters, and they are not consistent yet: js-reporters/js-reporters#30.

It looks like js-reporters does not support ES3 and even PhantomJS (using Function.prototype.bind without a fallback). QUnit still supports ES3 environments. We can't just ship it along with QUnit 1.x without breaking anything else.

For an actual reporter and adapter, we could also solve it easily if we have an **tap** output. Everything else could parse a tap log, as every test framework should do it just fine as well. Maybe if we have a method to call on a `QUnit.done` or an `'runEnd' ` event, we could return a string containing the whole tap log to be used on any adapter. That does not avoid implementing a real time tap reporter.

---

With that said, my suggestion is to land this PR as it is and work on tap reporter on the top of it (without using js-reporters). The events details should be addressed only after js-reporters/js-reporters#30 is done.

leobalter:
a1f766a solves the problem with running both `events` and `logs` on the node tests.

src/core/logging.js
```diff
@@ -73,3 +64,41 @@ function verifyLoggingCallbacks() {
...
+
+function emit( type, data ) {
+	var i, len, callbacks;
+
+	// Validate
+	if ( QUnit.objectType( type ) !== "string" ) {
+		throw new Error( "Emitting QUnit events requires an event type" );
+	}
+
+	callbacks = listeners[ type ];
```
Krinkle: May wanna call `.slice()` here at the end to ensure a consistent event run.

src/core/logging.js
```diff
@@ -73,3 +64,41 @@ function verifyLoggingCallbacks() {
...
+	if ( callbacks ) {
+		for ( i = 0, len = callbacks.length; i < len; i++ ) {
```
Krinkle: I think caching `.length` has become an anti-pattern now.

leobalter: That's interesting, didn't see it as an anti-pattern anywhere yet. I'll change it anyway.

leobalter:
@Krinkle 1c65763 calls slice() and bf2e02c does not cache `.length`

jzaefferer:
@leobalter I appreciate that you spent some time looking into js-reporters! This looks like a great opportunity to improve the js-reporters spec, instead of partially ignoring it.

Since there a quite a few details to discuss, could we do a Skype/Hangouts call sometime this week? Hopefully @fcarstens can join as well.

leobalter:
> Since there a quite a few details to discuss, could we do a Skype/Hangouts call sometime this week? 

I'm in!

leobalter:
Sure. I still need to work on the event details for this PR as well.

leobalter:
This is rebased and the test-on-node refactoring got its own PR (more rebases on the way).

I still need to finish the work on the events, which will consume some time but it's better to have new events with legacy details.

src/core/logging.js
```diff
@@ -1,23 +1,28 @@
...
-	var i, l, key,
-		callbackNames = [ "begin", "done", "log", "testStart", "testDone",
+// DEPRECATED: Register old logging callbacks
+// This will be removed on QUnit 2.0.0+
+// From now on use the QUnit.on( eventType, callback ) format
+function registerLoggingCallbacks( QUnit ) {
+	var i, l, key;
+	var callbackNames = [ "begin", "done", "log", "testStart", "testDone",
 			"moduleStart", "moduleDone" ];
+	var dictionary = {
```
gibson042: I think this should reference https://github.com/js-reporters/js-reporters#event-names

jzaefferer:
Since #981 was closed, bringing this up here again:

Since it looks like the new event emitter API won't make it into 2.0.0, can we release the new API somewhere in a 2.x release, while keeping the existing callbacks intact? We could then remove the old callbacks in 3.0.

New events with old event data (or both old + new data) seems like a bad idea, so I think it would be better to add the new API with new data, then eventually remove the old API....

flore77:
Hey,

I would like to continue this and to integrate the js-reporters standard into QUnit. How should I start? Should I rebase this onto master ? Thanks.

jzaefferer:
@trentmwillis since you're working on the conversion to ES6 modules, rebasing this now probably makes no sense, right? Any idea how long you'll be working on that?

trentmwillis:
@jzaefferer I'm hoping to have a PR ready within the next couple of days (mainly restructuring/organizing things right now). I don't necessarily think it'd be a bad idea to rebase this as the ES6 modules changes will be minimally invasive into the actual logic (mainly just adding import/export statements).

jzaefferer:
Gotcha, thanks.

@flore77 Let's try an initial rebase soon. If its non-trivial to resolve, we can still discuss when and how to make it work.

trentmwillis:
Closing this in favor of #1026.

Issue #656, opened 2014-09-05
Remove QUnit.asyncTest


JamesMGreene:
Related to #653.

In QUnit `v2.0.0`, `QUnit.asyncTest` will be removed as it doesn't mesh with the new `assert.async` API.

Issue #881, opened 2015-10-21
Throw warnings for deprecated features.


leobalter:
QUnit should throw warnings (reporter/console) for deprecated features, including:
- use of global methods instead of the namespaced methods. (some can be easily fixed with local variables like: `var test = QUnit.test;`)
- use of stop/start methods
- etc

Also it should mention the qunit-migrate tools.

jzaefferer:
I don't think we need an additional migration layer. 2.0 will throw helpful errors for removed features, for anything not yet migrated.

leobalter:
Having warnings for deprecated features on 1.x helps on the migration process, so a run using the last 1.x without any warnings would result on a safe and clean migration to 2.0

leobalter:
After talking on a meeting today with @jzaefferer, we decided to keep this issue and land it before QUnit 2.0

jzaefferer:
There should be an option to disable the warnings. For example, in jQuery UI there's the $.uiBackCompat` flag, which turns off the back-compat layer, to test if the next release (where deprecated features are dropped) will still work. That's different from warnings, but we can do something similar here.

jzaefferer:
Is this still valid?

leobalter:
not landing on 2.0 but useful to guide the Warnings API as the request is to get proper deprecation warning messages on the reporter, not only an exception pushing a failure, even if it's well handled by QUnit.

trentmwillis:
I'm going to close this. We added `Logger.warn` internally for logging deprecations.

PR #885, opened 2015-10-25
Test: Deprecate the 'expected' argument of QUnit.test

### LINKED ISSUE #356: Deprecate expect argument from test() and asyncTest()
Use expect(x) or assert.expect(x) instead.

berkerpeksag:
Fixes #356 

I left asyncTest untouched because of https://github.com/jquery/qunit/issues/656.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -550,6 +550,13 @@ function test( testName, expected, callback, async ) {
 	if ( arguments.length === 2 ) {
 		callback = expected;
 		expected = null;
+	} else if ( arguments.length === 3 && typeof expected !== "function" ) {
+		if ( global.console && global.console.warn ) {
+			global.console.warn(
+				"'expected' argument of QUnit.test is deprecated. " +
+				"Call the expect() function explicitly instead."
+			);
+		}
 	}
 
 	newTest = new Test({
--- a/src/test.js
+++ b/src/test.js
@@ -57,7 +57,8 @@ Test.prototype = {
 					failed: config.moduleStats.bad,
 					passed: config.moduleStats.all - config.moduleStats.bad,
 					total: config.moduleStats.all,
-					runtime: now() - config.moduleStats.started
+					runtime: now() - config.moduleStats.started,
+					warnings: QUnit.warnings
 				});
 			}
 			config.previousModule = this.module;
@@ -218,7 +219,8 @@ Test.prototype = {
 			source: this.stack,
 
 			// DEPRECATED: this property will be removed in 2.0.0, use runtime instead
-			duration: this.runtime
+			duration: this.runtime,
+			warnings: QUnit.warnings
 		});
 
 		// QUnit.reset() is deprecated and will be replaced for a new
@@ -280,7 +282,8 @@ Test.prototype = {
 				expected: expected,
 				testId: this.testId,
 				negative: negative || false,
-				runtime: now() - this.started
+				runtime: now() - this.started,
+				warnings: QUnit.warnings
 			};
 
 		if ( !result ) {
@@ -312,7 +315,8 @@ Test.prototype = {
 				message: message || "error",
 				actual: actual || null,
 				testId: this.testId,
-				runtime: now() - this.started
+				runtime: now() - this.started,
+				warnings: QUnit.warnings
 			};
 
 		if ( source ) {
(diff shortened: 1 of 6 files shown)


src/test.js
```diff
@@ -550,6 +550,13 @@ function test( testName, expected, callback, async ) {
 	if ( arguments.length === 2 ) {
 		callback = expected;
 		expected = null;
+	} else if ( arguments.length === 3 && typeof expected !== "function" ) {
+		if ( global.console && global.console.warn ) {
+			global.console.warn(
+				"'expected' argument of QUnit.test is deprecated. " +
+				"Call the expect() function explicitly instead."
```
leobalter: This is a great start for #861, but we need to make a better structure out of it.

QUnit 1.x.x still supports browsers and other environments without the `console` object, or not even the `console.warn`.

I believe we should make a better structure first before triggering the warnings, like prefixing all the deprecation messages with an all caps `DEPRECATED`, also the HTML Reporter might get this info from the logs and show it on the reporter messages as well.

Also, I would like to have an internal method for the warning messages instead of checking for `console.warn` everywhere.

test/main/test.js
```diff
@@ -7,6 +7,10 @@ QUnit.test( "expect query and multiple issue", function( assert ) {
 	assert.ok( true );
 });
 
+QUnit.test( "expected of QUnit.test is deprecated", 42, function( assert ) {
+	assert.expect( 0 );
+});
+
```
leobalter: This test triggers a false positive for the changes coverage. We need to actual test if it's warning anything. That's where having a warning details on the log might help.

src/test.js
berkerpeksag: Thanks for the review.

>  like prefixing all the deprecation messages with an all caps DEPRECATED

Done.

> also the HTML Reporter might get this info from the logs and show it on the reporter messages as well.

I'm not sure my design is the best one, but I came up with the following:

* Add a QUnit.deprecationWarnings list
* In deprecateWarns helper (added in core/utilities.js), use console.warn if available and add warnings to QUnit.deprecationWarnings
* In HTML reporter, use QUnit.deprecationWarnings to show warning list and total count of warnings(e.g. Tests completed in 1092 milliseconds. 511 assertions of 513 passed, 2 failed.1 warnings)

Does that sounds good to you?

> Also, I would like to have an internal method for the warning messages instead of checking for console.warn everywhere.

Done. Added in core/utilities.js.

test/main/test.js
berkerpeksag: Agreed. Do you have an API design to test warnings. For example:

```js
QUnit.test( "expected of QUnit.test is deprecated", 42, function( assert ) {
    assert.equal(
        assert.warnings[0],
        "'expected' argument of QUnit.test is deprecated." +
        "Call the expect() function explicitly instead."
    );
});
```

src/test.js
leobalter: > Does that sounds good to you?

It sounds great, but I would use a `QUnit.warnings` instead of `deprecationWarnings`, with that we can reuse it to any other warning type, even if it's not a deprecated one.

> Done. Added in core/utilities.js.

Did you pushed the commit with it? I would love to see it.

test/main/test.js
leobalter: Not only registering the warnings on a list/array, QUnit.log should include it.

src/test.js
berkerpeksag: > Did you pushed the commit with it? I would love to see it.

No, I was waiting for your feedback for my other comments. I will update the PR in a day or two.

Thanks!

leobalter: This warnings API will solve #881 partially, and it'll be crucial for the 2.0 release :+1:

leobalter:
Hey @berkerpeksag, I'm looking forward to see this when it's ready.

berkerpeksag:
My apologies, I have a job interview this week. I will work on the PR over the weekend.

leobalter:
No problem. Thanks for the update and the best of luck on the interview.

test/main/test.js
berkerpeksag: Done.

src/test.js
berkerpeksag: Done.

berkerpeksag:
Sorry about the delay.

I'm not so sure about storing all warnings in a global `QUnit.warnings` array. Perhaps it should be stored as `config.moduleStats.warnings` or something similar?

leobalter:
we got a conflict with #918, I need to figure out how to solve it.

leobalter:
@berkerpeksag I'm sorry for the delay and mess on this process. The warning api is not going to land for this deprecation. Thanks for the support.

Issue #788, opened 2015-03-18
improve grunt watch task


leobalter:
In a private chat I saw @jzaefferer's will to improve speed on our grunt watch task. 

Current options to try (not exclusive):

- grunt-contrib-watch [spawn and interrupt](https://github.com/gruntjs/grunt-contrib-watch#optionsspawn) options.
- [grunt-concurrent](https://www.npmjs.com/package/grunt-concurrent)
- [grunt-newer](https://www.npmjs.com/package/grunt-newer)

Some goals:
- grunt watch needs to be faster
- don't rebuild src files when only test files are modified
- can we use those tools to run jshint/jscs only at the modified files (src or each test files)
- run tests in parallel (if this goes faster)

rdugue:
I made some changes on a branch to run jshint/jscs only modified files but it is actually making grunt watch run ~2 seconds slower for me.  You can see the change I made [here.](https://github.com/rdugue/qunit/blob/issue-788/Gruntfile.js#L193)  I'm going to test some more.

leobalter:
Closed by #886

PR #886, opened 2015-10-25
Build: Modifies Gruntfile to improve grunt watch

rdugue:
I added spawn and interrupt options to watch task, and added a concurrent task.  I tried using grunt-newer for jshint/jscs but it did not improve speed, maybe I need to play with it some more.

Ref #788

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -64,9 +64,9 @@ grunt.initConfig({
 		options: {
 			config: ".jscsrc"
 		},
-		all: [
-			"<%= jshint.all %>",
-			"!test/main/deepEqual.js"
+		all:	[
+				"<%= jshint.all %>",
+				"!test/main/deepEqual.js"
 		]
 	},
 	search: {
@@ -119,9 +119,25 @@ grunt.initConfig({
 			src: "build/report/lcov/lcov.info"
 		}
 	},
+	concurrent: {
+		target1: [
+			  "build"
+		],
+		target2: [
+			"jshint",
+			"jscs",
+			"search",
+			"qunit"
+		],
+		target3: [
+			"test-on-node"
+		]
+	},
 	watch: {
 		options: {
-			atBegin: true
+			atBegin: true,
+			spawn: false,
+			interrupt: true
 		},
 		files: [
 			".jshintrc",
@@ -190,6 +206,7 @@ grunt.registerTask( "test-on-node", function() {
 });
 
 grunt.registerTask( "build", [ "concat" ] );
-grunt.registerTask( "default", [ "build", "jshint", "jscs", "search", "qunit", "test-on-node" ] );
+grunt.registerTask( "default", [ "concurrent:target1", "concurrent:target2", "concurrent:target3" ]
+ );
 
 };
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -120,9 +120,25 @@ grunt.initConfig({
...
+			  "build"
+		],
+		target2: [
+			"jshint",
+			"jscs",
+			"search",
+			"qunit"
+		],
+		target3: [
+			"test-on-node"
```
leobalter: instead of numbers I would rename them like:

- target1 => build
- target2 => test
- target3 => node

Also, I'm not sure if this might help, but how about having the qunit task as an individual test, before the test-on-node?

package.json
```diff
@@ -31,6 +31,7 @@
     "browserstack-runner": "0.3.7",
     "commitplease": "2.0.0",
     "grunt": "0.4.2",
+    "grunt-concurrent": "^2.0.3",
```
leobalter: As you can see on the other modules, we lock their versions, without the `^` prefix. Would you lock this one as well, please?

Gruntfile.js
```diff
@@ -64,9 +64,9 @@ grunt.initConfig({
 		options: {
 			config: ".jscsrc"
 		},
-		all: [
-			"<%= jshint.all %>",
-			"!test/main/deepEqual.js"
+		all:	[
+				"<%= jshint.all %>",
+				"!test/main/deepEqual.js"
```
leobalter: We've got some extra whitespace here.

leobalter:
Only some details to change or discuss, otherwise it looks good.

I might be wrong, but it looks it not only improves the speed on grunt watch, but also the regular runs on grunt, right?

Gruntfile.js
rdugue: I changed the target names as suggested.  I tried having qunit as it's own target but it actually runs slower fr me.

package.json
rdugue: Fixed.

rdugue:
That is correct.  The concurrent tasks run with just grunt also.

leobalter:
@rdugue I pulled your branch to check on the speed gain and after trying this config setting below, I was able to free extra 2+ seconds.

```js
	concurrent: {
		build: [
				"concat:src-js",
				"concat:src-css"
		],
		test: [
			"jshint",
			"jscs",
			"search",
			"qunit",
			"test-on-node"
		]
	},
```

It might work differently on other machines, as this plugin uses the available CPU cores. 

You might notice I moved the *test-on-node* task to the concurrent:test and also splitted the *concat* tasks on concurrent:build. I also had to remove the concurrent:node from the default task, so now it is just:

```js
grunt.registerTask( "default", [ "concurrent:build", "concurrent:test" ] );
```

---

If you like and agree, would you apply these changes?

rdugue:
@leobalter I tried that configuration just now and it is indeed faster fr me as well.  Not sure why I didn't try it that way now that I think about it.  I applied the changes.

leobalter:
Thanks, @rdugue! It's merged at 30197a8

PR #887, opened 2015-10-28
Update .travis.yml

paladox:
Test against node_js version 0.12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - "0.12"
   - "0.10"
 before_install:
   - npm install -g grunt-cli
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - "4"
   - "0.12"
   - "0.10"
 before_install:


paladox:
Ive just signed it.

leobalter:
It would be better if node 4.1 is included.

Also: we need to run the browserstack check only on one of them, and that's the case to use in node 4.1.

paladox:
If we do 4 it will automatically go for the latest stable 4.x release available.

leobalter:
Without handling the browsertack runner to run on a single node environment, I can't move this on. Otherwise each CI run will overload browserstack.

paladox:
How would I get browserstack to check only 4.1 or 4.2.

leobalter:
@paladox Honestly I don't know a better way, which probably exists. Maybe checking if the current version (`node -v` or `process.versions.node`) is not 0.10 or 0.12.

leobalter:
also, it would be great if you apple the [style guidelines](http://contribute.jquery.org/) on the commits

leobalter:
We need this PR to get in shape before a next review. I'll close it for now and we can reopen it when if it gets ready again.

mgol:
A version of this PR that runs the `browserstack` task only in one Node.js version: #997.

PR #903, opened 2015-12-09
HTML Reporter: Wrong expected output for notOk

### LINKED ISSUE #899: notOk assertion has wrong label
1. Write test with assert.notOk(true);
2. Run test and see expected: "NOT false", but really expected is "NOT true"

In the following example, $item has css-class "dx-state-focused", so, expression is TRUE, but the label is "NOT false"

uf4uqnr

MMD-Coding:
Fixes #899

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -84,7 +84,7 @@ QUnit.assert = Assert.prototype = {
 	notOk: function( result, message ) {
 		message = message || ( !result ? "okay" : "failed, expected argument to be falsy, was: " +
 			QUnit.dump.parse( result ) );
-		this.push( !result, result, false, message, true );
+		this.push( !result, result, false, message );
 	},
 
 	equal: function( actual, expected, message ) {


src/assert.js
```diff
@@ -84,7 +84,7 @@ QUnit.assert = Assert.prototype = {
 	notOk: function( result, message ) {
 		message = message || ( !result ? "okay" : "failed, expected argument to be falsy, was: " +
 			QUnit.dump.parse( result ) );
-		this.push( !result, result, false, message, true );
+		this.push( !result, result, false, message );
```
leobalter: It's weird how `notOk` is a different kind of a negative assertion. It would be negative if the expected value was true, but we are actually checking if "*actual is equivalent to false*", and not "*actual is __not__ equivalent to false*".

This fix ends up reasonable.

leobalter:
LGTM

@MMD-Coding, would you mind to fix the [CLA issue](https://github.com/jquery/qunit/pull/903#issuecomment-163367842)?

MMD-Coding:
@leobalter I would like to, but it says:
> MMD-Coding <...>: The name MMD-Coding requires manual verification.

Should I have used my real name? Is there anything I can do?

leobalter:
For legal reasons, it's necessary to use a legal name and last name on the git commits.

The CLA page has the steps to fix it. => http://contribute.jquery.org/CLA/status/?repo=qunit&sha=d29b9bfb12be8d040b54d522d1616f1f3754f308

MMD-Coding:
Got it. Thank you!

leobalter:
awesome! thanks!

leobalter:
Merged, thanks!

Issue #895, opened 2015-11-26
QUnit DeepEqual using two strings


joeJazdzewski:
I was testing some code using qunit my test kept failing so I tested 

    var str = new String("hey there");
    var s = new String("hey there");
    assert.deepEqual(s, str);
    assert.equal(s, str);

both equal trys failed

platinumazure:
`new String` will result in the types actually being objects, not strings. So I would expect `assert.equal` to fail since `str == s` is false. As for `deepEqual` failing, my guess is your implementation stores an ID (or hash code, or...) within the string object, and so `deepEqual` will fail as well when it compares property-to-property.

Try using string literals rather than objects. For example, if you have a function that returns `"hey there"`, the correct assertion would be:

```js
var result = myHeyThereFunction();
assert.equal(result, "hey there", "Got 'hey there' back!");
// Or better:
assert.strictEqual(result, "hey there", "Got 'hey there' back!");
```

leobalter:
I can't see a good reason to instantiate string objects using the new String form. If you really need to stick with it, keep in mind different object instances will not pass on the simple equal comparison, if what you want to compare is the value, you can catch its value instead.

```js
var s = new String( "hey there" )
s; // a String typed object, with a collection of each character
typeof s; // "object"

s.toString(); // the value of this object when converted to a simple regular string.
typeof s.toString(); // "string"
```

Both in `==` (equivalent to `assert.equal`) and `===` (equivalent to `assert.strictEqual`) you will only have `true` returning if both objects share the same reference.

What we can still check is the `deepEqual` assertion, as it should check for the object props.

```js
  assert.deepEqual(new String("foo"), new String("foo")); // fails
  assert.deepEqual(new String(), new String()); // fails
  assert.deepEqual(new Object(), new Object()); // pass
  assert.deepEqual({foo: "bar"}, {foo: "bar"}); // pass
```

PR #905, opened 2015-12-15
Assert: Support compare primitive objects in deepEqual

mixed:
Ref #895

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -10,20 +10,15 @@ QUnit.equiv = (function() {
 	var parentsB = [];
 
 	function useStrictEquality( b, a ) {
-
-		/*jshint eqeqeq:false */
-		if ( b instanceof a.constructor && a instanceof b.constructor ) {
-
-			// Compare same constructor.
-			return a.toString() == b.toString();
-		} else if ( b instanceof a.constructor || a instanceof b.constructor ) {
-			// To catch short annotation VS 'new' annotation of a declaration. e.g.:
-			// `var i = 1;`
-			// `var j = new Number(1);`
-			return a == b;
-		} else {
-			return a === b;
+		if ( typeof a === "object" ) {
+			a = a.valueOf();
 		}
+
+		if ( typeof b === "object" ) {
+			b = b.valueOf();
+		}
+
+		return a === b;
 	}
 
 	function compareConstructors( a, b ) {
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -10,8 +10,10 @@ QUnit.equiv = (function() {
 	var parentsB = [];
 
 	function useStrictEquality( b, a ) {
+
 		/*jshint eqeqeq:false */
 		if ( b instanceof a.constructor && a instanceof b.constructor ) {
+
 			// Compare same constructor.
 			return a.toString() == b.toString();
 		} else if ( b instanceof a.constructor || a instanceof b.constructor ) {
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -10,10 +10,11 @@ QUnit.equiv = (function() {
 	var parentsB = [];
 
 	function useStrictEquality( b, a ) {
-
 		/*jshint eqeqeq:false */
-		if ( b instanceof a.constructor || a instanceof b.constructor ) {
-
+		if ( b instanceof a.constructor && a instanceof b.constructor ) {
+			// Compare same constructor.
+			return a.toString() == b.toString();
+		} else if ( b instanceof a.constructor || a instanceof b.constructor ) {
 			// To catch short annotation VS 'new' annotation of a declaration. e.g.:
 			// `var i = 1;`
 			// `var j = new Number(1);`
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -1584,10 +1584,10 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
 QUnit.test( "Compare Number values", function( assert ) {
 	var SafeNumber = Number;
 
-	assert.ok( QUnit.equiv(new SafeNumber( 1 ), new SafeNumber( 1 ) ),
+	assert.ok( QUnit.equiv( new SafeNumber( 1 ), new SafeNumber( 1 ) ),
 		"Number objects with same values are equivalent."
 	);
-	assert.notOk( QUnit.equiv(new SafeNumber( 1 ), new SafeNumber( 2 ) ),
+	assert.notOk( QUnit.equiv( new SafeNumber( 1 ), new SafeNumber( 2 ) ),
 		"Number objects with different values are not equivalent."
 	);
 });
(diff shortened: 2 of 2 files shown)


test/main/deepEqual.js
```diff
@@ -1581,6 +1581,35 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
 	}
 );
 
+QUnit.test( "Compare primitive objects", function( assert ) {
+	var SafeNumber = Number,
+	SafeString = String,
+	SafeBoolean = Boolean;
```
leobalter: why saving these references? Does not look necessary.

mixed: @leobalter 
For lint validation. like it [lint](https://github.com/jquery/qunit/blob/master/test/main/deepEqual.js#L76)

:)

test/main/deepEqual.js
```diff
@@ -1581,6 +1581,35 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
 	}
 );
 
+QUnit.test( "Compare primitive objects", function( assert ) {
```
leobalter: We could split these tests in a module and create a test for each type.

leobalter: This should also be called "Compare primitive values", they are not primitive objects.

mixed: @leobalter 
OK. I going to update.

test/main/deepEqual.js
```diff
@@ -1581,6 +1581,35 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	var SafeNumber = Number,
+	SafeString = String,
+	SafeBoolean = Boolean;
+
+	assert.equal( QUnit.equiv(new SafeNumber(1), new SafeNumber(1) ), true,
+		"Number: When create same parameter should be equal."
+	);
```
leobalter: We may also simplify these tests using the ok and notOk assertions and/or using a single line for each parameter:

```js
assert.equal(
	QUnit.equiv( new SafeNumber(1), new SafeNumber(1) ),
	true,
	"Number: When create same parameter should be equal."
);

// or simply:
assert.ok(
	QUnit.equiv( new SafeNumber( 1 ), new SafeNumber( 1 ) ),
	"Number: When create same parameter should be equal."
);
```

leobalter: Also: the [code style](http://contribute.jquery.org/style-guide/js/) requires a space before and after the arguments list, so `new SafeNumber(1)` => `new SafeNumber( 1 )`

mixed: @leobalter 
Thank you for your advice. I going to update all review.

test/main/deepEqual.js
```diff
@@ -1581,6 +1581,35 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
...
+	assert.equal( QUnit.equiv(new SafeNumber(1), new SafeNumber(2) ), false,
+		"Number: When create different parameter should be not equal."
```
leobalter: Maybe replace the assertion message with:

"Number objects with different values are not equivalent" or "QUnit.equiv returns false for Number objects with different values"

test/main/deepEqual.js
leobalter: Oops, makes sense. 

We should fix the code style here, that's relatively new on the code style, but we should declare all the variables with a new var statement.

```js
var SafeNumber = Number;
var SafeString = String;
var SafeBoolean = Boolean;
```

leobalter:
I know I commented on several details, but this is a great work adding a great value to QUnit.equiv.

I'm looking forward to merge this PR. Thanks, @mixed!

mixed:
@leobalter 
Thank you. Good review for me.

test/main/deepEqual.js
```diff
@@ -1581,6 +1581,39 @@ QUnit.test( "Test that must be done at the end because they extend some primitiv
 	}
 );
 
+QUnit.test( "Compare Number values", function( assert ) {
```
leobalter: @mixed: would you create a single module on the top of these tests called "Compare primitive values"?

Example:

```js
QUnit.module( "Compare primitive values" );
```

src/equiv.js
```diff
@@ -10,10 +10,11 @@ QUnit.equiv = (function() {
 	var parentsB = [];
 
 	function useStrictEquality( b, a ) {
-
```
leobalter: this blank line should be kept, all comments (even the jshint ones) should be preceded by a blank line.

src/equiv.js
```diff
@@ -10,10 +10,11 @@ QUnit.equiv = (function() {
...
 		/*jshint eqeqeq:false */
-		if ( b instanceof a.constructor || a instanceof b.constructor ) {
-
+		if ( b instanceof a.constructor && a instanceof b.constructor ) {
+			// Compare same constructor.
```
leobalter: missing a preceding blank line here

leobalter:
We're close!

src/equiv.js
```diff
@@ -12,8 +12,11 @@ QUnit.equiv = (function() {
 	function useStrictEquality( b, a ) {
 
 		/*jshint eqeqeq:false */
-		if ( b instanceof a.constructor || a instanceof b.constructor ) {
+		if ( b instanceof a.constructor && a instanceof b.constructor ) {
 
+			// Compare same constructor.
+			return a.toString() == b.toString();
```
gibson042: I think this could be generalized and simplified a bit further, to something like
```js
if ( typeof a === "object" ) {
	a = a.valueOf();
}
if ( typeof b === "object" ) {
	b = b.valueOf();
}
return a === b;
```

mixed: @gibson042 
Thank you. I think of it again. I think this don't have to check type. Could you please explain to me more detail about check type? And How about think below logic?
```js
function useStrictEquality( b, a ) {
	if ( b instanceof a.constructor || a instanceof b.constructor ) {
		return a.valueOf() === b.valueOf();
	} else {
		return a === b;
	}
}
```

gibson042: Well, there are two issues. First, and most importantly, we should probably have an automatic `return true` in a separate `guaranteed` function for `null`, `undefined`, and `nan` "types". Having cleared those out of the way, there would be a few options for detecting wrapped primitives, and I like `typeof object` because it strikes me as the most semantic. And separate from _all_ of that, it's just nice to see a single `return` with uniform comparison logic.

mixed: @gibson042

Thank you. :)
I have misunderstood it. but I get it now. Check constuctor and check typeof before I understand your comment.

Good comments. I updating it soon.

leobalter:
@gibson042 can we merge this?

gibson042:
I increased test coverage a little and restored an explanatory comment, but yes! :grinning:

leobalter:
This is awesome! thanks, @mixed and @gibson042

Issue #904, opened 2015-12-14
Regex test filtering


trentmwillis:
Would it be possible to introduce regex test filtering? There are cases where we want a subset of our test suite to run but can't achieve it with a simple string filter. Maybe my thinking is naive, but it seems like we should be able to take the filter string and cast it to a regular expression and test the test names against it.

This would probably require an additional config option to treat the filter as a regular expression as opposed to a string value.

leobalter:
This is a good question and I would like to discuss before making a decision.

I currently have some limited filters as filtering by the test names, the module filters and using `QUnit.only` for single test blocks.

It does not look like a problem to cast the filter string to a regular expression.

QUnit.pending (not implemented yet) would be helpful here but it targets a different strategy.

leobalter:
The config option might be too much, we should try casting the filter string as a regular expression without any option to enable or disable it, the option would be an unnecessary complexity.

gibson042:
Using the existing parameter is fine by me, but we'll need some in-band method of differentiating string matches from regex patterns (similar to the preexisting [leading `!` for filter inversion](https://github.com/jquery/qunit/blob/efe03030a6f22ff002de8ee6164b74af74018254/src/test.js#L392-L395)) so regex-special characters can still be matched literally. E.g., something like `filter=/regexp?/i` (optional "p") vs. `filter=regexp?` (literal "?").

leobalter:
seems right.

We may also trigger the regexp match if the filter value matches a valid regexp literal notation (using the `/content/params?`

I'm interested to see how it might be implemented. @trentmwillis would you like to try a PR with it?

trentmwillis:
These cases are actually why I thought an additional config option might be valuable as it takes some of the guess work out of determining what sort of pattern we're using. For example, what I'm seeing with the above is:

`/components/g` -> Does a global regex search for `components`
`/components/s` -> Does a regex search for `/components/s`?

I think trying to determine if we should be treating values as regex patterns or literals would add uncertainty to the end user.

gibson042:
The mechanism doesn't matter to me, as long as behavior is predictable and we maintain backwards compatibility with respect to non-initial special characters.

trentmwillis:
Did an implementation that added an additional config option. @leobalter if you feel strongly that we should not add another config option, I can explore other alternatives.

jzaefferer:
Btw. I feel strongly that we should not add another config option.

trentmwillis:
[...]

PR #906, opened 2015-12-16
Core: Support filtering by regular expression

trentmwillis:
Introduces a "regexFilter" config option that, when enabled, treats the "filter"
value as a regular expression to test against instead of a string.

Fixes: #904

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -356,10 +356,10 @@ Test.prototype = {
 	},
 
 	valid: function() {
-		var include,
-			filter = config.filter && config.filter.toLowerCase(),
+		var filter = config.filter,
+			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
 			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
-			fullName = ( this.module.name + ": " + this.testName ).toLowerCase();
+			fullName = ( this.module.name + ": " + this.testName );
 
 		function testInModuleChain( testModule ) {
 			var testModuleName = testModule.name ? testModule.name.toLowerCase() : null;
@@ -389,7 +389,23 @@ Test.prototype = {
 			return true;
 		}
 
-		include = filter.charAt( 0 ) !== "!";
+		return regexFilter ?
+			this.regexFilter( !!regexFilter[1], regexFilter[2], regexFilter[3], fullName ) :
+			this.stringFilter( filter, fullName );
+	},
+
+	regexFilter: function( exclude, pattern, flags, fullName ) {
+		var regex = new RegExp( pattern, flags );
+		var match = regex.test( fullName );
+
+		return match !== exclude;
+	},
+
+	stringFilter: function( filter, fullName ) {
+		filter = filter.toLowerCase();
+		fullName = fullName.toLowerCase();
+
+		var include = filter.charAt( 0 ) !== "!";
 		if ( !include ) {
 			filter = filter.slice( 1 );
 		}
--- a/test/regex-filter.html
+++ b/test/regex-filter.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>QUnit Regex Filtering Test Suite</title>
+		<link rel="stylesheet" href="../dist/qunit.css">
+		<script src="../dist/qunit.js"></script>
+		<script>
+			QUnit.config.filter = '/foo|bar/i';
+		</script>
+		<script src="regex-filter.js"></script>
+	</head>
+	<body>
+		<div id="qunit"></div>
+	</body>
+</html>
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -107,7 +107,10 @@ grunt.initConfig({
 			"test/reporter-html/legacy-markup.html",
 			"test/reporter-html/no-qunit-element.html",
 			"test/reporter-html/single-testid.html",
-			"test/only.html"
+			"test/only.html",
+			"test/regex-filter.html",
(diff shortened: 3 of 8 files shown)


leobalter:
It's great and includes proper tests for it, even though, I believe controlling regex via config/url-param is not greatly discoverable or might be confusing when defined on the code config.

I'm not silver on this, but I still prefer the approach of triggering the regexFilter if the input field has a valid regexp literal notation.

@gibson042 @jzaefferer @rwjblue: would you give some feedback here? I believe this deserves a discussion before shipping.

gibson042:
screen

I really like the tests, and how clean the code is. However, I don't like the [toolbar clutter](https://jsfiddle.net/zos5whbh/)—especially with the separation between the new checkbox and what it affects—and the forced lowercasing not only precludes case-sensitive regular expression filters, but actually introduces bugs (since `\B`, `\D`, `\S`, and `\W` get converted into their negations).

Since we've already set the precedent for in-band signaling with negation-by-leading-`!`, I'm in favor of continuing that by triggering on regex literal syntax like @leobalter suggested (e.g., `isRegex = /^!?\/[\w\W]*\/i?$/.test( filter )`). An alternative would be a `filterType` parameter and one button for each value, but that just doesn't seem as user-friendly.

trentmwillis:
Turning off the forced lowercasing wouldn't be difficult (didn't think to do it before) and will need to happen even if we move to an in-band approach.

I'm not totally opposed to triggering on regex literals, but as I mentioned before that does become a slightly backwards incompatible change (e.g., filtering for `/components/` will now behave differently) and can result in some confusing functionality (e.g., filtering for `/components/i` is treated as a regex, but `/components/a` is not). If those drawbacks are acceptable I can try implementing the other solution.

Another option would be to restructure the toolbar such that toggling regex behavior is associated better with the filter, but that will involve some UX problems to solve.

trentmwillis:
@leobalter @gibson042 after hearing from @jzaefferer that he would prefer not adding a new config option as well, I've updated to check for regex literals.

src/test.js
```diff
@@ -356,8 +356,8 @@ Test.prototype = {
 	},
 
 	valid: function() {
-		var include,
-			filter = config.filter && config.filter.toLowerCase(),
+		var filter = config.filter,
+			regexFilter = /^!?\/([\w\W]*)\/(i?$)/.exec( filter ),
```
leobalter: this regexp also capture strings starting with `!`. IMHO, we may remove it and go with `/^\/([\w\W]*)\/(i?$)/` only.

The `!` prefix flag is good for is strings but not necessary at all for regexp.

Also the current regexFilter function is not handling this prefix.

Removing the `!` flag for regexps is not a breaking feature as it is still operational for regular string searches.

test/regex-filter.html
```diff
@@ -0,0 +1,16 @@
...
+    <script src="../dist/qunit.js"></script>
+    <script>
+    QUnit.config.filter = '/foo|bar/i'
+    </script>
+    <script src="regex-filter.js"></script>
+  </head>
+  <body>
+    <div id="qunit"></div>
+  </body>
+</html>
```
leobalter: Would you change the indentation from space to tabs? The same goes to:

- test/regex-filter.js
- test/string-filter.html
- test/string-filter.js

leobalter:
It's looking good, with only minor changes to go.

src/test.js
trentmwillis: Forgot to include that in the functionality. Personally, I think it would be a nice convenience feature since writing converse regexs can get ugly. I don't see a big downside to including it, but there are some benefits.

leobalter: Seems reasonable. If we include it on the method and add tests for it, I'm game.

src/test.js
```diff
@@ -356,8 +356,8 @@ Test.prototype = {
...
 			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
 			fullName = ( this.module.name + ": " + this.testName ).toLowerCase();
```
gibson042: I think this `.toLowerCase()` should move into `stringFilter`.

trentmwillis: Yeah, caught that too when writing more tests.

trentmwillis:
Updated with support for excluding regex matches and additional tests. Also fixed spacing issues.

src/test.js
```diff
@@ -389,7 +389,23 @@ Test.prototype = {
...
+		return regexFilter ?
+			this.regexFilter( regexFilter[1], regexFilter[2], regexFilter[3], fullName ) :
+			this.stringFilter( filter, fullName );
+	},
+
+	regexFilter: function( exclude, pattern, flags, fullName ) {
+		var regex = new RegExp( pattern, flags );
+		var include = regex.test( fullName );
+
+		return exclude ? !include : include;
```
gibson042: This `include` variable is dependent upon `fullName`, unlike the `include` variable in `stringFilter`. But honestly, I don't think it improves clarity anyway. What about something like
```js
var match = regex.test( fullName );
return match === !exclude;
```
?

`stringFilter` can be similarly cleaned up in another PR.

gibson042:
One last nit, but LGTM!

src/test.js
trentmwillis: Makes sense. I went back and forth on what to name this. Will update.

trentmwillis:
@gibson042 @leobalter PR has been updated with the last minor changes.

trentmwillis:
@leobalter any chance this can land soon?

leobalter:
I'm on vacation this week (from mobile only, no computers around). @gibson042 might be able to help, if not traveling for the holidays as well.

I'll be back on Sunday.

leobalter:
Merged at e972135.

Thanks, @trentmwillis!

PR #912, opened 2015-12-30
Assert: Improve size and speed of QUnit.equiv

gibson042:
Inspired by gh-895 review. Inner comparisons are always `QUnit.objectType`-equal.

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -9,6 +9,12 @@ QUnit.equiv = (function() {
 	var parents = [];
 	var parentsB = [];
 
+	var getProto = Object.getPrototypeOf || function( obj ) {
+
+		/*jshint proto: true */
+		return obj.__proto__;
+	};
+
 	function useStrictEquality( b, a ) {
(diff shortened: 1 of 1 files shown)


src/equiv.js
```diff
@@ -239,37 +217,23 @@ QUnit.equiv = (function() {
 	};
 
 	function typeEquiv( a, b ) {
-		var prop = QUnit.objectType( a );
-		return callbacks[ prop ]( b, a );
+		var type = QUnit.objectType( a );
+		return QUnit.objectType( b ) === type && callbacks[ type ]( b, a );
```
leobalter: Until checking this change above, it is hard to see why the `"nan"` check is returning true only.

leobalter:
It looks good! I'm rerunning to check the tests results on different browsers.

leobalter:
the new file has -36 lines from a previous file with 276 lines. This is great and - due to the code quality - a smart size cut.

Issue #909, opened 2015-12-19
Regular expression comparison ignores ES2015 "unicode" flag


gibson042:
e.g., `/\u{1D306}/u.flags === "u"`.

Using the new [`flags` property](http://www.ecma-international.org/ecma-262/6.0/#sec-get-regexp.prototype.flags) in [`equiv`:`callbacks["regex"]`](https://github.com/jquery/qunit/blob/41322e7a8f521096859cc9c767df088797ed0608/src/equiv.js#L82-L88) should future-proof us against all future additions.

leobalter:
and it seems like a easy one to fix. 

@mixed, would you like to help fixing this?

mixed:
@leobalter 
OK. I have fixed it. :smirk:

PR #916, opened 2016-01-07
Core: Regular expression comparison includes ES2015 "unicode" flag

mixed:
When Regexp has flags It checks flags.
If Regexp hasn`t flags It checks toString.

Ref #909

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -60,7 +60,7 @@ QUnit.equiv = (function() {
 	}
 
 	function getRegExpFlags( regexp ) {
-		return regexp.flags || regexp.toString().match(/[gimuy]*$/)[0];
+		return "flags" in regexp ? regexp.flags : regexp.toString().match( /[gimuy]*$/ )[ 0 ];
 	}
 
 	var callbacks = {
@@ -80,7 +80,7 @@ QUnit.equiv = (function() {
 			return a.source === b.source &&
 
 				// Include flags in the comparison
-				getRegExpFlags(a) === getRegExpFlags(b);
+				getRegExpFlags( a ) === getRegExpFlags( b );
 		},
 
 		// - skip when the property is a method of an instance (OOP)
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -59,6 +59,10 @@ QUnit.equiv = (function() {
 		return false;
 	}
 
+	function getRegExpFlags( regexp ) {
+		return regexp.flags || regexp.toString().match(/[gimuy]*$/)[0];
+	}
+
 	var callbacks = {
 		"string": useStrictEquality,
 		"boolean": useStrictEquality,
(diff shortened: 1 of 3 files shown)


leobalter:
Just small whitespace issues but it seems great.

@gibson042 /r?

src/equiv.js
```diff
@@ -76,10 +80,7 @@ QUnit.equiv = (function() {
 			return a.source === b.source &&
 
 				// Include flags in the comparison
-				a.global === b.global &&
-				a.ignoreCase === b.ignoreCase &&
-				a.multiline === b.multiline &&
-				a.sticky === b.sticky;
+				getRegExpFlags(a) === getRegExpFlags(b);
```
leobalter: missing spaces between arguments list `(a)` => `( a )`...

src/equiv.js
```diff
@@ -59,6 +59,10 @@ QUnit.equiv = (function() {
 		return false;
 	}
 
+	function getRegExpFlags( regexp ) {
+		return regexp.flags || regexp.toString().match(/[gimuy]*$/)[0];
```
leobalter: missing a space between arguments list and inside square brackets:

- `(/[gimuy]*$/)` => `( /[gimuy]*$/ )`
- `[0]` => `[ 0 ]`

gibson042: The approach seems fine, but I'd prefer `"flags" in regexp ? regexp.flags : …` over `regexp.flags || …` to avoid unnecessary execution. Also, I'm not sure that `regexp.toString()` will work as we need on all environments—this needs unit tests to confirm.

JamesMGreene: > I'd prefer `"flags" in regexp ? regexp.flags : …`

:+1:

mixed: @leobalter 
Thank you. When I ran `grunt` It have been success. So I wonder that why this case not found it in `jscs` task. I found cause. Current It normally doesn`t work `jscs task` because `grunt-jscs` version is low. If `jscs` version is high It is normally working.  I opened #917 for this case.
I have a question.  Can i change up version of `grunt-jscs`?  If it do task I going to change files.
I going to apply rules independently of #917.

@gibson042 
> I'd prefer "flags" in regexp ? regexp.flags : …
:+1: 
This pr that CI ran is success. How can i test more environments?

gibson042: The CI coverage is fine, I just want to see additional tests added to [test/main/deepEqual.js](https://github.com/mixed/qunit/blob/enhanceCompareRegexpOption/test/main/deepEqual.js#L447) that include regular expressions with the unicode flag set (skipped in environments that don't support it).

mixed: @gibson042 
Ah.. I got it. I going to add testcode.

mixed: @gibson042 
An error in jshint occurred when i made testcase for `u` option. It working this issue in [js](https://github.com/jshint/jshint/issues/2361) [hint](https://github.com/jshint/jshint/pull/2369). Can i use jshint except options in test code?

gibson042: You can make JSHint exceptions, but it's probably better to define the regular expressions by constructor (e.g., `new RegExp( "foo", "u" )`).

mixed: @gibson042 
Thank you for your review. I've learned a your tips. :+1:

leobalter:
@mixed, That's great!

I'm sorry I wasn't fast enough to answer on the lint issue, but it's good to see it's covered now.

PR #918, opened 2016-01-12
Warnings API for deprecated features

raphamorim:
Ref: #881 
**Please don't merge this PR yet**

I will finalize the PR if you have no problems with this implementation.
- [x] `done()`  return all warnings.
- [x] Deprecated warning to `start`  
- [x] Deprecated warning to `stop` 
- [x] Deprecated warning for `reset` (this might need an extra discussion before removal, but it's deprecated)
- [x] Deprecated warning for `asyncTest`
- [x] Deprecated warning for global call `test()`
- [x] Deprecated warning for `push`
- [x] Deprecated warning for `init`
- [x] Deprecated warning for global call `module()`
- [x] Module hooks setup and teardown calls
- [x] Deprecated warning when use _duration_ property in testDone
- [x] Assert methods on the QUnit and global objects
- [x] Logging callbacks being assigned with new values instead of called as functions

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -123,9 +123,7 @@ grunt.initConfig( {
 			"test/random.html",
 			"test/regex-filter.html",
 			"test/regex-exclude-filter.html",
-			"test/string-filter.html",
-			"test/warning.html",
-			"test/warning-once.html"
+			"test/string-filter.html"
 		]
 	},
 	coveralls: {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -42,6 +42,7 @@ grunt.initConfig( {
 				"src/equiv.js",
 				"src/dump.js",
 				"src/export.js",
+				"src/warning.js",
 				"src/outro.js",
 				"reporter/intro.js",
 				"reporter/urlparams.js",
@@ -122,7 +123,9 @@ grunt.initConfig( {
 			"test/random.html",
 			"test/regex-filter.html",
 			"test/regex-exclude-filter.html",
-			"test/string-filter.html"
+			"test/string-filter.html",
+			"test/warning.html",
+			"test/warning-once.html"
 		]
 	},
 	coveralls: {
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -49,6 +49,8 @@ module.exports = function( grunt ) {
 
 		QUnit.config.autorun = false;
 
+		QUnit.warning.off();
+
 		require( "../../" + file );
 
 		QUnit.load();
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -49,8 +49,6 @@ module.exports = function( grunt ) {
 
 		QUnit.config.autorun = false;
 
-		QUnit.warning.off();
-
 		require( "../../" + file );
 
 		QUnit.load();
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -8,8 +8,6 @@ if ( typeof window === "undefined" || !window.document ) {
 QUnit.init = function() {
 	var config = QUnit.config;
 
-	QUnit.warning( "init" );
-
 	config.stats = { all: 0, bad: 0 };
 	config.moduleStats = { all: 0, bad: 0 };
 	config.started = 0;
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -8,6 +8,8 @@ if ( typeof window === "undefined" || !window.document ) {
 QUnit.init = function() {
 	var config = QUnit.config;
 
+	QUnit.warning( "init" );
+
 	config.stats = { all: 0, bad: 0 };
 	config.moduleStats = { all: 0, bad: 0 };
 	config.started = 0;
--- a/src/core.js
+++ b/src/core.js
@@ -145,6 +145,7 @@ extend( QUnit, {
 
 	// DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.
 	stop: function( count ) {
+
 		// If there isn't a test running, don't allow QUnit.stop() to be called
 		if ( !config.current ) {
 			throw new Error( "Called stop() outside of a test context" );
--- a/src/core.js
+++ b/src/core.js
@@ -145,7 +145,6 @@ extend( QUnit, {
 
 	// DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.
 	stop: function( count ) {
-
 		// If there isn't a test running, don't allow QUnit.stop() to be called
 		if ( !config.current ) {
 			throw new Error( "Called stop() outside of a test context" );
@@ -318,6 +317,7 @@ function done() {
 		failed: config.stats.bad,
 		passed: passed,
 		total: config.stats.all,
+		warnings: config.stats.warnings,
 		runtime: runtime
 	} );
 }
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -55,8 +55,6 @@ function verifyLoggingCallbacks() {
 	for ( loggingCallback in loggingCallbacks ) {
 		if ( QUnit[ loggingCallback ] !== loggingCallbacks[ loggingCallback ] ) {
 
-			QUnit.warning( "loggingCallbacks" );
-
 			userCallback = QUnit[ loggingCallback ];
 
 			// Restore the callback function
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -55,6 +55,8 @@ function verifyLoggingCallbacks() {
 	for ( loggingCallback in loggingCallbacks ) {
 		if ( QUnit[ loggingCallback ] !== loggingCallbacks[ loggingCallback ] ) {
 
+			QUnit.warning( "loggingCallbacks" );
+
 			userCallback = QUnit[ loggingCallback ];
 
 			// Restore the callback function
--- a/src/export.js
+++ b/src/export.js
@@ -16,12 +16,6 @@
 	}
 }() );
 
-function globalWarning( key ) {
-	return function() {
-		return QUnit[ key ].apply( this, arguments );
-	};
-}
-
 // For browser, export only select globals
 if ( defined.document ) {
 
@@ -49,7 +43,7 @@ if ( defined.document ) {
 			];
 
 		for ( i = 0, l = keys.length; i < l; i++ ) {
-			window[ keys[ i ] ] = globalWarning( keys[ i ] );
+			window[ keys[ i ] ] = QUnit[ keys[ i ] ];
 		}
 	}() );
 
--- a/src/export.js
+++ b/src/export.js
@@ -16,6 +16,12 @@
 	}
 }() );
 
+function globalWarning( key ) {
+	return function() {
+		return QUnit[ key ].apply( this, arguments );
+	};
+}
+
 // For browser, export only select globals
 if ( defined.document ) {
 
@@ -43,7 +49,7 @@ if ( defined.document ) {
 			];
 
 		for ( i = 0, l = keys.length; i < l; i++ ) {
-			window[ keys[ i ] ] = QUnit[ keys[ i ] ];
+			window[ keys[ i ] ] = globalWarning( keys[ i ] );
 		}
 	}() );
 
--- a/src/test.js
+++ b/src/test.js
@@ -37,6 +37,7 @@ function Test( settings ) {
 	} else {
 		this.assert = new Assert( this );
 	}
+
 }
 
 Test.count = 0;
@@ -196,6 +197,8 @@ Test.prototype = {
 		this.runtime = now() - this.started;
 		config.stats.all += this.assertions.length;
 		config.moduleStats.all += this.assertions.length;
+
+		config.stats.warnings = QUnit.warning.log;
 
 		for ( i = 0; i < this.assertions.length; i++ ) {
 			if ( !this.assertions[ i ].result ) {
@@ -444,7 +447,6 @@ Use testStart or testDone for custom cleanup.
 This method will throw an error in 2.0, and will be removed in 2.1
 */
 QUnit.reset = function() {
-
 	// Return on non-browser environments
 	// This is necessary to not break on node tests
 	if ( !defined.document ) {
--- a/src/test.js
+++ b/src/test.js
@@ -37,7 +37,6 @@ function Test( settings ) {
 	} else {
 		this.assert = new Assert( this );
 	}
-
 }
 
 Test.count = 0;
--- a/src/test.js
+++ b/src/test.js
@@ -446,6 +446,7 @@ Use testStart or testDone for custom cleanup.
 This method will throw an error in 2.0, and will be removed in 2.1
 */
 QUnit.reset = function() {
+
 	// Return on non-browser environments
 	// This is necessary to not break on node tests
 	if ( !defined.document ) {
--- a/src/warning.js
+++ b/src/warning.js
@@ -1,0 +1,42 @@
+QUnit.warning = ( function() {
+    var messages = [];
+
+    var once = [],
+        disableLogs = false;
+
+    function find( array, warning ) {
+        for ( var i = 0; i < array.length; i++ ) {
+            if ( array[ i ].type === warning ) {
+                return array[ i ];
+            }
+        }
+        return null;
+    }
+
+    function warning( message ) {
+        if ( find( once, message ) ) {
+            return;
+        }
+
+        var warningObj = find( messages, message );
+        if ( !warningObj ) {
+            return;
+        }
+
+        once.push( warningObj );
+        if ( !disableLogs ) {
+            if ( global.console && global.console.warn ) {
+                global.console.warn( warningObj.message );
+            }
+        }
+    }
+
+    function off() {
+        disableLogs = true;
+    }
+
+    warning.log = once;
+    warning.off = off;
+
+    return warning;
+}() );
--- a/test/autostart.html
+++ b/test/autostart.html
@@ -12,7 +12,7 @@
 	var beginData,
 		times = {};
 
-	(function() {
+	( function() {
 		var now = Date.now || function() {
 			return new Date().getTime();
 		};
--- a/test/autostart.html
+++ b/test/autostart.html
@@ -12,8 +12,6 @@
 	var beginData,
 		times = {};
 
-	QUnit.warning.off();
-
 	(function() {
 		var now = Date.now || function() {
 			return new Date().getTime();
--- a/test/autostart.html
+++ b/test/autostart.html
@@ -12,7 +12,9 @@
 	var beginData,
 		times = {};
 
-	( function() {
+	QUnit.warning.off();
+
+	(function() {
 		var now = Date.now || function() {
 			return new Date().getTime();
 		};
--- a/test/headless.html
+++ b/test/headless.html
@@ -8,7 +8,7 @@
 	<script src="main/test.js"></script>
 	<script src="main/deepEqual.js"></script>
 	<script>
-		(function () {
+		( function () {
 
 			function createCallback(logType) {
 				QUnit[logType]( function () {
--- a/test/headless.html
+++ b/test/headless.html
@@ -9,7 +9,6 @@
 	<script src="main/deepEqual.js"></script>
 	<script>
 		(function () {
-			QUnit.warning.off();
 
 			function createCallback(logType) {
 				QUnit[logType]( function () {
--- a/test/headless.html
+++ b/test/headless.html
@@ -8,7 +8,9 @@
 	<script src="main/test.js"></script>
 	<script src="main/deepEqual.js"></script>
 	<script>
-		( function () {
+		(function () {
+			QUnit.warning.off();
+
 			function createCallback(logType) {
 				QUnit[logType]( function () {
 					console.log(logType, arguments);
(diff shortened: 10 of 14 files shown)


src/test.js
```diff
@@ -537,18 +544,31 @@ function checkPollution() {
 
 // Will be exposed as QUnit.asyncTest
 function asyncTest( testName, expected, callback ) {
+	var warning;
+
+	if ( global.console && global.console.warn ) {
+		warning = "DEPRECATED: QUnit.asyncTest() will be removed in QUnit 2.0.\n" +
+			"Reference: https://api.qunitjs.com/QUnit.asyncTest/";
+		global.console.warn(warning);
```
leobalter: the console.warn call needs to be abstracted. Otherwise you'll need to check for it every time you trigger a warning.

Also, the messages could be stored on a dictionary object, easier to keep.

src/test.js
```diff
@@ -537,18 +544,31 @@ function checkPollution() {
...
+	}
+
 	if ( arguments.length === 2 ) {
 		callback = expected;
 		expected = null;
 	}
 
-	QUnit.test( testName, expected, callback, true );
+	QUnit.test( testName, expected, callback, warning, true );
```
leobalter: changing the function signature is not the best approach here.

src/test.js
```diff
@@ -537,18 +544,31 @@ function checkPollution() {
...
 
 // Will be exposed as QUnit.test
-function test( testName, expected, callback, async ) {
+function test( testName, expected, callback, warning, async ) {
 	if ( focused )  { return; }
 
+	if ( this === global ) {
+		warning = "DEPRECATED: Global 'test()' will be removed in QUnit 2.0, " +
+			"use 'QUnit.test() instead";
+	}
```
leobalter: this check is confusing, I also wonder what could happen if I need to copy the function at QUnit.test into a variable.

leobalter:
We should focus initially on the warning abstracted implementation and then on the places we should have it.

The implementation should consider: 

- It can be toggled off (probably trough the config object)
- Avoid changing any other external part of the API, including function signatures
- It might - or not - be exposed (maybe through a `QUnit.warning` method). It's good if it's easy to export.
- The deprecation messages can stored on a object, easier to find and change them if necessary. (e.g.: `QUnit.warning.deprecated`)

In the mean time, I should provide a complete list containing all the deprecated parts, including:

- [`asyncTest`](http://api.qunitjs.com/QUnit.asyncTest/)
- Module hooks [`setup` and `teardown`](http://api.qunitjs.com/QUnit.module/) calls
- Partial deprecation of [`QUnit.start`](http://api.qunitjs.com/QUnit.start/)
- [`QUnit.stop`](http://api.qunitjs.com/QUnit.stop/)
- [`QUnit.init`](http://api.qunitjs.com/QUnit.init/)
- [`QUnit.push`](http://api.qunitjs.com/QUnit.push/)
- [`QUnit.reset`](http://api.qunitjs.com/QUnit.reset/) (this might need an extra discussion before removal, but it's deprecated)
- `QUnit.test` [exposal to the global object](https://github.com/jquery/qunit/blob/master/src/export.js#L25)
- `Qunit.module` [exposal to the global object](https://github.com/jquery/qunit/blob/master/src/export.js#L26)
- Assert methods on the [`QUnit` and global objects](https://github.com/jquery/qunit/blob/master/src/export.js#L4-L20)
- Logging callbacks being [assigned with new values instead of called as functions](https://github.com/jquery/qunit/blob/3592ff19021e3f552e3b2c26fd8b53760f34c6e3/src/core/logging.js#L49-L52)
- testDone [`duration` property](https://github.com/jquery/qunit/blob/eb27714ae090af41cb4585708e562b9d03bbad8d/src/test.js#L223)

Needs discussion or implementation:

- `test` with the [expected parameter signature](https://github.com/jquery/qunit/blob/master/src/test.js#L549)
- `test` with the [async parameter signature](https://github.com/jquery/qunit/blob/master/src/test.js#L549)
- the [callback methods](http://api.qunitjs.com/category/callbacks/) after we get the EventEmitter API done

---

I need to check if anything is missing but this list is at least closer to reflect all the pre-2.0 deprecations.

src/test.js
raphamorim: Updated: 775e9cb2f6d1cce54366494b04349540c7f50b48

src/test.js
raphamorim: I try implement this idea in aee09a2b58bd9f84e2c95d0c92ba6c37ebdcb65f, what do you think?

src/test.js
raphamorim: Hmmm, good point. It doesn't work :/ 
Have any suggestion about it?

raphamorim:
Sorry about delay :octocat:

src/warning.js
```diff
@@ -0,0 +1,29 @@
...
+            "it's recommended to use it through its direct reference " +
+            "in the assertion context.",
+        stop: "DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.",
+        test: "DEPRECATED: Global 'test()' will be removed in " +
+            "QUnit 2.0, use 'QUnit.test() instead"
+    };
+
+    function log(warningMessage) {
+        if ( global.console && global.console.warn ) {
+            if (warningMessage) {
```
leobalter: you may blend this warningMessage to the if statement above. It cost less to check for it before at first too.

raphamorim: Updated: a6a8c1b43b5e5fc1bef0d4a1af258516fc1da667

raphamorim:
I agree with you that's not abstract. But in the case of the `test` method I would like to take this message to use in `new Test()`, to get this warning in `done()`

Like: 
```javascript
new Test({
   warning: warningMessage,
```

So the best approach is:
```javascript
var warning = QUnit.warning.deprecated.test;
QUnit.warning( "test" ); // or "qunit-test"
```

Or: In addition to print, return the message
```javascript
var warning = QUnit.warning( "test" );
````

File: src/core.js [hunk 5]
```diff
@@ -147,6 +147,13 @@ extend( QUnit, {
 	// DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.
 	stop: function( count ) {
 
+		var warning;
+
+		if ( this === global ) {
+			warning = QUnit.warning.deprecated.stop;
+			QUnit.warning(warning);
```
leobalter: We may avoid the new `warning` variable calling the function directly with it, less use of memory.

Also: I still believe it's better to map the messages as string keys, the point is that we may want to trigger only once for each warning. If you find another way to solve this (the single trigger) you don't need to worry on the key map.

leobalter:
Ouch! some bugs found on legacy browsers. https://travis-ci.org/jquery/qunit/builds/106548000

The Warning API is getting great, we're close to move this forward.

raphamorim:
@leobalter can you check out the latest changes?
I will see this bug in IE7 and start writing tests for this PR :) :octocat:

leobalter:
I need two udpates: 

- a rebase with the master branch.
- please make each warning trigger only once. I can't even check the CI the way it is. :)

Extra (probably for an extra PR in the future): you may try using QUnit.stack to mention where the first of each kind of deprecation has happened.

raphamorim:
Updated @leobalter :octocat:

leobalter:
We are close. Not we need to disable the warnings on the regular runs.

Maybe a `QUnit.warning.off()` being called on the main html test files (at least on `test/index.html`). Then, exclusive tests to check every warning (running on an isolated files, maybe a legacy-warnings.html and legacy-warnings.js).

The other improvements might come after these changes, potentially on another PR.

src/test.js
```diff
@@ -204,6 +207,8 @@ Test.prototype = {
 			}
 		}
 
+		QUnit.warning( "testDone-duration" );
```
gibson042: Warnings shouldn't be generated when there's nothing the user can do about it. I'd like to see this appear only in a getter for `duration` where `Object.defineProperty` exists.

src/warning.js
```diff
@@ -0,0 +1,48 @@
...
+            "In QUnit 2.0, invoking it will ONLY affect the " +
+            "`QUnit.config.autostart` blocking behavior.",
+        stop: "DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.",
+        "testDone-duration": "DEPRECATED: duration property will be removed " +
+            "in 2.0.0, use runtime instead.",
+        "exported-globals": "DEPRECATED: QUnit 2.0 will not export its API " +
+            "methods to the window object."
+    };
+
+    var once = [];
```
gibson042: This would be friendlier as an object, in which the associated values could either be numeric (e.g., a count) or even more informative (e.g., an object identifying offending instances). @leobalter?

src/export.js
```diff
@@ -1,24 +1,31 @@
...
-		};
-	}
-
-	for ( i in assertions ) {
-		QUnit[ i ] = applyCurrent( assertions[ i ] );
-	}
-})();
+function globalWarning( key ) {
+	return function() {
+		QUnit.warning( "exported-globals" );
```
gibson042: I'd prefer distinct warnings for every method, and for calls on `QUnit` vs. the global object (`qunit-ok`, `global-ok`, `qunit-notOk`, `global-notOk`, etc.).

src/warning.js
leobalter: yes, it makes sense. 

Having object values on this array would help with this problem too.

Even if we don't have too much extra information to store for now, ordinary objects would allow it to be more extensive.

I wish we could already use Maps and Sets here.

src/export.js
leobalter: :+1: this also helps identifying whatever is happening to trigger the warning.

raphamorim:
@leobalter about the tests for QUnit.warning, do you think it's better to create a separate file (`warning.html` for example) to test individual cases?

leobalter:
yes. I wanna make sure we trigger warnings only on isolated cases.

This will help us to update the main tests before the next major release.

test/amd.html
```diff
@@ -18,6 +18,7 @@
 	});
 
 	require( [ "qunit", "./amd" ], function( QUnit, tests ) {
+		QUnit.warning.off();
```
leobalter: these specific tests should not worry about triggering it off. That should occur only for the main test or anything else that are still covering deprecated API.

raphamorim:
@leobalter see the lastest change. Chek if this tests make sense to you? :octocat:

src/warning.js
```diff
@@ -0,0 +1,56 @@
...
+            "In QUnit 2.0, invoking it will ONLY affect the " +
+            "`QUnit.config.autostart` blocking behavior.",
+        stop: "DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.",
+        "testDone-duration": "DEPRECATED: duration property will be removed " +
+            "in 2.0.0, use runtime instead.",
+        "exported-globals": "DEPRECATED: QUnit 2.0 will not export its API " +
+            "methods to the window object."
+    };
+
+    var once = [],
```
gibson042: Copied from [now-outdated comments](https://github.com/jquery/qunit/pull/918#discussion_r53466092):
> This would be friendlier as an object, in which the associated values could ~~either~~ be ~~numeric (e.g., a count) or even more informative (e.g.,~~ an object identifying offending instances).

raphamorim:
@leobalter can you check the lastest changes?

leobalter:
It is LGTM, but I would like to ask for a final minor adjust. Can we hide the unnecessary warnings output from the grunt run?

test/moduleId.html
```diff
@@ -5,6 +5,9 @@
 		<title>QUnit ModuleId Test Suite</title>
 		<link rel="stylesheet" href="../dist/qunit.css">
 		<script src="../dist/qunit.js"></script>
+		<script>
+			QUnit.warning.off();
+		</script>
```
leobalter: this should not happen here and other minor and specifc tests files, unless they intentionally test a deprecated feature.

raphamorim: Following Warnings: 
```
DEPRECATED: duration property will be removed in 2.0.0, use runtime instead.
DEPRECATED: Use multiple tests instead of resetting inside a test. Use testStart or testDone for custom cleanup. This method will throw an error in 2.0, and will be removed in 2.1.```
```

Doesn't make sense throw this warnings?

leobalter: I need to investigate why they're being triggered here. I'll give some feedback after this.

src/warning.js
```diff
@@ -0,0 +1,101 @@
...
+                "in 2.0.0, use runtime instead."
+        },
+        {
+            type: "exported-globals",
+            message: "DEPRECATED: QUnit 2.0 will not export its API " +
+                "methods to the window object."
+        }
+    ];
+
+    var once = [],
```
gibson042: To repost from the https://github.com/jquery/qunit/pull/918#discussion_r53466092 thread, this should be an object in which the keys correspond to message identifiers  and the values are plain objects with data to be determined.

src/export.js
```diff
@@ -16,6 +16,13 @@
 	}
 }() );
 
+function globalWarning( key ) {
+	return function() {
+		QUnit.warning( "exported-globals" );
```
gibson042: https://github.com/jquery/qunit/pull/918#discussion_r53466479
> I'd prefer distinct warnings for every method, and for calls on `QUnit` vs. the global object (`qunit-ok`, `global-ok`, `qunit-notOk`, `global-notOk`, etc.).

test/moduleId.html
gibson042: The first is because of https://github.com/jquery/qunit/pull/918/files#r53465639 , which should be implemented via getters (and only in environments where getters are available), and the second is because we automatically call `QUnit.reset` in the `testDone` handler, which I think we can factor out so the warning is only generated inside of a test.

raphamorim:
@leobalter do you want that I remove ```QUnit.Warning.off()``` from the tests?

Gruntfile.js
```diff
@@ -122,7 +123,9 @@ grunt.initConfig( {
 			"test/random.html",
 			"test/regex-filter.html",
 			"test/regex-exclude-filter.html",
-			"test/string-filter.html"
+			"test/string-filter.html",
+			"test/warning.html",
+			"test/warning-once.html"
```
leobalter: are are these files? `test/warning.html` and `test/warning-once.html` does not exist.

build/tasks/test-on-node.js
```diff
@@ -49,6 +49,8 @@ module.exports = function( grunt ) {
 
 		QUnit.config.autorun = false;
 
+		QUnit.warning.off();
+
```
leobalter: this should be removed

reporter/html.js
```diff
@@ -8,6 +8,8 @@ if ( typeof window === "undefined" || !window.document ) {
 QUnit.init = function() {
 	var config = QUnit.config;
 
+	QUnit.warning( "init" );
+
```
leobalter: this should be removed

src/core.js
```diff
@@ -145,7 +145,6 @@ extend( QUnit, {
 
 	// DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.
 	stop: function( count ) {
-
```
leobalter: this should be removed

src/core/logging.js
```diff
@@ -55,6 +55,8 @@ function verifyLoggingCallbacks() {
 	for ( loggingCallback in loggingCallbacks ) {
 		if ( QUnit[ loggingCallback ] !== loggingCallbacks[ loggingCallback ] ) {
 
+			QUnit.warning( "loggingCallbacks" );
+
```
leobalter: this should be removed

src/export.js
```diff
@@ -16,6 +16,12 @@
 	}
 }() );
 
+function globalWarning( key ) {
+	return function() {
+		return QUnit[ key ].apply( this, arguments );
+	};
+}
+
```
leobalter: this should be removed

src/export.js
```diff
@@ -43,7 +49,7 @@ if ( defined.document ) {
 			];
 
 		for ( i = 0, l = keys.length; i < l; i++ ) {
-			window[ keys[ i ] ] = QUnit[ keys[ i ] ];
+			window[ keys[ i ] ] = globalWarning( keys[ i ] );
```
leobalter: this should be removed

src/test.js
```diff
@@ -37,6 +37,7 @@ function Test( settings ) {
 	} else {
 		this.assert = new Assert( this );
 	}
+
```
leobalter: this should be removed

src/test.js
```diff
@@ -444,7 +447,6 @@ Use testStart or testDone for custom cleanup.
 This method will throw an error in 2.0, and will be removed in 2.1
 */
 QUnit.reset = function() {
-
```
leobalter: this should be removed

test/autostart.html
```diff
@@ -12,7 +12,9 @@
 	var beginData,
 		times = {};
 
-	( function() {
+	QUnit.warning.off();
+
+	(function() {
```
leobalter: this should be removed

test/headless.html
```diff
@@ -8,7 +8,9 @@
 	<script src="main/test.js"></script>
 	<script src="main/deepEqual.js"></script>
 	<script>
-		( function () {
+		(function () {
+			QUnit.warning.off();
+
```
leobalter: this should be removed

test/main/globals.js
```diff
@@ -3,27 +3,13 @@
 
 QUnit.module( "globals" );
 
-function checkExported( assert, methods, isAssertion ) {
+function checkExported( assert, methods ) {
```
leobalter: this should be removed

test/main/globals.js
```diff
@@ -3,27 +3,13 @@
...
-			"QUnit exports QUnit." + method + " to the global scope"
-		);
-
-		if ( isAssertion ) {
-			assert.strictEqual(
-				window[ method ],
-				assert[ method ],
-				"Global " + method + " is the same of assert." + method
-			);
-		}
```
leobalter: this should be removed

test/main/globals.js
```diff
@@ -33,8 +19,7 @@ QUnit.test( "QUnit exported methods", function( assert ) {
 			"start", "stop"
 		];
 
-	// 2 assertions per item on checkExported
-	assert.expect( globals.length * 2 );
+	assert.expect( globals.length );
```
leobalter: this should be removed

test/reporter-html/legacy-markup.html
```diff
@@ -5,6 +5,9 @@
 	<title>QUnit HTML Reporter - Legacy Markup</title>
 	<link rel="stylesheet" href="../../dist/qunit.css">
 	<script src="../../dist/qunit.js"></script>
+	<script type="text/javascript">
+		QUnit.warning.off();
+	</script>
```
leobalter: this should be removed

test/reporter-html/no-qunit-element.html
```diff
@@ -5,6 +5,9 @@
 	<title>QUnit HTML Reporter - No Markup</title>
 	<link rel="stylesheet" href="../dist/qunit.css">
 	<script src="../../dist/qunit.js"></script>
+	<script type="text/javascript">
+		QUnit.warning.off();
+	</script>
```
leobalter: this should be removed

test/reporter-html/single-testid.html
```diff
@@ -5,6 +5,9 @@
 	<title>QUnit Main Test Suite</title>
 	<link rel="stylesheet" href="../../dist/qunit.css">
 	<script src="../../dist/qunit.js"></script>
+	<script type="text/javascript">
+		QUnit.warning.off();
+	</script>
```
leobalter: this should be removed

raphamorim:
@leobalter check last changes :)

leobalter:
there's still some pending comments above, like: https://github.com/jquery/qunit/pull/918#discussion_r60269519

test/main/globals.js
raphamorim: Updated: 82ddcb3

test/main/globals.js
raphamorim: Updated: 82ddcb3

test/autostart.html
```diff
@@ -12,7 +12,7 @@
 	var beginData,
 		times = {};
 
-	( function() {
+	(function() {
```
trentmwillis: Undo? Seems like an accidental change.

test/headless.html
```diff
@@ -8,7 +8,8 @@
 	<script src="main/test.js"></script>
 	<script src="main/deepEqual.js"></script>
 	<script>
-		( function () {
+		(function () {
+
```
trentmwillis: Same here.

trentmwillis:
Overall this seems okay to me, just a few minor comments

test/autostart.html
raphamorim: Updated: 21cd239

test/headless.html
raphamorim: Updated: 21cd239

trentmwillis:
Sorry for letting this sit for so long. @leobalter is this still desirable/needed? I wasn't super involved here, so I'm unsure what the entire goal of the API was. I'd be fine to just use a simple warning as well for things like https://github.com/qunitjs/qunit/issues/986.

leobalter:
Makes more sense for now. Looking at this seems the warnings API is overkill at this point.

@raphamorim I'm closing this for now, thanks for the hard work and I'm sorry to not help with it so much.

raphamorim:
No problemo :)

Issue #893, opened 2015-11-12
Introduce beforeAll/afterAll module hooks?


trentmwillis:
I'd like to propose adding in `beforeAll`/`afterAll` hooks to test modules. Similar to the functions of the same name in Jasmine or `before`/`after` in Mocha. It would be convenient to have a way to access the testing context at the start and end of a module on a case-by-case basis instead of just the sweeping `moduleDone` and `moduleStart` callbacks.

I apologize if this has been discussed before, couldn't seem to find any background on the matter.

cc @nathanhammond

gibson042:
If I correctly understand what you want, this is [already possible](http://api.qunitjs.com/QUnit.module/): http://jsfiddle.net/8zfLqk0r/
```js
QUnit.module( "module", function( obj ) {
	obj.beforeEach( function() {
		this.setup = true;
	} );
	obj.afterEach( function() {
		console.log( "teardown" );
	} );
	QUnit.test( "test", function( assert ) {
        assert.strictEqual( this.setup, true,
		"beforeEach ran" );
    } );
} );
```

trentmwillis:
@gibson042 not quite, guess I should've explain myself a bit more clearly.

What I'm looking for are hooks to run at the beginning of a module and at the end of the module, rather than at the beginning and end of each test. [Similar to what is done in Jasmine.](http://jasmine.github.io/2.3/introduction.html#section-11)

cjmaynar:
This would be pretty useful indeed. I'm looking for something like Python UnitTest's setUpClass/tearDownClass methods that will run once before any tests are run, and then once after all the tests are complete

leobalter:
@trentmwillis that would be an interesting feature. If it was discussed before - certainly mentioned at least - I believe we should try it.

I would name them before/after module hooks, removing the `All`, it's easier to match beforeEach and afterEach.

PRs are welcome!

JamesMGreene:
It is definitely a feature that we've seen requests for before. I myself would like that capability as well.

JamesMGreene:
P.S. We didn't implement it in the past because there was already a decent plugin available for it: [qunit-once](https://github.com/bahmutov/qunit-once).

However, I would imagine that [old] plugin does not currently offer all the niceties that a built-in set of hooks easily could, i.e. guarantee `before` callbacks to run _before_ any `beforeEach` callbacks, guarantee `after` callbacks to run _after_ any `afterEach` callbacks, first-class support for async, first-class support for promises, etc

[...]

PR #919, opened 2016-01-12
Core: Introduce before/after hooks for modules

trentmwillis:
As discussed in #893. Included tests for the following in support of this feature:

- [x] `before` runs before `beforeEach`
- [x] `after` runs after `afterEach`
- [x] `after` only runs after last test in module
- [x] `before` only runs before first test in module
- [x] first test supports `before` with a promise
- [x] last test supports `after` with a promise
- [x] first test supports `before` with an assert.async
- [x] last test supports `after` with an assert.async
- [x] support nested module `before` hooks
- [x] support nested module `after` hooks
- [x] context changes made in `before` are applied to all tests

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -32,8 +32,10 @@ extend( QUnit, {
 		module = createModule();
 
 		moduleFns = {
+			before: setHook( module, "before" ),
 			beforeEach: setHook( module, "beforeEach" ),
-			afterEach: setHook( module, "afterEach" )
+			afterEach: setHook( module, "afterEach" ),
+			after: setHook( module, "after" )
 		};
 
 		if ( executeNow instanceof Function ) {
@@ -55,11 +57,13 @@ extend( QUnit, {
 				name: moduleName,
 				parentModule: parentModule,
 				tests: [],
-				moduleId: generateHash( moduleName )
+				moduleId: generateHash( moduleName ),
+				testsRun: 0
 			};
 
 			var env = {};
 			if ( parentModule ) {
+				parentModule.childModule = module;
 				extend( env, parentModule.testEnvironment );
 				delete env.beforeEach;
 				delete env.afterEach;
--- a/src/test.js
+++ b/src/test.js
@@ -75,8 +75,10 @@ Test.prototype = {
 		config.current = this;
 
 		if ( this.module.testEnvironment ) {
+			delete this.module.testEnvironment.before;
 			delete this.module.testEnvironment.beforeEach;
 			delete this.module.testEnvironment.afterEach;
+			delete this.module.testEnvironment.after;
 		}
 		this.testEnvironment = extend( {}, this.module.testEnvironment );
 
@@ -133,10 +135,22 @@ Test.prototype = {
 		checkPollution();
 	},
 
-	queueHook: function( hook, hookName ) {
+	queueHook: function( hook, hookName, hookOwner ) {
 		var promise,
 			test = this;
 		return function runHook() {
+			if ( hookName === "before" ) {
+				if ( hookOwner.testsRun !== 0 ) {
+					return;
+				}
+
+				test.preserveEnvironment = true;
+			}
+
+			if ( hookName === "after" && hookOwner.testsRun !== numberOfTests( hookOwner ) - 1 ) {
+				return;
+			}
+
 			config.current = test;
 			if ( config.notrycatch ) {
 				callHook();
@@ -166,7 +180,7 @@ Test.prototype = {
 			}
 			if ( module.testEnvironment &&
 				QUnit.objectType( module.testEnvironment[ handler ] ) === "function" ) {
-				hooks.push( test.queueHook( module.testEnvironment[ handler ], handler ) );
+				hooks.push( test.queueHook( module.testEnvironment[ handler ], handler, module ) );
 			}
 		}
 
@@ -205,6 +219,7 @@ Test.prototype = {
 			}
 		}
 
+		notifyTestsRan( this.module );
 		runLoggingCallbacks( "testDone", {
 			name: this.testName,
 			module: this.module.name,
@@ -233,6 +248,13 @@ Test.prototype = {
 		config.current = undefined;
 	},
 
+	preserveTestEnvironment: function() {
+		if ( this.preserveEnvironment ) {
+			this.module.testEnvironment = this.testEnvironment;
+			this.testEnvironment = extend( {}, this.module.testEnvironment );
+		}
+	},
+
 	queue: function() {
 		var priority,
 			test = this;
@@ -249,16 +271,25 @@ Test.prototype = {
 					test.before();
 				},
 
+				test.hooks( "before" ),
+
+				function() {
+					test.preserveTestEnvironment();
+				},
+
 				test.hooks( "beforeEach" ),
+
 				function() {
 					test.run();
 				},
 
 				test.hooks( "afterEach" ).reverse(),
+				test.hooks( "after" ).reverse(),
 
 				function() {
 					test.after();
 				},
+
 				function() {
 					test.finish();
 				}
@@ -644,4 +675,19 @@ function only( testName, expected, callback, async ) {
 	} );
 
 	newTest.queue();
+}
+
+function numberOfTests( module ) {
+	var count = module.tests.length;
+	while ( module = module.childModule ) {
+		count += module.tests.length;
+	}
+	return count;
+}
+
+function notifyTestsRan( module ) {
+	module.testsRun++;
+	while ( module = module.parentModule ) {
+		module.testsRun++;
+	}
 }
--- a/test/main/async.js
+++ b/test/main/async.js
@@ -233,6 +233,33 @@ QUnit.test( "fails if callback is called more than callback call count", functio
 } );
 
 QUnit.module( "assert.async fails if callback is called more than once in", {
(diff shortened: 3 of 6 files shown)


trentmwillis:
@leobalter I think this is ready for review. Tried my best to keep the implementation within existing constructs, but required a few new properties to get all the features/behavior this should support.

src/core.js
```diff
@@ -56,11 +58,14 @@ extend( QUnit, {
 			var module = {
 				name: moduleName,
 				parentModule: parentModule,
-				tests: []
+				tests: [],
+				testsRun: 0
```
leobalter: this might be my only concern on this PR. Exposing a new detail on the API conflicts with the plans for the EventEmitter changes (#882).

Although, the `tests` array [will still prevail](https://github.com/js-reporters/js-reporters/issues/30) and I wonder if we can set an equivalent property on the [reference object for the test on that array](https://github.com/jquery/qunit/pull/919/files#diff-f6c2ded91c9881cdc31b5bfcabb54c3bR25). Doing that, we would have the tradeoff to filter the array, but it would save us a new API exposure.

test/main/async.js
```diff
@@ -233,6 +233,35 @@ QUnit.test( "fails if callback is called more than callback call count", functio
...
+
+		assert.throws(function() {
+			overDone();
+		}, new RegExp( "Too many calls to the `assert.async` callback" ) );
+
+		done();
+	}
+});
+
+QUnit.test( "before", function( /* assert */ ) {
```
leobalter: we may omit the parameter comment.

leobalter:
The tests look great and I like the implementation.

My concerns remained only on the API as in https://github.com/jquery/qunit/pull/919#discussion_r50030019

Other noticed minor style issues for missing spaces between parenthesis, which are not blockers.

src/core.js
trentmwillis: Looking at the ["events details types" issue](https://github.com/js-reporters/js-reporters/issues/30), I think we would be able to replace `module.testsRun` with `module.status.total` when it is implemented since it looks like it will be similar.

I would also be fine with setting an equivalent property on the `tests` array, though it feels a little less correct.

leobalter: > Looking at the "events details types" issue, I think we would be able to replace module.testsRun with module.status.total when it is implemented since it looks like it will be similar.

:+1: that would be great!

trentmwillis: In the interim, should we proceed with this current implementation? (After I clean up the style issues, of course.)

leobalter: Yes, sure.

We will need to have the API docs on http://github.com/jquery/api.qunitjs.com to be able to land it, would you mind to help on that part too?

trentmwillis: I'll open a PR documenting the new `before`/`after` hooks for `module`. Is there anything else I should document?

trentmwillis:
Updated the PR (think I got all the style issues).

src/core.js
leobalter: I believe they'll be enough for now, QUnit docs will probably face a whole logging details review on a near future.

src/test.js
```diff
@@ -132,10 +134,22 @@ Test.prototype = {
...
 		return function runHook() {
+			if ( hookName === "before" ) {
+				if ( hookOwner.testsRun !== 0 ) {
+					return;
+				}
+
+				test.preserveEnvironment = true;
+			}
+
+			if ( hookName === "after" && hookOwner.testsRun !== numberOfTests(hookOwner) - 1 ) {
```
leobalter: a small style issue here `numberOfTests(hookOwner)` => `numberOfTests( hookOwner )

trentmwillis: Ah, missed one.

src/test.js
```diff
@@ -74,8 +74,10 @@ Test.prototype = {
 		config.current = this;
 
 		if ( this.module.testEnvironment ) {
+			delete this.module.testEnvironment.before;
 			delete this.module.testEnvironment.beforeEach;
 			delete this.module.testEnvironment.afterEach;
+			delete this.module.testEnvironment.after;
```
gibson042: This may require treating the change as breaking. @leobalter?

gibson042:
I wonder if this should cover edge cases like [delaying the definition of a test](https://jsfiddle.net/obuj0czg/). They're a bit pathological, but it would still be bad to run an `after` hook more than once.

trentmwillis:
@gibson042 is that a real use case? If so, I don't really see a realistic way to handle it other than making an explicit declaration of the number of tests your module should have, since we need advanced knowledge of whether or not another test is expected to run.

src/test.js
leobalter: I don't think so. We have this delete statement only because we're introducing new properties to the modules' hook: `before` and `after`. Anyone previously extending that module with these keywords would get their tests affected. 

That same could happen when we moved `setup` and `teardown` to `beforeEach` and `afterEach`. Similarly we could mention the introduction of `QUnit.skip` to the API that could  interfere with any other use of custom made methods.

leobalter:
> @trentmwillis: is that a real use case?

That example reflects tests added through async files loading, as in AMD. A dirty check could use a global variable to tell they are added.

Anyway, I also believe it's a developer responsibility to organize well tests introduced asynchronously. An update to the example - where I only added a single line: reveals a structure antipattern: https://jsfiddle.net/obuj0czg/1/

trentmwillis:
Okay, makes sense. Should I attempt to cover that use case? Based on the example you gave, it probably doesn't make much sense without guaranteeing what module a test belongs to

leobalter:
> Should I attempt to cover that use case?

I believe a simple timeout example is enough to cover it up, testing it works properly on async operations and it does not leak anywhere else. AMD cases will be the responsibility for those who implements it.

src/test.js
gibson042: > Anyone previously extending that module with these keywords would get their tests affected.

Exactly. Code like `QUnit.module( name, { after: new Date(2016, 0, 1) }, … )`, _which is currently supported_, will be broken. This is why I abhor the conflation of module options and test context in a single parameter, but there it is. And it's different from adding `QUnit.skip` because `QUnit` is our object, but that parameter is explicitly wide open for users, minus our exceptions.

Anyway, while maybe not a huge deal, this is technically breaking.

gibson042:
> Anyway, I also believe it's a developer responsibility to organize well tests introduced asynchronously. An update to the example - where I only added a single line: reveals a structure antipattern: https://jsfiddle.net/obuj0czg/1/

Right, pathological. :smile:

gibson042:
Sorry about the noise. But my point was that we should add documentation and assertions that lock down how `after` behaves in such cases (either "invoked exactly once, and therefore ignored delayed tests" or "invoked whenever the module's test queue is empty, and therefore after delayed tests").

trentmwillis:
> "invoked whenever the module's test queue is empty, and therefore after delayed tests"

Sorry if I'm missing something obvious, but going back to my earlier point, if a test definition is delayed, then it may not get added to the module's test queue until after the module has already emptied it's test queue. How do we notify the module that even though it has ran all of it's tests, it should not run the `after` hook until more tests are added and ran?

gibson042:
> How do we notify the module that even though it has ran all of it's tests, it should not run the `after` hook until more tests are added and ran?

We don't. We just document and assert our behavior, which will take one of two forms:
* Invoke `after` exactly once. Tests added after its invocation will have to manage their own cleanup.
* Invoke `after` whenever the module's test queue becomes empty. Tests added after its first invocation will result in successive invocations, and users employing such patterns must structure it to accommodate.

I'm just forcing the above as an explicit choice, rather than an accident of implementation.

trentmwillis:
> I'm just forcing the above as an explicit choice, rather than an accident of implementation.

Okay, cleared that all up. Currently the implementation means your second option is true (it will result in successive invocations), though I think the first option makes more sense. This is because I imagine `after` callbacks will often pair with `before` callbacks, which are only run once.

If that makes sense, I'll add a test for that here and document it in the documentation PR.

leobalter:
> If that makes sense,  I'll add a test for that here and document it in the documentation PR.

@trentmwillis ++. 

do you mind? It would be great to have it tested and documented.

trentmwillis:
@leobalter updated with a test and updated the doc PR as well. Travis failed but it looks like it's for the same reason the master build is failing currently.

leobalter:
I'm sick of browserstack-runner timeouts.

src/test.js
leobalter: > This is why I abhor the conflation of module options and test context in a single parameter

We should discuss a way to solve this before 2.0. Maybe we need to isolate the hooks from the environment or event remove the possibility to define env values on this object (let `this` do its job).

trentmwillis:
@leobalter @gibson042 is the only remaining question whether or not we should consider this breaking? If so, I'd be fine with feature-flagging this somehow or delaying merge until 2.0.

leobalter:
@trentmwillis, I really want to merge and release a new version with it, but I might need to release a major version.

I'm running on other necessary changes (the warning api) to release the next major, so releasing this feature won't be a problem.

trentmwillis:
@leobalter a major version sounds good. Will let this sit until ready then.

stefanpenner:
@trentmwillis good work!

trentmwillis:
@leobalter this has been rebased and should be good to go. Care to do a once over since it's been awhile?

PR #920, opened 2016-01-21
Assert: Change push to pushResult

### LINKED ISSUE #827: Replace push(result, actual, expected, message, negative) with object signature
As discussed in #822, the now even longer signature for `.push()` is bad. Especially with ES6 an object signature would work very well, `.push({ result, actual, expected, message, negative })`.

As I also mentioned in that PR: If we deprecate the current signature anyway, we might as well come up with a better name, since `push` is very generic. Any ideas for a better name?

mixed:
I have fixed #827.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -48,7 +48,19 @@ QUnit.assert = Assert.prototype = {
 		};
 	},
 
-	// Exports test.pushResult() to the user API
+	// Exports test.push() to the user API
+	// Alias of pushResult.
+	push: function( result, actual, expected, message, negative ) {
+		var currentAssert = this instanceof Assert ? this : QUnit.config.current.assert;
+		return currentAssert.pushResult( {
+			result: result,
+			actual: actual,
+			expected: expected,
+			message: message,
+			negative: negative
+		} );
+	},
+
 	pushResult: function( resultInfo ) {
 
 		// resultInfo = { result, actual, expected, message, negative }
@@ -74,6 +86,7 @@ QUnit.assert = Assert.prototype = {
 		if ( !( assert instanceof Assert ) ) {
 			assert = currentTest.assert;
 		}
+
 		return assert.test.pushResult( resultInfo );
 	},
 
--- a/src/assert.js
+++ b/src/assert.js
@@ -48,8 +48,10 @@ QUnit.assert = Assert.prototype = {
 		};
 	},
 
-	// Exports test.push() to the user API
-	push: function( /* result, actual, expected, message, negative */ ) {
+	// Exports test.pushResult() to the user API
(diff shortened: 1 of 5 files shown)


src/assert.js
```diff
@@ -48,8 +48,10 @@ QUnit.assert = Assert.prototype = {
 		};
 	},
 
-	// Exports test.push() to the user API
-	push: function( /* result, actual, expected, message, negative */ ) {
+	// Exports test.pushResult() to the user API
+	pushResult: function( resultInfo ) {
```
leobalter: The implementation looks great, but we need to keep a deprecated `.push` method until 2.0 is released.

the deprecated `.push()` can be a simple abstraction to call `.pushResult` with the given arguments.

mixed: @leobalter 
Thank you. I have applied your review.

Arkni:
Why changing the test as well ?
I think the test should remain untouched until `push` is removed in `2.0` (If I'm not mistaken).
But if it's ok to change them, the you should also modify this line in `test/main/test.js` as well: https://github.com/jquery/qunit/blob/master/test/main/test.js#L33

mixed:
@Arkni 
Yes. I think that you right.

@Arkni @leobalter 
I think that It can change `push` to `pushResult` in `test/main/assert.js`. Becacuse It doesn't seem to be common usage.  How about it? If It can't change `push` to `pushResult`  in `test/main/assert.js`. I have to revert to `push` in assertion(`ok`, `notOk`...)

leobalter:
d1989ce7a4a115181405c47cf23c81d07066abbd is not necessary, it's testing dump.js, and it's ok to use pushResult there.

The change on https://github.com/jquery/qunit/blob/master/test/main/test.js#L33 suggested by @Arkni is interesting too.

What we could have is a test to call pushResult and check the results on the test log.

mixed:
@leobalter 
Thank you.

leobalter:
@mixed, can we improve this PR with @Arkni suggestion? Applying the change on https://github.com/jquery/qunit/blob/master/test/main/test.js#L33 to use pushResult

Other: I believe we could have a test to mockup `.pushResult`, and call `assert.push` to check it calls `pushResult` with parameters.

After that we are good to use merge. (I'll add the docs)

mixed:
@leobalter 
Thank you for your review. I have updated it.
- Apply pushResult on https://github.com/jquery/qunit/blob/master/test/main/test.js#L33
- Add testcase for pushResult using `log`.

leobalter:
LGTM, I'll merge this soon.

leobalter:
done! merged at ecf0b36

PR #922, opened 2016-01-29
Support deprecated assert methods in environments other than browsers.

lehni:
The way QUnit currently exposes deprecated assert methods on the global QUnit object only works in browsers, because the code is wrapped by a check for `defined.document`.

I am using QUnit for hybrid tests inside Node.js as well as in a browser environment through PhantomJS (as part of my new gulp plugin for such hybrid test scenarios, see https://github.com/lehni/gulp-qunits), and I expect it to behave the same way in both environments.

At the moment, the Node QUnit Runner needs to use such trickery to make this work, it would be nice to be able to get rid of it:

```js
// Abuse a test to fetch the assert object along with its prototype and methods.
QUnit.test('setup', function(assert) {
    var Assert = assert.constructor,
        asserts = Object.getPrototypeOf(assert);
    Object.keys(asserts).forEach(function(key) {
        var current = asserts[key];
        global[key] = QUnit[key] = function() {
            current.apply(new Assert(QUnit.config.current), arguments);
        };
    });
    assert.expect(0);
});
```

### DIFF
--- a/src/export.js
+++ b/src/export.js
@@ -1,23 +1,23 @@
+// Deprecated
+// Extend assert methods to QUnit for Backwards compatibility
+(function() {
+	var i,
+		assertions = Assert.prototype;
+
+	function applyCurrent( current ) {
+		return function() {
+			var assert = new Assert( QUnit.config.current );
+			current.apply( assert, arguments );
+		};
+	}
+
+	for ( i in assertions ) {
+		QUnit[ i ] = applyCurrent( assertions[ i ] );
+	}
+})();
+
 // For browser, export only select globals
 if ( defined.document ) {
-
-	// Deprecated
-	// Extend assert methods to QUnit and Global scope through Backwards compatibility
-	(function() {
-		var i,
-			assertions = Assert.prototype;
-
-		function applyCurrent( current ) {
-			return function() {
-				var assert = new Assert( QUnit.config.current );
-				current.apply( assert, arguments );
-			};
-		}
-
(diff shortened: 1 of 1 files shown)


leobalter:
I'm not convinced yet to expose QUnit assertions to the global scope on node.

The implementation problem can be solved using the assertions from the assert argument on each test. With a little bit of ES6's destructuring (if available) it's even better.

A new major version is on the way and all assertions won't be exposed to the global even on browsers.

lehni:
It's not exposed to the global scope on Node, it's just added to the QUnit object. The exposing to the global scope only happens in browsers, even with this PR. But due to the way the code currently is written, the QUnit object does not give access to the same methods between these two environments.

I think in any given version, the same things should be exposed on the QUnit object, as otherwise I can't guarantee that the same tests run in such a hybrid environment as offered by my gulp plugin. Of course people should convert to the assert object, but until then it would be nice if things didn't behave differently based on wether QUnit is running in a window or in Node.js

leobalter:
cc @gibson042

gibson042:
I'm convinced... this PR still avoids writing properties onto the global object in Node.js, but there's no reason why `QUnit.ok` shouldn't exist there as it does in browsers.

lehni:
:+1:

leobalter:
Good enough, I'll merge this.

lehni:
Great, thanks!

leobalter:
merged at e010381

PR #924, opened 2016-02-01
Remove browserstack runs on legacy browsers

leobalter:
Removing IE 6, 7, and 8, Safari 5.1 and 6.

@mgol, I want to sync this up with jQuery Core while QUnit is still on 1.x.

### DIFF
--- a/build/browserstack-legacy-1.json
+++ b/build/browserstack-legacy-1.json
@@ -10,11 +10,6 @@
 		"chrome_previous",
 		"firefox_previous",
 		"opera_previous",
-		"safari_5_1",
-		{
-			"browser": "safari",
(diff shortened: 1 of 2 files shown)


mgol:
jQuery's `master` branch has already dropped all those browsers that you mentioned so this is only about jQuery 1.x. We should have `1.12.1` ready soon and, barring any critical issues, we don't currently plan more patch releases; this shouldn't block you, though, we don't have to update QUnit for this branch.

leobalter:
That's great to know, @mgol!

@gibson042 thumbs up?

leobalter:
LoDash supports IE9+ and Safari 7+, that was my last major concern.

Issue #926, opened 2016-02-02
Run tests in random order


trentmwillis:
In the [docs for the `reorder` config option](https://api.qunitjs.com/QUnit.config/), it notes that reordering failed tests can:

> also lead to random errors, in case your testsuite has non-atomic tests, where the order is important. You should fix those issues, instead of disabling reordering!

So, on a similar note, I'd like to propose adding the ability to specify that tests run in a seeded-random order. I think it would be relatively straightforward to introduce this as QUnit already has the infrastructure to reorder tests. This may not be particularly useful in browser, but could be very beneficial for CI/Post-Commit tooling looking to identify leaky/dependent tests.

As a point of reference, I looked into what other test frameworks have done in regards to this and found that this feature was [recently introduced in Jasmine](https://github.com/jasmine/jasmine/pull/927) and has been [requested multiple times for Mocha](https://github.com/mochajs/mocha/issues/902).

PR #931, opened 2016-02-17
Core: Add ability to run tests in seeded-random order

trentmwillis:
Addresses #926.

Didn't get any feedback on the issue, but since I knew it would be simple to implement, I figured I'd give a sample implementation and let that spur conversation.

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -79,6 +79,16 @@ if ( urlParams.filter === true ) {
 // String search anywhere in moduleName+testName
 config.filter = urlParams.filter;
 
+// Randomize the order in which tests are run
+if ( urlParams.seed ) {
+	config.seed = urlParams.seed;
+
+	// Generate a random seed if the seed option is specified without a value
+	if ( config.seed === true ) {
+		config.seed = Math.random().toString( 36 ).slice( 2 );
+	}
+}
+
 config.testId = [];
 if ( urlParams.testId ) {
 	// Ensure that urlParams.testId is an array
--- a/src/test.js
+++ b/src/test.js
@@ -1,5 +1,6 @@
 var focused = false;
 var priorityCount = 0;
+var unitSampler;
 
 function Test( settings ) {
 	var i, l;
@@ -268,7 +269,7 @@ Test.prototype = {
 		priority = QUnit.config.reorder && defined.sessionStorage &&
 				+sessionStorage.getItem( "qunit-test-" + this.module.name + "-" + this.testName );
 
-		return synchronize( run, priority );
+		return synchronize( run, priority, config.seed );
 	},
 
 	pushResult: function( resultInfo ) {
@@ -481,8 +482,9 @@ function generateHash( module, testName ) {
 	return hex.slice( -8 );
 }
 
-function synchronize( callback, priority ) {
-	var last = !priority;
+function synchronize( callback, priority, seed ) {
+	var last = !priority,
+		index;
 
 	if ( QUnit.objectType( callback ) === "array" ) {
 		while ( callback.length ) {
@@ -493,6 +495,14 @@ function synchronize( callback, priority ) {
 
 	if ( priority ) {
 		config.queue.splice( priorityCount++, 0, callback );
+	} else if ( seed ) {
+		if ( !unitSampler ) {
+			unitSampler = unitSamplerGenerator( seed );
+		}
+
+		// Insert into a random position after all priority items
+		index = Math.floor( unitSampler() * ( config.queue.length - priorityCount + 1 ) );
+		config.queue.splice( priorityCount + index, 0, callback );
 	} else {
 		config.queue.push( callback );
 	}
@@ -500,6 +510,25 @@ function synchronize( callback, priority ) {
 	if ( config.autorun && !config.blocking ) {
 		process( last );
 	}
+}
+
+function unitSamplerGenerator( seed ) {
+
+	// 32-bit xorshift, requires only a nonzero seed
+	// http://excamera.com/sphinx/article-xorshift.html
+	var sample = parseInt( generateHash( seed ), 16 ) || -1;
+	return function() {
+		sample ^= sample << 13;
+		sample ^= sample >>> 17;
+		sample ^= sample << 5;
+
+		// ECMAScript has no unsigned number type
+		if ( sample < 0 ) {
+			sample += 0x100000000;
+		}
+
+		return sample / 0x100000000;
+	};
 }
 
 function saveGlobal() {
--- a/test/random.html
+++ b/test/random.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>QUnit Randomization Test Suite</title>
+    <link rel="stylesheet" href="../dist/qunit.css">
+    <script src="../dist/qunit.js"></script>
+    <script>
+      QUnit.config.seed = "7x9";
+    </script>
+    <script src="random.js"></script>
+  </head>
+  <body>
+    <div id="qunit"></div>
+  </body>
+</html>
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -111,6 +111,7 @@ grunt.initConfig({
 			"test/reporter-html/no-qunit-element.html",
 			"test/reporter-html/single-testid.html",
 			"test/only.html",
+			"test/random.html",
 			"test/regex-filter.html",
 			"test/regex-exclude-filter.html",
 			"test/string-filter.html"
--- a/build/browserstack-current-1.json
+++ b/build/browserstack-current-1.json
@@ -4,7 +4,8 @@
 	"test_framework": "qunit",
 	"test_path": [
 		"test/index.html",
-		"test/logs.html"
+		"test/logs.html",
+		"test/random.html"
 	],
 	"browsers": [
 		"chrome_current",
--- a/build/browserstack-legacy-1.json
+++ b/build/browserstack-legacy-1.json
@@ -4,7 +4,8 @@
 	"test_framework": "qunit",
 	"test_path": [
 		"test/index.html",
-		"test/logs.html"
+		"test/logs.html",
+		"test/random.html"
 	],
 	"browsers": [
 		"chrome_previous",
--- a/test/random.js
+++ b/test/random.js
@@ -1,0 +1,23 @@
+var lastTest = "";
+
+QUnit.module( "QUnit.config.random" );
+
+QUnit.test( "1", function( assert ) {
+  assert.equal( lastTest, "2", "runs third" );
+  lastTest = "1";
+});
+
+QUnit.test( "2", function( assert ) {
+  assert.equal( lastTest, "3", "runs second" );
+  lastTest = "2";
+});
+
+QUnit.test( "3", function( assert ) {
+  assert.equal( lastTest, "", "runs first" );
+  lastTest = "3";
+});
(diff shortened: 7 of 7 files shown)


src/test.js
```diff
@@ -491,6 +492,13 @@ function synchronize( callback, priority ) {
 
 	if ( priority ) {
 		config.queue.splice( priorityCount++, 0, callback );
+	} else if ( random ) {
+		if ( !randomNumber ) {
+			randomNumber = randomNumberGenerator( random );
+		}
+
+		var index = Math.ceil( randomNumber() * ( config.queue.length - priorityCount ) );
```
leobalter: this var should be declared on the top of the function.

leobalter:
@trentmwillis this is very interesting, and I like it needs to be toggled.

Would you add the test file on the [browserstack runs](https://github.com/jquery/qunit/blob/master/build/browserstack-current-1.json)?

trentmwillis:
@leobalter updated and added the tests to the current and legacy browserstack runs. Also, for my own info, can you explain how the browserstack tests are broken up?

leobalter:
browserstack-runner uses to timeout more than I would like, the chances increases with more individual runs per call. Each browser and each html file increases a matrix line/column.

I broke the runs on: 

- 1 file with the top A+ browsers, which I expect to be the most used, with the most important test html files, containing core tests. 
- 1 file with the top A+ browsers, with the other html files.
- two other files with the same html files respectively, but running the tests on legacy browsers only.

On that way I was able to have 4 browserstack-runner calls with ~20 individual runs on each.

leobalter:
It looks good to me, I'll run some trolling tests locally (I want to see how it behaves) and merge if everything goes ok.

Meanwhile, would you mind adding an API doc?

trentmwillis:
Thanks for the info. Makes a lot of sense.

I'll open a PR for documentation as well.

src/test.js
```diff
@@ -500,6 +509,13 @@ function synchronize( callback, priority ) {
 	}
 }
 
+function randomNumberGenerator( seed ) {
```
gibson042: I'd like to see _this_ be the point where we consume user-specified data, rather than src/core/config.js. And I think it makes sense to reuse [`generateHash`](https://github.com/jquery/qunit/blob/ecf0b36fdde81db9deb122a2feecb75c5080a769/src/test.js#L462) (possibly with modifications) instead of special-casing numbers. For example:
```js
// 32-bit xorshift, requires only a nonzero seed
// http://excamera.com/sphinx/article-xorshift.html
seed = parseInt( generateHash( seed ), 16 ) || -1;
return function() {
	seed ^= seed << 13;
	seed ^= seed >>> 17;
	seed ^= seed << 5;

	// ECMAScript has no unsigned number type
	if ( seed < 0 ) {
		seed += 0x100000000;
	}

	return seed / 0x100000000;
};
```

src/test.js
```diff
@@ -491,6 +493,13 @@ function synchronize( callback, priority ) {
 
 	if ( priority ) {
 		config.queue.splice( priorityCount++, 0, callback );
+	} else if ( random ) {
+		if ( !randomNumber ) {
+			randomNumber = randomNumberGenerator( random );
```
gibson042: I'm not a fan of these misleading names. What about `seededUnitSample = unitSamplerGenerator( random );`?

gibson042:
I haven't had the chance to comment on #926, but I really like the idea. And this implementation is well on its way to landing.

src/test.js
trentmwillis: Would that mean the only operation done in `src/core/config.js` is `config.random = urlParams.random`? We then do the actual seed generation at this point?

src/test.js
trentmwillis: Hmm, I'll explore more options. Not sure how I feel about `unitSample` as it seems a bit vague (but maybe that's just me).

src/test.js
gibson042: Correct.

src/test.js
gibson042: "Unit" is intended to be understood in the [mathematical sense](https://en.wikipedia.org/wiki/Unit_interval), but I'm not married to it. However, an ideal name should capture both the "seeded" detail and indicate uniform distribution over the [0, 1) output range.

src/test.js
gibson042: But reading jquery/api.qunitjs.com/pull/121, I see why special-casing numbers is important. It would slightly vary the above example:
```js
if ( typeof seed !== "number" || seed % 1 !== 0 ) {
	seed = parseInt( generateHash( seed ), 16 );
}
seed = (seed % 0x100000000) || -1;
```

src/core/config.js
```diff
@@ -79,6 +79,16 @@ if ( urlParams.filter === true ) {
...
 
+// Randomize the order in which tests are run
+if ( urlParams.random ) {
+	config.random = parseInt( urlParams.random, 10 );
+
+	// Generate a random seed if the random option is specified with a non-number
+	if ( isNaN( config.random ) ) {
+		config.random = Math.floor( Math.random() * 10000 );
+	}
+}
```
gibson042: Taking the other comments into consideration, this would become something like:
```js
if ( urlParams.random ) {
	config.random = urlParams.random;
	if ( config.random === true ) {
		config.random = Math.random();
	}
}
```

So `?random` would yield a different seed each time, but e.g. `?random=foo` would always yield the same seed (and `?random=bar` would always yield its own unique seed).

trentmwillis:
@gibson042 updated with your suggestions slightly altered. It will now support taking any arbitrary string as a seed and in the event that it doesn't have a seed, it will generate a random string to use as the hash. I felt this was more ergonomic than just using `Math.random()` in case you want to reproduce a run.

src/core/config.js
```diff
@@ -79,6 +79,16 @@ if ( urlParams.filter === true ) {
 // String search anywhere in moduleName+testName
 config.filter = urlParams.filter;
 
+// Randomize the order in which tests are run
+if ( urlParams.random ) {
+	config.random = urlParams.random;
+
+	// Generate a random seed if the random option is specified without a seed
+	if ( config.random === true ) {
+		config.random = Math.random().toString( 36 ).substr( 2 );
```
gibson042: Nice! Note that `.slice( 2 )` would be more in line with our conventions.

src/test.js
```diff
@@ -491,6 +493,13 @@ function synchronize( callback, priority ) {
 
 	if ( priority ) {
 		config.queue.splice( priorityCount++, 0, callback );
+	} else if ( random ) {
+		if ( !unitSampler ) {
+			unitSampler = unitSamplerGenerator( random );
+		}
+
+		index = Math.ceil( unitSampler() * ( config.queue.length - priorityCount ) );
```
gibson042: This expression doesn't look quite right. I think it should be `index = Math.floor( unitSampler() * (config.queue.length - priorityCount + 1) )`, and definitely merits a comment like "Insert into a random queue position after all priority items".

test/random.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>QUnit Randomization Test Suite</title>
+    <link rel="stylesheet" href="../dist/qunit.css">
+    <script src="../dist/qunit.js"></script>
+    <script>
+      QUnit.config.random = '7x9';
```
gibson042: Double quotes, please.

Krinkle:
> I felt this was more ergonomic than just using Math.random() in case you want to reproduce a run.

That's an interesting idea. I hadn't previously considered that. However for that to work we'll need to make it easy for users to actually do reproduce it with the same hash. Perhaps we should expose it via the data report and in the html reporter. Could also use History API to backfill the query parameter?

trentmwillis:
> Perhaps we should expose it via the data report and print it in the html reporter by default. Could also perhaps use History API to backfill the query parameter?

Both thoughts that I've had. However, I am unsure of how much value this provides as I don't foresee using this feature in browser frequently (mostly CI usage), so you most likely won't see any URL updates or the HTML output. The query param backfilling in particular might be problematic if you were to want to run tests repeatedly in new orders by just reloading the page when the suite has finished.

Currently, the seed will be kept in `QUnit.config.random` (potentially renaming to `QUnit.config.seed`) so it'd be possible for reporters to simply log that at the beginning or end of the run.

Krinkle:
@trentmwillis: Hm.. okay. So if we don't foresee usage, why use the seed? Maybe we wanna go ahead with a popping off items from the tests queue with plain `Math.random()`?

Or is it meant more as a way of debugging manually? One will have to find a seed manually in that case (since the CI failure wouldn't give it), but I can see how that might be useful yeah.

trentmwillis:
We still want a seed for reproducibility. Here's the flow I am envisioning:

1. Pre- or Post- commit process runs tests in a random order
2. Reporter logs the seed from `QUnit.config.seed`
3. User can then plug that seed into the browser locally to debug any test failures

So when I say I don't foresee usage in the browser, I specifically meant usage of the "run this with a random seed", since that just seems like a poor development flow. Instead random seeds should primarily be used as a form of [automated fuzz testing](https://en.wikipedia.org/wiki/Fuzz_testing). Seem reasonable?

Krinkle:
@trentmwillis Yep. Sounds great!

trentmwillis:
@gibson042 @leobalter updated to call the option `config.seed` to more accurately reflect what the property does. I feel like this is good to go.

src/test.js
```diff
@@ -502,6 +512,24 @@ function synchronize( callback, priority ) {
...
+	// http://excamera.com/sphinx/article-xorshift.html
+	var sample = parseInt( generateHash( seed ), 16 ) || -1;
+	return function() {
+		sample ^= sample << 13;
+		sample ^= sample >>> 17;
+		sample ^= sample << 5;
+
+		// ECMAScript has no unsigned number type
+		if ( sample < 0 ) {
+				sample += 0x100000000;
```
leobalter: extra tab?

src/test.js
```diff
@@ -502,6 +512,24 @@ function synchronize( callback, priority ) {
 	}
 }
 
+function unitSamplerGenerator( seed ) {
+	// 32-bit xorshift, requires only a nonzero seed
```
leobalter: missing empty line before comment

trentmwillis:
Fixed the style issues and updated the documentation PR accordingly as well.

test/random.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>QUnit Randomization Test Suite</title>
+    <link rel="stylesheet" href="../dist/qunit.css">
+    <script src="../dist/qunit.js"></script>
+    <script>
+      QUnit.config.seed = "7x9";
```
gibson042: Could you move this into `random.js`? It's a little weird to see separation from the code assuming it. After that, I think we're good to go. Thanks @trentmwillis!

trentmwillis: I can do that, but we have other tests following this pattern as well `logs.html` and `string-filter.html` for example. Should we update those in a separate PR?

leobalter: > we have other tests following this pattern as well logs.html and string-filter.html for example. Should we update those in a separate PR?

Probably yes.

gibson042: Ok, let's leave this for now then. I've opened #934 for discussion.

trentmwillis: So is this good to go then?

leobalter:
Merged! :100:

Issue #879, opened 2015-10-20
Update jscs to latest


arghgr:
As mentioned in #876, grunt-jscs may not be catching certain style errors and needs to be updated to the latest version.

leobalter:
a jscs update might include jshint as well.

leobalter:
@arghgr I'm looking forward a PR to fix this one. :)

PR #936, opened 2016-02-26
Build: Update JSCS and JSHint

trentmwillis:
Addresses #879 since it didn't appear to be moving along. Little style issues sure do build up over time.


leobalter:
`npm test` is failing on my machine probably by other reasons (running from a new environment on Windows 10 with the newest version of Node).

I'll need more time before I merge this to get back to my original env and see if this branch works and potentially file an issue to fix QUnit development on Windows 10/node.

trentmwillis:
Okay, sounds good. Do you think it'd be worthwhile at all to look into integrating something like [AppVeyor](https://www.appveyor.com/) into the project? Would help us verify things work with Windows environments.

leobalter:
It's an extra CI step that I wouldn't mind to have. This would allow tests for node-chakra as well, eventually.

leobalter:
It works here, and on Node 4 to 5.7.0 too.

I'll check what is happening on Windows on another issue

Issue #896, opened 2015-11-30
hidepassed needs to check for the right element


jzaefferer:
From #819:

---

Just got around to looking into this. The problem above can be reproduced by removing the element with id `qunit`.

Basically, we wanted to have our own header, so we had manually added the following HTML to the page:

```html
   <!--
    <div id="qunit"></div>
   -->
    <h1 id="qunit-header">
          // Custom header here
    </h1>
    <div id="qunit-testrunner-toolbar"></div>
    <ol id="qunit-tests"></ol>
```

This actually works, except that `hidepass` is not honoured.

The problem code is the following in `QUnit.begin`

```JavaScript

	if ( qunit && config.hidepassed ) {
		addClass( qunit.lastChild, "hidepass" );
	}

```

This code is inconsistent, because the `hidepass` is actually in `qunit-testrunner-toolbar` and existence of that element should imply that the code runs correctly. Probably

```
	addClass( qunitToolbar.parentElement.lastChild, "hidepass" );
```

leobalter:
good catch. I'll see if I can fix it soon.

PR #939, opened 2016-03-02
HTML Reporter: Fix hidepassed element check

trentmwillis:
Addresses #896.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -533,7 +533,8 @@ function appendTest( name, testId, moduleName ) {
 
 // HTML Reporter initialization and load
 QUnit.begin(function( details ) {
-	var qunit = id( "qunit" );
+	var tests,
+		qunit = id( "qunit" );
 
 	// Fixture is the only one necessary to run without the #qunit element
 	storeFixture();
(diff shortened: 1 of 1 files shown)


reporter/html.js
```diff
@@ -533,7 +533,8 @@ function appendTest( name, testId, moduleName ) {
 
 // HTML Reporter initialization and load
 QUnit.begin(function( details ) {
-	var qunit = id( "qunit" );
+	var qunit = id( "qunit" ),
+		tests;
```
leobalter: from the style guide, variables declared without an initializer (`= ...`) must be declared first. Also, we should use a single `var` for each variable.

leobalter:
Just a small nit pick, but it LGTM.

reporter/html.js
trentmwillis: Will update. I don't see that particular rule mentioned in the [jQuery JS style guide](https://contribute.jquery.org/style-guide/js/) though, or perhaps I need to look elsewhere? Regardless, should probably get #936 merged so that it's easier to pick up on these before PR.

leobalter: you're right, it is not there anymore. :)

trentmwillis:
Fixed the nitpick.

leobalter:
Merged, thanks!

Issue #940, opened 2016-03-02
Development is broken on Windows 10 environment


leobalter:
The current development is broken on Windows 10 environment.

There are more than one breaking point. The first is on grunt-contrib-concat, where it concat files with the Windows EOL character. JSCS throws a code style error.

```
λ grunt jscs
Running "jscs:all" (jscs) task
Invalid line break at dist/qunit.js :
    11 |
    12 |( function( global ) {
    13 |
--------^
    14 |var QUnit = {};
    15 |
>> 1 code style errors found!
Warning: Task "jscs:all" failed. Use --force to continue.
``` 

The other error is due to the test-on-node tests on QUnit.stack (`test/main/stack`), where it fails at `QUnit.stack: returns the proper stack line`.

I did a clean install of npm packages both on my original environment and windows environment. Both are using Node 5.7.0 and npm 3.6.0.

I need to investigate this further.

leobalter:
As @trentmwillis suggested at #936, it's interesting to use a windows based CI to avoid this in the future.

trentmwillis:
Bah, auto-closed but there is still one more fix that needs to go in, so reopening.

trentmwillis:
Fixes have been pushed. Should we use this issue or open a new one to track the possibility of using a Windows-based CI?

leobalter:
It's better if we keep track of it on another issue (#943), as it is not critical or urgent as the others already addressed here.

Thanks!

PR #941, opened 2016-03-02
Build: Fix linefeed to LF on Grunt configuration

leobalter:
Plus, updates grunt-contrib-concat to 1.0.0

Ref gh-940

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,11 +19,15 @@ function process( code, filepath ) {
 		.replace( /@DATE/g, ( new Date() ).toISOString().replace( /:\d+\.\d+Z$/, "Z" ) );
 }
 
+// Prevents CRLF EOL on Windows environments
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -19,11 +19,14 @@ function process( code, filepath ) {
 		.replace( /@DATE/g, ( new Date() ).toISOString().replace( /:\d+\.\d+Z$/, "Z" ) );
 }
 
+// Prevents CRLF EOL on Windows environments
```
trentmwillis: Might be worth mentioning that this is important for style checks after the concat, otherwise seems a bit arbitrary

leobalter: I included this second line: https://github.com/jquery/qunit/pull/941/files#diff-35b4a816e0441e6a375cd925af50752cR23

trentmwillis: Good enough for me. Assuming CI passes (should), I'll merge.

PR #942, opened 2016-03-02
Tests: Fix QUnit.stack test to account different path systems

### LINKED ISSUE #940: Development is broken on Windows 10 environment

leobalter:
Fixes #940

Let's check how this goes on the CI.

That was the last bug to fix #940 and have a fine development suite on Windows.

### DIFF
--- a/test/main/stack.js
+++ b/test/main/stack.js
@@ -5,7 +5,7 @@
 
 	// Flag this test as skipped on browsers that doesn't support stack trace
 	QUnit[ stack ? "test" : "skip" ]( "returns the proper stack line", function( assert ) {
-		assert.ok( /\/test\/main\/stack\.js/.test( stack ) );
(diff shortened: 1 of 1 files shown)


trentmwillis:
:+1: looks good to me. Silly Windows and their different separator.

Issue #938, opened 2016-02-26
Add warning if `!details.result` and `actual === expected` for unknown reason.


Parakleta:
In [reporter/html.js#L674-L713](https://github.com/jquery/qunit/blob/65706a1cef576d008b2a714a21cc98a37cf54f29/reporter/html.js#L674-L713) if the test has failed a decision is made about what to actually report.

Prior to 87061fb82bcf9a248c56ccf235b164e989ae2935 the check that `actual != expected` was required to suppress printing the `actual` and difference since there was no `if (!result)` test.  After that change the `actual != expected` test only served to suppress the diff calculation when there would be no diff.  Now (as of ecc33ac75eda91bbe8f31349211325a127bfb913) the `else` clause of that test is also used together with the check that `expected` contains an unexplored array or object to warn that the `QUnit.dump.parse()` has insufficient `QUnit.dump.maxDepth` to express a difference between the objects.

I'd like to propose that a general warning be added as an additional `else` clause to catch all other cases where the test has failed but the `QUnit.dump.parse()` couldn't detect the difference.  The absence of any mention of the `actual` value in some failing tests has recently caused me some confusion because I couldn't tell if the lack of reporting was a bug in QUnit or a problem with my tests. It turned out in this case (obviously in hindsight) that `deepEqual` was needed to compare `NaN`, but other causes of failure could be differing document objects or functions.

trentmwillis:
@Parakleta I'm having a bit of trouble groking what the precise issue you're describing is. Can you provide a reproduction jsfiddle or something similar? Or, at least, what you feel this warning should say?

Parakleta:
This [fiddle](https://jsfiddle.net/parakleta/37s44y1n/) shows three cases where the test fails but the `expected` and `actual` appear the same according to `QUnit.dump.parse()` and so there is no reporting of why the test failed.  I think I'd just like if there was some message such as for the `maxDepth` case, something like:

"Message: Diff suppressed as the expected and actual results have an equivalent serialization"

trentmwillis:
Okay, thanks a lot for the examples! This seems like a reasonable ask, would you mind submitting a PR for it? We might work the exact wording of the error a bit, but I like the idea.

PR #944, opened 2016-03-02
Message explaining missing Report Diff.

Parakleta:
As per #938

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -713,6 +713,10 @@ QUnit.log( function( details ) {
 				" run with a higher max depth or <a href='" +
 				escapeText( setUrl( { maxDepth: -1 } ) ) + "'>" +
 				"Rerun</a> without max depth.</p></td></tr>";
+		} else {
+			message += "<tr class='test-message'><th>Message: </th><td>" +
+				"Diff suppressed as the expected and actual results have an equivalent" +
+				" serialization</td></tr>";
 		}
 
 		if ( details.source ) {


Parakleta:
I've signed the CLA, but the 'check the status' link just returns 404.

leobalter:
maybe it needs an empty commit to check the CLA again.

cc @jzaefferer

trentmwillis:
Seems like maybe the email or full name you provided might not match up with your git config. You can check what those are set as by running `git config -l` in your repository. The `user.name` and `user.email` properties should match what you put in the CLA.

Parakleta:
I don't use the cli tools for PRs, just the built in github editor.  If you click on the commit and add `.patch` to the end of the URL you get the patch email which has 'Parakleta <parakleta@darkreality.org>' as the `From: ` field which is the name and email I entered in the CLA.  How does your CLA tool work?

trentmwillis:
I'm not super familiar with that code, but it looks like it may require the name to contain at least two parts: https://github.com/jquery/jquery-license/blob/master/lib/validate.js#L18-L25

@jzaefferer seems like he would have more insight.

Parakleta:
That looks like it, someone will have to manually put it through then.

jzaefferer:
I don't have access to do the manual confirmation. @scottgonzalez can you help out here?

kborchers:
This should be fixed now @leobalter @Parakleta. Sorry for the delay.

trentmwillis:
Thanks again @kborchers!

@Parakleta sorry for the wait on this, but would you mind adding a test case for this as well?

leobalter:
for the records, @csnover was coming to help too. 

Thanks @kborchers and @csnover.

leobalter:
LGTM

reporter/html.js
```diff
@@ -713,6 +713,10 @@ QUnit.log( function( details ) {
 				" run with a higher max depth or <a href='" +
 				escapeText( setUrl( { maxDepth: -1 } ) ) + "'>" +
 				"Rerun</a> without max depth.</p></td></tr>";
+		} else {
+			message += "<tr class='test-message'><th>Message: </th><td>" +
```
leobalter: this is consistent with [line 710](https://github.com/jquery/qunit/pull/944/files#diff-5b7e94e2a10b848c4ae4b539388f86aeR710) but I wonder if we should keep this `<th>Message: </th>`.

Anyway, this should be addressed in another patch.

leobalter:
Thanks, @Parakleta! This is merged at  8460ced

Issue #943, opened 2016-03-02
Use a windows based CI


leobalter:
Ref #940

cc @trentmwillis

trentmwillis:
A quick search for ["windows ci"](https://www.google.com/#q=windows+ci), doesn't yield a lot of promising results. I'm familiar with using [AppVeyor](https://www.appveyor.com/) and know that the Ember-CLI project also uses it. It is also quite similar to Travis and is easy to set up, so this may be the best option unless anyone has another recommendation.

leobalter:
I am ++ to use AppVeyor.

trentmwillis:
I'll look into setting it up sometime soon. What scripts should we run? I'm thinking we can just do `npm test`, I don't see much reason to run coveralls and browserstack in both CIs.

leobalter:
npm test is the one we should care about.

PR #946, opened 2016-03-03
Build: Introduce AppVeyor for Windows CI

trentmwillis:
Introduces a simple AppVeyor config to enable Windows CI as discussed in #943.

Edit: Unfortunately, looks like it won't actually do an AppVeyor build until after the config has been committed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+# https://www.appveyor.com/docs/build-configuration
+
+# Test against this version of Node.js
+environment:
+  nodejs_version: "4"
+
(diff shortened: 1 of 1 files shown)


leobalter:
Do we need to setup anything on AppVeyor before?

trentmwillis:
I've already enabled it and ran one build (which failed since there is no configuration). After merge it should run for all PRs/merges.

leobalter:
It's merged. Let's see how it goes! :+1:

leobalter:
It worked just fine! That's great!

PR #948, opened 2016-03-04
Readme: Add AppVeyor badge

trentmwillis:
Adding a badge for the AppVeyor build. Also testing to make sure AppVeyor picks up PRs.

It looks like AppVeyor hosts the project under the account name, which in this case is my username, so the CI url is: https://ci.appveyor.com/project/trentmwillis/qunit. Is there or should we set up a "jquery" account so that the URL can look more correct? If so, who can do that?


trentmwillis:
So looks like AppVeyor doesn't have great support for projects owned by GitHub orgs ([relevant article](http://help.appveyor.com/discussions/questions/744-using-appveyor-for-a-github-org)).

@stefanpenner do you have any insight on how Ember-CLI handles this? Did you just set up an extra account?

stefanpenner:
I set up an ember-cli account

trentmwillis:
Okay, cool. Thanks for that info.

@leobalter can you set up a jquery or qunit account that we can use?

leobalter:
It is also running on https://ci.appveyor.com/project/leobalter/qunit. (at least the hook is set to my acc) :P

The link should be https://ci.appveyor.com/project/jquery/qunit. I don't know if someone else at the jQuery Foundation handled it. I couldn't find it being used on other projects from the org (Mobile, UI, Globalize, PEP, etc).

I'll ping @jzaefferer as he might have access to a jQuery org-acc. Meanwhile, let's keep it as it is now. This should not prevent this PR from being merged (LGTM), we may change the badge url later and the badge result is based on the last build regardless the acc.

trentmwillis:
Sounds good. I'll let you handle that. Meanwhile I'll update the badge link to point to your account (since it looks like it took precedence from mine) and then merge this.

Issue #915, opened 2015-12-31
QUnit.equiv treats Map and Set as ordered


gibson042:
https://jsfiddle.net/xd2jnbLa/

I think we should ignore insertion order when comparing these objects ~~(along with WeakMap and WeakSet, which we ignore completely)~~.

leobalter:
There is no way to discover the collections on WeekMaps and WeakSets, that's why it's ignored and compared as regular objects.

> we should ignore insertion order

+1 to this.

PR #949, opened 2016-03-05
Assert: Ignore insertion order for Maps and Sets

trentmwillis:
Addresses #915. Not sure this is the best algorithm in terms of performance, but it gets the job done and is easy to follow.

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -131,18 +131,28 @@ QUnit.equiv = ( function() {
 		},
 
 		"set": function( b, a ) {
-			var aArray, bArray;
+			var innerEq,
+				outerEq = true;
 
-			aArray = [];
-			a.forEach( function( v ) {
-				aArray.push( v );
-			} );
-			bArray = [];
-			b.forEach( function( v ) {
-				bArray.push( v );
+			if ( a.size !== b.size ) {
+				return false;
+			}
+
+			a.forEach( function( aVal ) {
+				innerEq = false;
+
+				b.forEach( function( bVal ) {
+					if ( innerEquiv( bVal, aVal ) ) {
+						innerEq = true;
+					}
+				} );
+
+				if ( !innerEq ) {
+					outerEq = false;
+				}
 			} );
 
-			return innerEquiv( bArray, aArray );
+			return outerEq;
 		},
 
 		"map": function( b, a ) {
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -146,18 +146,28 @@ QUnit.equiv = ( function() {
 		},
 
 		"map": function( b, a ) {
-			var aArray, bArray;
+			var innerEq,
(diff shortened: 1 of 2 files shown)


src/equiv.js
```diff
@@ -131,33 +131,53 @@ QUnit.equiv = ( function() {
...
+
+				b.forEach( function( bVal ) {
+					if ( innerEquiv( bVal, aVal ) ) {
+						innerEq = true;
+					}
+				} );
+
+				if ( !innerEq ) {
+					outerEq = false;
+				}
```
gibson042: Checking Sets and Maps means we _have_ Sets and Maps, so this could be as simple as:
```js
a.forEach( function( aVal ) {
	eq = eq && b.has( aVal );
} );
```
or the faster but more complex
```js
var iterator = a.values(),
	item = iterator.next();
for ( ; !item.done; item = iterator.next() ) {
	if ( !b.has( item ) ) {
		return false;
	}
}
return true;
```
with analogous options for Map (maybe even _identical_ if both use `entries`).

trentmwillis: So I tried this approach. However, it seems `has` simply checks the reference for objects. That means if one of the entries has identical objects but they're not the same object, the check will fail. That's not the current behavior and I don't believe it is the behavior we want.

gibson042: Ah, of course! Can you add a test for that case (Sets containing distinct but deeply-equal non-primitive values with different insertion orders)?

trentmwillis: Will do. I'll add an equivalent test for Maps as well.

gibson042: I think that one already exists: https://github.com/jquery/qunit/pull/949/files#diff-d901125d4db3b69d066f186fe4a455b1R1834

trentmwillis: That doesn't cover the different insertion orders part of your ask. For both Maps and Sets, there are tests for distinct but deeply-equal non-primitive values and tests for different insertion orders, but none that do both together.

trentmwillis:
@gibson042 added some additional test cases. I think this should be good now.

@leobalter AppVeyor seems to have failed due to runtime being too long. Looks to be a transient issue, could you set me up as a team member on AppVeyor so I can restart when issues like those arise?

leobalter:
I might have set it. I tried to add the Github's QUnit team as admin. Please tell if it works.

trentmwillis:
@leobalter it doesn't seem to have worked. I think you'll have to add individuals as I don't think AppVeyor has good support for organizations

leobalter:
@trentmwillis would you try it again?

I couldn't add @gibson042's email there, it says it's not valid. Maybe he should register.

We need to check with @jzaefferer about creating a team acc there, which should probably be a jQuery acc.

Sorry for side tracking this issue.

Issue #824, opened 2015-06-06
Quoted "<" breaks diff oO


FagnerMartinsBrack:
Using version 1.18.0: http://jsfiddle.net/yptmch4k/1/

When I compare the string `c=<` against `c="<"` using `strictEqual` the diff breaks, is it intended?

![image](https://cloud.githubusercontent.com/assets/835857/8020973/89b50a14-0c64-11e5-9294-88fac7257360.png)

jzaefferer:
For the record, no, that's not intended. Looking for help to get this fixed.

PR #950, opened 2016-03-05
HTML Reporter: Fix escaping of diffs

trentmwillis:
Attempts to address #824.

The root cause was that the `actual` and `expected` values were escaped before getting passed into `QUnit.diff()`. This in odd behavior for diffs that included entities. The proposed solution is to allow `QUnit.diff()` to perform the escaping as it constructs the output HTML.

One open question on this: I've copied the `escapeText` function from the reporter, this causes duplicate code. However, it seems like the alternative would be to add it to the `QUnit` global. Which approach seems better? I tend to think the latter, but went with the former to prevent polluting the global more.


gibson042:
First of all, thanks! Addressing the deep issues here is long overdue.

> One open question on this: I've copied the `escapeText` function from the reporter, this causes duplicate code. However, it seems like the alternative would be to add it to the `QUnit` global. Which approach seems better? I tend to think the latter, but went with the former to prevent polluting the global more.

I'd rather see `QUnit.diff` itself move into the HTML reporter, since that is the site of its [only internal use](https://github.com/jquery/qunit/blob/a685a17131a746260c55bfeb522eab512f3d6d77/reporter/html.js#L697) (which is fitting, since the function [_returns HTML_](https://github.com/jquery/qunit/blob/0f9ae3443a0fc78d96baa2eb86f90b0cba9264b5/src/diff.js#L1162)). I'd also be open to rearchitecting it for external separation of patch generation from presentation, but let's start with just a move.

reporter/html.js
```diff
@@ -676,20 +676,20 @@ QUnit.log( function( details ) {
...
 		if ( details.negative ) {
-			expected = escapeText( "NOT " + QUnit.dump.parse( details.expected ) );
+			expected = "NOT " + QUnit.dump.parse( details.expected );
 		} else {
-			expected = escapeText( QUnit.dump.parse( details.expected ) );
+			expected = QUnit.dump.parse( details.expected );
 		}
 
-		actual = escapeText( QUnit.dump.parse( details.actual ) );
+		actual = QUnit.dump.parse( details.actual );
```
gibson042: Note the [risk of regression](https://github.com/jquery/qunit/blame/a32ccfaefbe75bea2a4c84f46656d560e2f2fbed/History.md#L531) here. Do we have any tests protecting us from faulty assumptions regarding escaped vs. unescaped input to `QUnit.diff`?

trentmwillis:
> I'd rather see `QUnit.diff` itself move into the HTML reporter

Makes sense; I hadn't considered that option before. Will try that approach.

reporter/html.js
trentmwillis: What sort of test would you expect? We have tests of the output of `QUnit.diff` though I'm not sure those cover the cases you're alluding to.

gibson042: I would like to see assertions covering the behavior of `QUnit.diff` [when provided strings containing escaped and/or unescaped HTML](https://jsfiddle.net/gL6djqm0/) (and also Javascript escape sequences, for that matter). Looking at [test/reporter-html/diff.js](https://github.com/jquery/qunit/blob/0f9ae3443a0fc78d96baa2eb86f90b0cba9264b5/test/reporter-html/diff.js), I see this as a sizable gap.

trentmwillis: Okay, will add those. Is `QUnit.diff` considered public API? Because this change will definitely alter the behavior of the function.

leobalter: > Is QUnit.diff considered public API?

yes.

> Because this change will definitely alter the behavior of the function.

If it's not a fix, but a breaking change, we might consider deprecating it.

trentmwillis: > If it's not a fix, but a breaking change, we might consider deprecating it. (from the public api)

It's the correct fix for the reported issue, but it is also breaking because the current behavior is broken. We should be escaping the values we put into the returned markup rather than assuming the passed in values are correctly escaped.

I don't necessarily see a reason to remove the function from the public API, but we will probably want to deprecate the current behavior and move to the new behavior in 2.0.

leobalter: > breaking because the current behavior is broken

Unless the bug becomes a feature. :)

I'm +1 to merge this if we address the [test gap](https://github.com/jquery/qunit/pull/950#discussion_r55228353) mentioned by @gibson042

trentmwillis: Okay, will try to work on that at some point this week.

trentmwillis:
@gibson042 I moved `QUnit.diff` into the HTML reporter closure. Note that without additional work, this means that the function will no longer be available to non-browser environments. Not sure if that is what we want or not.

One other option would be to move the html-reporter closure into the main closure and pull `escapeText` into the main `utilities` file. This would preserve the current scoping of `diff` while also allowing us to not change the available environments.

reporter/diff.js
```diff
@@ -26,7 +26,7 @@
  * Usage: QUnit.diff(expected, actual)
  *
  */
-QUnit.diff = ( function() {
+window.QUnit.diff = ( function() {
```
gibson042: This shouldn't be necessary, and might very well be harmful.

gibson042:
> I moved `QUnit.diff` into the HTML reporter closure. Note that without additional work, this means that the function will no longer be available to non-browser environments. Not sure if that is what we want or not.

That is acceptable to me because `QUnit.diff` returns HTML. Later fixes will [separate patch generation from presentation](https://github.com/jquery/qunit/pull/950#issuecomment-192704600), and presumably include deprecating the public `diff` name altogether rather than give it a completely different signature.

leobalter:
> That is acceptable to me because QUnit.diff returns HTML.

agreed.

reporter/diff.js
trentmwillis: Oh yeah, good catch. This was a remnant from trying to debug a test failure before I realized that the diff tests were running in node.

trentmwillis:
Okay, makes sense. I added the requested tests, let me know if anything else is needed.

leobalter:
LGTM.

@gibson042 ?

gibson042:
LGTM.

Issue #951, opened 2016-03-09
Suggestion [HTMLReporter]: make it possible to pass comma separated modules to `module` param


Arkni:
Hi all,

I wonder if it's possible to specify a comma separated names, instead of specifying one module to run, like so:
```
http://localhost:8000?module=module1,module2
```

or
```
http://localhost:8000?module=module1&module=module2
```

Thanks :)

trentmwillis:
You could accomplish this using the recently introduced [regex filtering](https://github.com/jquery/qunit/commit/e97213599c87a505d599386b4b98e25d4b5ee14d):

```
http://localhost:8000?filter=/module1|module2/
```

We should probably add documentation for that feature...will try to do that this week.

Arkni:
Please correct me if I'm wrong.
So if I'm using a filter like: `filter=core`, this will run all the tests that are under the module `core`, or also the tests that have `core` in the name.

From the commit you linked, this only available starting from `v1.21.0`.

> We should probably add documentation for that feature...will try to do that this week.

I appreciate that :)

Arkni:
Also, the config option `moduleFilter` doesn't seem to work in latest releases.
**Simple code:**
```js
// Only the module `module1` should run, unfortunately doesn't work
QUnit.config.moduleFilter = "module1";
QUnit.module( "module1" );
QUnit.test( "test11", function( assert ) {
	assert.ok( true );
} );
QUnit.test( "test12", function( assert ) {
	assert.ok( true );
} );
QUnit.module( "module2" );
QUnit.test( "test21", function( assert ) {
	assert.ok( true );
} );
QUnit.test( "test22", function( assert ) {
	assert.ok( true );
} );
```

**JSfiddle:** https://jsfiddle.net/Arkni/tg6pzL4a/

Arkni:
I tested the `filter` option , even though it's not what I'm after. But I can live with that.
Feel free to close this issue if you see it as resolved. But please notice the issue I mentioned in my previous comment (https://github.com/jquery/qunit/issues/951#issuecomment-194515024).

trentmwillis:
@leobalter can you give more insight into removing `config.moduleFilter`? Seems like it was intentional. If so, we should update docs. Also, how would someone now programmatically set the module filter?

@Arkni I can understand your desire for a multi-module filter. I think the primary concern I have is introducing too many ways for users to filter their tests. Additionally, the current module filter is tied to a drop-down, how would we represent multiple modules in that sort of UI?

[...]

PR #952, opened 2016-03-14
Core: Handle multiple moduleId parameters

mlato-yahoo:
What: Adding moduleId support for multiple module ids
Why: Enable users to run a subset of all modules, rather than one/all modules. Adds symmetry to the existing testId feature.This will enable UI improvements, which we can now explore.

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -90,12 +90,15 @@ if ( urlParams.seed ) {
 	}
 }
 
-config.testId = [];
-if ( urlParams.testId ) {
+function getIds ( params ) {
+	var paramArray = [];
+	if ( params ) {
 
-	// Ensure that urlParams.testId is an array
-	urlParams.testId = decodeURIComponent( urlParams.testId ).split( "," );
-	for ( var i = 0; i < urlParams.testId.length; i++ ) {
-		config.testId.push( urlParams.testId[ i ] );
+		// Ensure that params is an array
+		paramArray = decodeURIComponent( params ).split( "," );
 	}
+	return paramArray;
 }
+
+config.testId = getIds( urlParams.testId );
+config.moduleId = getIds( urlParams.moduleId );
--- a/src/test.js
+++ b/src/test.js
@@ -360,7 +360,8 @@ Test.prototype = {
 	},
 
 	valid: function() {
-		var filter = config.filter,
+		var moduleId,
+			filter = config.filter,
 			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
 			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
 			fullName = ( this.module.name + ": " + this.testName );
@@ -379,6 +380,13 @@ Test.prototype = {
 		// Internally-generated tests are always valid
 		if ( this.callback && this.callback.validTest ) {
 			return true;
+		}
+
+		if ( config.moduleId.length >  0 ) {
+			moduleId = generateHash( this.module.name );
+			if ( inArray( moduleId, config.moduleId ) < 0 ) {
+				return false;
+			}
 		}
 
 		if ( config.testId.length > 0 && inArray( this.testId, config.testId ) < 0 ) {
--- a/test/moduleId.js
+++ b/test/moduleId.js
@@ -1,0 +1,19 @@
+QUnit.config.moduleId = [ "720ab266", "0af5a573" ];
+
+QUnit.module( "QUnit.config.moduleId.foo" );
+
+QUnit.test( "foo test should be run", function( assert ) {
+	assert.ok( true, "foo test should be run" );
+} );
+
+QUnit.module( "QUnit.config.moduleId.bar" );
+
+QUnit.test( "bar test should be run", function( assert ) {
+	assert.ok( true, "bar test should be run" );
+} );
+
+QUnit.module( "QUnit.config.moduleId.foobar" );
+
+QUnit.test( "foobar test should not be run", function( assert ) {
+	assert.ok( false, "foobar test should not be run" );
+} );
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -113,6 +113,7 @@ grunt.initConfig( {
 			"test/reporter-html/legacy-markup.html",
 			"test/reporter-html/no-qunit-element.html",
 			"test/reporter-html/single-testid.html",
+			"test/moduleId.html",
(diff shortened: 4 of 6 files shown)


mlato-yahoo:
I just signed the CLA.

src/test.js
```diff
@@ -362,7 +362,7 @@ Test.prototype = {
 	valid: function() {
 		var filter = config.filter,
 			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
-			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
+			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(), moduleId,
```
leobalter: would you mind positioning this unnassigned `moduleId` variable to the top line of the var declarations? At least to use a single line for each one.

test/moduleId.js
```diff
@@ -0,0 +1,19 @@
...
+QUnit.module( "QUnit.config.moduleId.bar" );
+
+QUnit.test( "bar test should be run", function( assert ) {
+	assert.ok( true, "bar test should be run" );
+} );
+
+QUnit.module( "QUnit.config.moduleId.foobar" );
+
+QUnit.test( "foobar test should not be run", function( assert ) {
+  assert.ok( false, "foobar test should not be run" );
```
leobalter: indent with tabs instead of spaces

leobalter:
Just a minor concern: `testId` is registered on for each test, maybe we should register `moduleId` as well. This way makes it easier to identify the module's hash. This might be addressed later, in a further PR, though.

Would you mind to add the docs for it? Refs: http://api.qunitjs.com/QUnit.config/ and http://github.com/jquery/api.qunitjs.com

> I just signed the CLA.

Some rebased/squashed or even an empty commit might fix the CLA check. The link on the error message would be useful if it wasn't 404ing. So just ping me if you need more help.

src/core/config.js
```diff
@@ -91,11 +91,22 @@ if ( urlParams.seed ) {
...
 }
+
+if ( urlParams.moduleId ) {
+
+	// Ensure that urlParams.moduleId is an array
+	urlParams.moduleId = decodeURIComponent( urlParams.moduleId ).split( "," );
+	for ( i = 0; i < urlParams.moduleId.length; i++ ) {
+		config.moduleId.push( urlParams.moduleId[ i ] );
+	}
+}
```
trentmwillis: Since this logic is identical to the logic for `testId`, could we abstract this into a function? Maybe something like:

```js
config.moduleId = getIds(urlParams.moduleId);
```

src/test.js
```diff
@@ -381,6 +381,13 @@ Test.prototype = {
 			return true;
 		}
 
+		if ( config.moduleId.length >  0 ) {
+			moduleId = generateHash( this.module.name );
```
trentmwillis: Similar to @leobalter's comment:

> Just a minor concern: testId is registered on for each test, maybe we should register moduleId as well. This way makes it easier to identify the module's hash.

If we're only generating the `moduleId` when a module id is provided as a URL param, how do users know what a module's id is?

trentmwillis:
Overall, I am :+1: on this idea. I think it will provide a solution to #951 while not conflating with the current module filter implementation.

cc @Arkni does this seem like a good compromise?

src/core/config.js
mlato-yahoo: Good idea. I've added a getIds function.

Arkni:
> cc @Arkni does this seem like a good compromise?

This looks like what I'm after, thanks for tagging me.
:+1: to this implementation.

Just one issue as you said in your comment above (https://github.com/jquery/qunit/pull/952#discussion_r56055346), how do a user, like me, know the ID of a specific module ?

mlato-yahoo:
@trentmwillis @Arkni currently a user wouldn't have a way of knowing the ID of a specific module.
Registering the moduleId for each module (as @leobalter suggested) would create an index accessible to anyone digging into the config, but it wouldn't be accessible through the UI.
Currently testIds are exposed when a user clicks "Rerun" next to a test. Would something similar for modules suffice?
Ultimately I could see the module selector dropdown be replaced with a multi-selector, but that's a larger UI change. What do you guys think?

leobalter:
> Ultimately I could see the module selector dropdown be replaced with a multi-selector, but that's a larger UI change.

We should progress this feature to a UI change for sure. I'm ok if we have it documented on the api page and follow up the missing parts on another PR. It makes the PR easier to review, merge and avoid frustration as long PRs.

mlato-yahoo:
Sounds good to me @leobalter . I'll go ahead and make a PR with the documentation changes.

Arkni:
> @trentmwillis @Arkni currently a user wouldn't have a way of knowing the ID of a specific module.
> Registering the moduleId for each module (as @leobalter suggested) would create an index accessible to anyone digging into the config, but it wouldn't be accessible through the UI.

Make sense.

Arkni:
@mlato-yahoo @leobalter 

I don't know if this will make sense or not, but here is a suggestion. Instead of passing an array of hashes (as you did in your test), we use an array of module names (case sensitive). Then, check if the hash of each module's name is found in the index of moduleId, and only run the matched modules.

wdyt ?

stefanpenner:
> Instead of passing an array of hashes (as you did in your test), we use an array of module names (case sensitive).

that would be an extremely long list.

Remember, this is largely aimed as an internal refactor to enable UI enhancements. It is ok that it isn't directly accessible to everyday users yet.

Arkni:
> that would be an extremely long list, i would imagine we would easily hit some limit.
> 
> Remember, this is largely aimed as an internal refactor to enable UI enhancements. It is ok that it isn't directly accessible to everyday users yet.

Make sense, thanks for clarifying the situation for me :)

mlato-yahoo:
I'm not sure why the AppVeyor and Travis CI builds are failing now - can someone re-trigger them?

trentmwillis:
@mlato-yahoo looks good to me. One minor nitpick, can we make the commit message a bit more clear? This is mainly for when we publish release notes; saying "handle multiple moduleId parameters" makes it sound like this is an extension of an existing feature. Maybe "Introduce moduleId filtering"?

After that I am good to merge this.

leobalter:
@trentmwillis I just did the renaming on the commit, as we already need to amend it to reference the respective issue and PR.

I also included a reference for the UI issue.

Thanks @mlato-yahoo, @Arkni and @trentmwillis!

Issue #954, opened 2016-03-15
Document and prefer multiply-specified URL parameters over comma-separated


gibson042:
Ref gh-951

`QUnit.urlParams` already stores repeated URL parameters in arrays: https://github.com/jquery/qunit/blob/0f9ae3443a0fc78d96baa2eb86f90b0cba9264b5/src/core/utilities.js#L115
For consistency with `QUnit.config` and to preclude ambiguous parsing, we should prefer that method over comma-separated values for specifying composite parameters (e.g., `?moduleId=720ab266&moduleId=0af5a573` instead of `?moduleId=720ab266,0af5a573`), and deprecate the latter anywhere it's already documented.

trentmwillis:
I was unaware of the auto-concatenating of repeated params. I agree we should standardize on a single format. However, I'm not a huge fan of the first approach due to verbosity, but I guess it makes more sense than attempting to figure out if a comma is meant as a separator or if it is just part of the value.

leobalter:
The changes for gh-951 are not released yet, do you want me to revert until we fix it? If we fix it soon we don't need to mind on any breaking change (or reverting commits).

gibson042:
No, I wouldn't revert because e47a7c3db1572387e62846e2e4f39acf7db528f2 just carried forward the already-present pattern of [`testId`](https://github.com/jquery/qunit/commit/e47a7c3db1572387e62846e2e4f39acf7db528f2#diff-217941a76651434529372f12f6de3cbe) (and we don't have any documentation claiming support for comma-separated values, just code).

Issue #754, opened 2015-02-27
`testIds` are invalidated when using toolbar options


lamchau:
**Version**: 1.17.1 / 1.17.2-pre

**Steps to repro**:
 1. Specify two (2) `testId` (e.g. `http://localhost:3003/?testId=1a9c4dda&testId=f0342ae5`)
 2. Click/select 'Check for Globals'

**Result**: Test refreshes and the URL is encoded `http://localhost:3003/?testId=1a9c4dda%2Cf0342ae5` and none of the tests run.

**Expected**: Same are repro step 1 but with option flag(s) set.

Related to: #569

leobalter:
@lamchau, your 7069029 seems to solve the problem.

Having two `testId`s set in the url params is not supported, and I don't recommend using it,  Maybe filtering by the module would be better.

Even though, I don't see any reason to not ship your commit.

I'm waiting for the PR.

lamchau:
Thanks! Just signed the CLA! 

I was looking for tests to write for parity with `config.testId = [...]` before I submitted a PR -- but I didn't see any.

leobalter:
We don't have proper tests for the urlParams yet. :(

PR #957, opened 2016-03-15
Core: Stop splitting URL parameter values by commas

gibson042:
Repeating parameters is the preferred mechanism for specifying
multi-valued options by URL.

Fixes gh-954

Since we don't have any documentation of URL parameters, this doesn't break an official API.

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -90,15 +90,6 @@ if ( urlParams.seed ) {
 	}
 }
 
-function getIds ( params ) {
-	var paramArray = [];
-	if ( params ) {
-
-		// Ensure that params is an array
-		paramArray = decodeURIComponent( params ).split( "," );
-	}
-	return paramArray;
-}
-
-config.testId = getIds( urlParams.testId );
-config.moduleId = getIds( urlParams.moduleId );
+// Force array-valued testId and moduleId configuration values
+config.testId = [].concat( urlParams.testId || [] );
+config.moduleId = [].concat( urlParams.moduleId || [] );
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -281,10 +281,9 @@ function setUrl( params ) {
 	params = QUnit.extend( QUnit.extend( {}, QUnit.urlParams ), params );
 
 	for ( key in params ) {
-		if ( hasOwn.call( params, key ) ) {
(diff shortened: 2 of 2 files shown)


src/core/config.js
```diff
@@ -90,15 +90,6 @@ if ( urlParams.seed ) {
...
-		paramArray = decodeURIComponent( params ).split( "," );
-	}
-	return paramArray;
-}
-
-config.testId = getIds( urlParams.testId );
-config.moduleId = getIds( urlParams.moduleId );
+// Force array-valued testId and moduleId configuration values
+config.testId = [].concat( urlParams.testId || [] );
+config.moduleId = [].concat( urlParams.moduleId || [] );
```
leobalter: from a performance perspective we could use `Array.prototype.concat.call` here.

leobalter:
LGTM

src/core/config.js
gibson042: The `urlParams` properties are not guaranteed to be arrays. I could replace the `|| []` with an `if ( urlParams.moduleId )`, but I prefer the single-line form and am not worried about creating a couple of unnecessary arrays during initialization.

leobalter: still works on:

```js
Array.prototype.concat.call( urlParams.moduleId || [] );
```

trentmwillis:
Overall :+1:, though I just want to make sure we acknowledge that even though it is not officially documented, many users are probably used to being able to use comma-separated parameter values. So we should probably get on top of [documenting the url params](https://github.com/jquery/api.qunitjs.com/issues/117).

gibson042:
I don't know about _many_ users... this was introduced a year ago in b52ceb553b712e7e94df3a1bb6bfd444aa5e771f, and as it turns out was just an easy way to deal with automatic array joining from a [properly-specified initial state](https://github.com/jquery/qunit/issues/754) (a case which, it turns out, my first stab at this had regressed on). I do agree about gh-117, though.

As for this PR, and in light of the above, do we feel comfortable introducing it now? I'm on board, but an argument could be made for waiting.

trentmwillis:
> As for this PR, and in light of the above, do we feel comfortable introducing it now?

I'm fine with introducing it now, so long as we get the documentation portion completed before the next release.

PR #959, opened 2016-03-16
Build: Update browserstack-runner from 0.3.7 to 0.4.2

### LINKED ISSUE #958: Update browserstack-runner
QUnit uses an obsolete browserstack-runner version that still uses the 3rd version of the API. BrowserStack has been requesting all jQuery teams to stop using older API versions so we need to do the upgrade ASAP.

mgol:
This update is necessary to stop using the 3rd version of the BrowserStack API.

Fixes #958

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "dependencies": {},
   "devDependencies": {
     "async": "1.4.2",
-    "browserstack-runner": "0.3.7",
+    "browserstack-runner": "0.4.2",
     "commitplease": "2.0.0",
     "grunt": "0.4.5",
     "grunt-cli": "0.1.13",


mgol:
The build breaks for some reason. Any ideas? I asked BrowserStack for help as well.

leobalter:
> The build breakson Travis for some reason.

It breaks all the time due to timeouts from Browserstack, not the runner.

I merged this as it works on my local machine. If we get more timeouts we need to investigate a better way to use it.

leobalter:
thanks!

leobalter:
I didn't update my global browserstack-runner. Yes, it is timing out more often than before on random browsers.

PR #960, opened 2016-03-16
Build: Remove browserstack-runner from CI tasks

leobalter:
The new updated version of browserstack-runner (0.4.x) is returning a
constant timeout failure on the amd tests.

Instead of reverting browserstack-runner back to the previous and
deprecated version (0.3.x), we should investigate the current error
and find a proper fix. The old version is expected to stop working
soon, as Browserstack will maintain compatibility only to the new
version.

Ref gh-959

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   },
   "scripts": {
     "browserstack": "sh build/run-browserstack.sh",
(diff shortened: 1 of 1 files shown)


leobalter:
with that said, it's good if maintainers can run browserstack locally before merging.

PR #961, opened 2016-03-16
Core: Implement moduleId support for nested modules

mlato-yahoo:
What: Supporting module nesting when filtering by `moduleId`. Also registering `moduleId` on the module itself to be consistent with `testId`.
Why: `moduleId` param did not support module nesting.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -360,21 +360,25 @@ Test.prototype = {
 	},
 
 	valid: function() {
-		var moduleId,
-			filter = config.filter,
+		var filter = config.filter,
 			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
 			module = QUnit.urlParams.module && QUnit.urlParams.module.toLowerCase(),
 			fullName = ( this.module.name + ": " + this.testName );
 
-		function testInModuleChain( testModule ) {
+		function moduleChainNameMatch( testModule ) {
 			var testModuleName = testModule.name ? testModule.name.toLowerCase() : null;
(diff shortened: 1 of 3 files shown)


src/test.js
```diff
@@ -377,16 +376,23 @@ Test.prototype = {
 			}
 		}
 
+		function moduleInModuleChain( testModule ) {
```
mlato-yahoo: Borrowed naming from the `testInModuleChain` function. Should this be changed?

gibson042: Yes, I think it would be more clear as something like `moduleChainIdMatch` (with the other one being renamed to `moduleChainNameMatch`). We could also simplify the body:
```js
return inArray( testModule.moduleId, config.moduleId ) > -1 ||
	testModule.parentModule && moduleChainIdMatch( testModule.parentModule );
```

gibson042:
Nice catch, @mlato-yahoo!

src/test.js
mlato-yahoo: Good suggestions, fixed!

mlato-yahoo:
Thanks @gibson042! Any idea why the checks are failing? 🙈

gibson042:
It looks like just general slowness: https://ci.appveyor.com/project/leobalter/qunit/build/1.0.27#L206 (from [test/logs.js](https://github.com/jquery/qunit/blob/0f9ae3443a0fc78d96baa2eb86f90b0cba9264b5/test/logs.js#L105))

At any rate, I'll verify locally before merging.

PR #962, opened 2016-03-17
Build: Split BrowserStack and Tests into separate Travis processes

trentmwillis:
Bringing this up as an alternative to #960. This will run BrowserStack and the actual Tests in different Travis processes. It will also allow BrowserStack to fail without reporting the entire build as a failure.


leobalter:
:clap: :clap: :clap: I should have done this on the other PR, but my available time is almost 0. That's great.

Issue #955, opened 2016-03-15
Confine urlParams references to browser-specific files


gibson042:
`urlParams` is obviously intended for browser utilization, but we have some references to it elsewhere. Most pressing is [test.js:Test#valid](https://github.com/jquery/qunit/blob/e47a7c3db1572387e62846e2e4f39acf7db528f2/src/test.js#L366) (which should be using a `QUnit.config` property), but I'd also like to see [`getUrlParams`](https://github.com/jquery/qunit/blob/e47a7c3db1572387e62846e2e4f39acf7db528f2/src/core/utilities.js#L115-L139) and the other `core` references ([core/config.js](https://github.com/jquery/qunit/blob/e47a7c3db1572387e62846e2e4f39acf7db528f2/src/core/config.js#L71-L104) & [core.js#L1](https://github.com/jquery/qunit/blob/e47a7c3db1572387e62846e2e4f39acf7db528f2/src/core.js#L1)) move into e.g. a "browser reporter" that sets and initializes configuration from `QUnit.urlParams` (which will probably require testing conflict resolution for cases where both inline Javascript and URL parameters try to affect the same `QUnit.config` property).

trentmwillis:
> move into e.g. a "browser reporter" that sets and initializes configuration from QUnit.urlParams

So would it make sense to introduce something like `reporter/url-params.js`? This would keep that functionality confined to the browser and give a clear separation.

> (which will probably require testing conflict resolution for cases where both inline Javascript and URL parameters try to affect the same QUnit.config property)

What should be the resolution here? I tend to think inline JS values would trump the URL params.

PR #969, opened 2016-03-21
Core: Confine URL parameter interactions to browser-specific code

gibson042:
Fixes gh-955

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -27,6 +27,8 @@ var config = QUnit.config,
 	document = window.document,
 	collapseNext = false,
 	hasOwn = Object.prototype.hasOwnProperty,
+	unfilteredUrl = setUrl( { filter: undefined, module: undefined,
+		moduleId: undefined, testId: undefined } ),
 	defined = {
 		sessionStorage: ( function() {
 			var x = "qunit-test-string";
@@ -113,8 +115,8 @@ function addClass( elem, name ) {
 	}
 }
 
-function toggleClass( elem, name ) {
-	if ( hasClass( elem, name ) ) {
+function toggleClass( elem, name, force ) {
+	if ( typeof force !== "undefined" ? !force : hasClass( elem, name ) ) {
 		removeClass( elem, name );
 	} else {
 		addClass( elem, name );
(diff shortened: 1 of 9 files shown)


reporter/html.js
```diff
@@ -113,8 +115,8 @@ function addClass( elem, name ) {
 	}
 }
 
-function toggleClass( elem, name ) {
-	if ( hasClass( elem, name ) ) {
+function toggleClass( elem, name, force ) {
+	if ( typeof force !== "undefined" ? !force : hasClass( elem, name ) ) {
```
leobalter: this is hard to read

reporter/urlparams.js
```diff
@@ -0,0 +1,96 @@
...
+	return;
+}
+
+var urlParams = getUrlParams();
+
+QUnit.urlParams = urlParams;
+
+// Match module/test by inclusion in an array
+QUnit.config.moduleId = [].concat( urlParams.moduleId || [] );
+QUnit.config.testId = [].concat( urlParams.testId || [] );
```
gibson042: This leaves in place the current sloppiness (privileged parameters are applied immediately and unconditionally, while `urlConfig` parameters wait for runStart and only overwrite `undefined` values), but I think it's fine to save that cleanup for another ticket and commit (cf. https://github.com/jquery/qunit/issues/955#issuecomment-198474631 ).

gibson042:
Any thoughts around how to test this?

reporter/urlparams.js
leobalter: yes. I'm +1 to progressive cleanups rather than holding PRs for too long.

leobalter:
I really like what this PR does, and I'm afraid this might require defining several extra html files to observe the results of using urlParams.

gibson042:
> I'm afraid this might require defining several extra html files to observe the results of using urlParams.

I don't think so... the new file really doesn't have that many responsibilities, and I'm pretty sure they can be tested together:
* parsing of a query string into a `QUnit.urlParams` object
* setting a few predefined properties on `QUnit.config`
* setting user-defined `QUnit.urlConfig` properties on `QUnit.config`

leobalter:
Makes sense.

We already have tests for config when it's set, we only need to check if urlParams affect the config, right?

leobalter:
where config my be QUnit.config, urlConfig, etc

reporter/urlparams.js
```diff
@@ -0,0 +1,96 @@
+( function() {
```
trentmwillis: Once #950 is merged, we can probably drop this closure.

trentmwillis:
Overall, huge :+1: on this. As for the tests, what @gibson042 lists makes sense to me, so long as "setting a few predefined properties on `QUnit.config`" means "setting a few predefined properties on `QUnit.config` **via urlParams**".

Looking forward to the tests.

gibson042:
Tests added! :tada:

trentmwillis:
Tests look good to me. After a rebase I think this will be ready to go.

PR #971, opened 2016-03-24
Core: Export AMD with explicit name.

### LINKED ISSUE #970: Export AMD with name
Right now QUnit exports to AMD without specifying a name. When you transpile your code using Babel, and then try to load it with Almond, it fails because no name has been specified. 

I was able to workaround this issue by manually exporting the module as the following:

```js
define("qunit", function() { return QUnit; });
```

Is it possible to change QUnit's definition to explicitly set the exporting name?

fnando:
Closes #970

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -111,6 +111,7 @@ grunt.initConfig( {
 			"test/logs.html",
 			"test/setTimeout.html",
 			"test/amd.html",
+			"test/amd_almond.html",
 			"test/reporter-html/index.html",
 			"test/reporter-html/legacy-markup.html",
 			"test/reporter-html/no-qunit-element.html",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   ],
   "dependencies": {},
   "devDependencies": {
+    "almond": "^0.3.1",
     "async": "1.5.2",
     "browserstack-runner": "0.4.3",
     "commitplease": "2.3.0",
--- a/src/export.js
+++ b/src/export.js
@@ -64,7 +64,7 @@ if ( typeof exports !== "undefined" && exports ) {
 }
 
 if ( typeof define === "function" && define.amd ) {
-	define( function() {
+	define( "qunit", function() {
 		return QUnit;
 	} );
 	QUnit.config.autostart = false;
--- a/test/amd.js
+++ b/test/amd.js
@@ -1,5 +1,5 @@
 /* global beginData */
-define( [ "qunit" ], function( QUnit ) {
+define( "amd", [ "qunit" ], function( QUnit ) {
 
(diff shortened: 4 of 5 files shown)


trentmwillis:
:+1: this seems good to me.

One question though: is this specific to Almond? It seems like a more general issue of QUnit not being available to import via name without specifying a configuration ([like we have in the current amd test](https://github.com/jquery/qunit/blob/master/test/amd.html#L16)). If that is the case (and I suspect it is), I'd rather this test be something like `amd_named` and just use standard RequireJS so we don't have to add another library to the codebase.

fnando:
@trentmwillis good question. I do not know if other implementations forbid anonymous definition. I know for sure that Almond does.

JamesMGreene:
Almond is only meant to be used with optimized Requires build bundles, so that doesn't seem like a driving factor.

AFAIK, once an AMD module is named, it's consumers must require it by that exact name (no config of anonymous modules allowed), so I'd say this is a breaking change (meaning a major version upgrade).

leobalter:
If it breaks I'll need to hold this before releasing the next major, which I hope I can release it soon.

trentmwillis:
I looked into this more and @JamesMGreene is correct. Once the module is named it must be required by that exact name. The [recommendation from RequireJS](http://requirejs.org/docs/api.html#modulename) is to _not_ code in names:

> You can explicitly name modules yourself, but it makes the modules less portable -- if you move the file to another directory you will need to change the name. It is normally best to avoid coding in a name for the module and just let the optimization tool burn in the module names. The optimization tool needs to add the names so that more than one module can be bundled in a file, to allow for faster loading in the browser.

Due those limitations, I'm now of the opinion we should keep the anonymous module and if support for a named module is needed (as it is here), they should use the [optimization tool recommended by RequireJS](http://requirejs.org/docs/optimization.html).

leobalter:
@jzaefferer @scottgonzalez @jugglinmike you have experience using QUnit with AMD, right? Would you provide some feedback here?

AFAIK, I have two problems and I'll need to choose between solving only one of them.

JamesMGreene:
IIRC, @jrbuke's recommendation was always to just include QUnit on the page as a normal script tag before kicking off the RequireJS main driver.

scottgonzalez:
My understanding is that using an explicit name is always the wrong thing to do.

jrburke:
To confirm, qunit should not name its define. The core issue in #970 is a build process concatenating modules without naming them as part of the concatenation. This is required of a build tool if the built file is to be used by an AMD loader.

trentmwillis:
I'm going to close this as it seems the proper path forward is for consumers to use r.js to do module naming during their build process.

Issue #890, opened 2015-10-30
QUnit.reporter( reporter [, options ] );


leobalter:
After talking to Jörn today, we decided to create this reporter method to call custom reporters for QUnit.

On QUnit 2.0, it's gonna be necessary to call it even for the HTML Reporter.

This should be made on top of the new EventsEmitter interface, so it depends on #882.

To make it work and test it, we need to implement a simple custom reporter, probably a tap output. [`qunit-tap`](https://github.com/twada/qunit-tap) should be observed to avoid extra work.

Other custom reporters should follow `js-reporters` (on QUnit 2+) or published as other npm modules.

---

# Features:

## Reporter name / object

`QUnit.reporter` needs to be called with at least one argument identifying the reporter to be used. It might be a string with the name of the reporter or a reporter object.

TODO: The reporter object is not structured yet

## Options

`QUnit.reporter` might be called with a second argument as an object with property options.

The main options, used as a suggested interface for most reporter are the following, with their respective defaults:

- **log**: `console.log` => the main print output method.
- **error**: `console.error` => the output method for errors.
- **warn**: `console.warn` => the output method for warnings.
- **all**: `null` => if this property is set, it replaces the values for all the other output properties: `log`, `error`, and `warn`.

These print options might not be used no all reporters, as it won't be used on an eventual HTML Reporter built on top of it.

## Examples:

```js
// built-in reporter
QUnit.reporter('tap');

// built-in reporter with options
QUnit.reporter('tap', {
  all: print
});

// custom reporter
QUnit.reporter(require('js-reporters-junit'));
```

# The output on the `runEnd` event

The `runEnd` event might include a custom output property containing the printed results. This is optional on different reporters, but it can be used with different reporters, where `output` is an object and their children keys are the names of the reporters, their values should be strings containing the output result.

## Example:

```js
// do whatever you want with the reporter results
QUnit.on('runEnd', function(details) {
  console.log(details.output.junit);
  console.log(details.output.tap);
  fs.writeFileSync(details.output.tap);
});
```

This part might be made in a separate issue and it does not need to block the 2.0 Milestone.

jzaefferer:
> On QUnit 2.0, it's gonna be necessary to call it even for the HTML Reporter.

That wasn't my understanding. Unless there's a good reason to require that, I'd stick with the default.

> QUnit.reporter('tap');

I'm not sure this is a good idea either. Even if QUnit bundles reporters, it should still be easy enough to get an instance and pass it to the reporter method.
[...]

PR #972, opened 2016-03-24
Tap reporter (prototype yet - please come and discuss)

leobalter:
This is my first prototype for a QUnit's native tap reporter. Following the specs for TAP version 13.

The tap output allows QUnit to be integrated anywhere that reads tap. With this prototype I would like to propose a tap adapter for js-reporter so it could (and should) read any tap result, and potentially validate.

This reporter is browser compatible and can be used anywhere with a console.log or by setting a custom print method.

`QUnit.reporter( "tap" )` needs to be called to activate this reporter, this way allows extensiblity without polluting the namespace with other methods.

### DIFF
--- a/reporter/tap.js
+++ b/reporter/tap.js
@@ -1,0 +1,77 @@
+(function() {
+
+function reporter( rep ) {
+	if ( rep === "tap" ) {
+		tapReporter();
+	}
+};
+
+reporter.print = function( val ) {
+	console.log( val );
+};
+
+QUnit.reporter = reporter;
+
+function tapReporter() {
+	var testCount = 0;
+	var failed = 0;
+
+	QUnit.begin( function() {
+		QUnit.reporter.print( "TAP version 13" );
+	} );
+
+	QUnit.done( function( details ) {
+		var output = [
+			"",
+			"1.." + testCount,
+			"# tests " + testCount,
+			"# pass " + ( testCount - failed )
+		];
+
+		QUnit.reporter.print( output.join( "\n" ) );
+	} );
+
+	QUnit.moduleStart( function( details ) {
+		QUnit.reporter.print( "# module: " + details.name );
+	} );
+
+	QUnit.testDone( function( details ) {
+		var assertion;
+		var failedMessage;
+		var output;
+		var i = 0;
+
+		testCount++;
+
+		output = "ok " + testCount + " - ";
+
+		if ( details.skipped ) {
+			output += "# SKIP " + details.name;
+		} else if ( details.failed == 0 ) {
+			output += details.name;
+		} else {
+			failed++;
+			output = "not " + output + details.name;
+
+			for ( i = 0; i < details.assertions.length; i++ ) {
+				assertion = details.assertions[ i ];
+				if ( assertion.result ) {
+					continue;
+				}
+				failedMessage = [
+					"",
+					"  ---",
+					"  message: '" + assertion.message + "'",
+					"  severity: fail",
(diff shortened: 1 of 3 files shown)


trentmwillis:
Finally got around to looking at this. Overall, I like the idea. I've seen plenty of node-based shims to get reporter output so providing a default would be nice. A couple other questions:

- How would one register another reporter to use? Seems like we should define that interface so plugin authors have a standard to follow.
- Would this be the only reporter we maintain other than the HTML reporter? Similarly, how would this play with the HTML reporter? Should we support running both at the same time and will we move the HTML reporter to be initialized in a similar fashion (e.g., `QUnit.reporter('html')`)?

reporter/tap.js
```diff
@@ -0,0 +1,77 @@
+(function() {
+
+function reporter( rep ) {
+	if ( rep === "tap" ) {
+		tapReporter();
+	}
+};
+
+reporter.print = function( val ) {
+	console.log( val );
```
trentmwillis: Should we make this a bit more generic? Like:

```js
console.log.apply(console, arguments);
```

I guess this depends on the circumstances under which `print` will be used.

reporter/tap.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+QUnit.reporter = reporter;
```
gibson042: This _definitely_ doesn't belong in reporter/tap.js, and I [still](https://github.com/jquery/qunit/issues/890#issuecomment-152739427) have issues with `QUnit.reporter( string )`.

reporter/tap.js
```diff
@@ -0,0 +1,77 @@
...
+
+function tapReporter() {
+	var testCount = 0;
+	var failed = 0;
+
+	QUnit.begin( function() {
+		QUnit.reporter.print( "TAP version 13" );
```
gibson042: This also feels a bit too global.

gibson042:
I love the results, but I think the interface needs some tweaking.

trentmwillis:
Maybe this is thinking too broadly, but what if we defined an interface that reporter's then implemented? Consumers can register reporters by simply passing in an object that meets our defined interface (e.g., `QUnit.setReporter(reporterInstance)`) and we will invoke methods at specific points (e.g., during `begin`, `testStart`, etc.).

I think an ideal solution would be interoperable with other platforms (e.g., mocha).

gibson042:
I agree, and think that's https://github.com/js-reporters/js-reporters/issues/3 .

reporter/tap.js
leobalter: > This definitely doesn't belong in reporter/tap.js

definitely, I agree.

I'll review this, crossing with the efforts on the js-reporter and provide some good rebased thing.

reporter/tap.js
leobalter: I wanted something I can patch/replace/inspect. Do you have any other suggestion?

gibson042: It should be specific to the TAP reporter, or even better to a TAP reporter _instance_. Testing code code then boil down to something like `QUnit.reporter( TAPReporterFactory({ print: captureOutput }) )` (modulo construction/configuration/registration bikeshedding).

trentmwillis:
@leobalter will this be superseded by the work happening in https://github.com/qunitjs/qunit/pull/1026? It seems like we should use the js-reporters reporter combined with the new event emitter interface.

leobalter:
this would be fair. at least we should give a try using js-reporters.

Issue #953, opened 2016-03-15
Update UI to allow multiple test/module selection


trentmwillis:
As discussed in #951 and #952, it would be beneficial to have a better interface that allows for filtering multiple modules or tests. While this is nice to support those particular features, it would also be beneficial to take this as a chance to address some pain points of working with large test suites. In general, having a dropdown that is thousands of items long is not very useful.

In general, the following features should be considered:

- [x] select individual modules
- [x] select multiple modules
- [ ] select multiple tests
- [x] support grouping of modules (e.g., in Ember tests we have Acceptance, Integration, or Unit tests) [can be accomplished through nested modules]
- [x] filter/search the list of tests/modules

Anything else?

mlato-yahoo:
Based on those requirements I've drafted this simple multi-selector.

![multi-select-mock](https://cloud.githubusercontent.com/assets/15200965/13832492/e6cd9f62-eb99-11e5-8d3c-cae53c3a48d1.png)

Features:
* Multi-select window with checkboxes and scrolling.
* Search box for filtering by module or test name.
* "Submit" button to run selected tests or modules.

Other things that we need to consider:
* How do we tag tests/modules as part of groups? (Acceptance, Integration, etc.)
* Select all/clear all functionality
* Would a module multi-selector replace the module dropdown, and where would it be in the UI?
* **Idea: What about replacing the module select dropdown with a dropdown that opens up into this multi-selector?

Let's make this happen!

trentmwillis:
I think this is a good start.

> How do we tag tests/modules as part of groups? (Acceptance, Integration, etc.)

I have three ideas on how to solve this:

1. Enforce some naming convention (e.g., `<group-name> | <module-name>`)
2. Set a configurable "separator" that then denotes when you're entering a new group (e.g., `QUnit.config.groupSeparator = '|';`). This could make it easier to support nested groups if needed.
3. Introduce metadata into a `module` that would make grouping declarative.

I personally lean towards either 2 or 3.

> Select all/clear all functionality

I believe we could accomplish this similarly to how the current module dropdown works. Reserve the top option as a "clear all" option. This works since selecting all modules is essentially the same as selecting none of the modules.

> Would a module multi-selector replace the module dropdown...

Yes. I think having both would cause too much UI clutter.

> ...and where would it be in the UI?

I like your idea: replace "the module select dropdown with a dropdown that opens up into this multi-selector"

gibson042:
[...]

PR #973, opened 2016-03-24
HTML Reporter: Multi-select module dropdown replacing module selector

mlato-yahoo:
A multi-select module dropdown, in response to [#953](https://github.com/jquery/qunit/issues/953)

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -92,6 +92,30 @@ function addEvent( elem, type, fn ) {
 }
 
 /**
+ * @param {HTMLElement} elem
+ * @param {string} type
+ * @param {Function} fn
+ */
+function removeEvent( elem, type, fn ) {
+	if ( elem.removeEventListener ) {
+
+		// Standards-based browsers
+		elem.removeEventListener( type, fn, false );
+	} else if ( elem.detachEvent ) {
+
+		// Support: IE <9
+		elem.detachEvent( "on" + type, function() {
+			var event = window.event;
+			if ( !event.target ) {
+				event.target = event.srcElement || document;
+			}
+
+			fn.call( elem, event );
+		} );
+	}
+}
+
+/**
  * @param {Array|NodeList} elems
  * @param {string} type
  * @param {Function} fn
@@ -260,20 +284,23 @@ function setUrl( params ) {
 }
 
 function applyUrlParams() {
-	var selectedModule,
-		modulesList = id( "qunit-modulefilter" ),
+	var i,
+		selectedModules = [],
+		modulesList = id( "qunit-modulefilter-dropdown-list" ).getElementsByTagName( "input" ),
 		filter = id( "qunit-filter-input" ).value;
 
-	selectedModule = modulesList ?
-		decodeURIComponent( modulesList.options[ modulesList.selectedIndex ].value ) :
-		undefined;
+	for ( i = 0; i < modulesList.length; i++ )  {
+		if ( modulesList[ i ].checked ) {
+			selectedModules.push( modulesList[ i ].getAttribute( "module-id" ) );
+		}
+	}
 
 	window.location = setUrl( {
-		module: ( selectedModule === "" ) ? undefined : selectedModule,
 		filter: ( filter === "" ) ? undefined : filter,
+		moduleId: ( selectedModules.length === 0 ) ? undefined : selectedModules,
 
-		// Remove moduleId and testId filters
-		moduleId: undefined,
+		// Remove module and testId filter
+		module: undefined,
 		testId: undefined
 	} );
 }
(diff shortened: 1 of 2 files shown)


trentmwillis:
Excited for this! I'll try to take a look sometime this weekend and mess around with it locally.

Arkni:
Nice, I'll try to test it locally and provide feedback :)

reporter/html.js
```diff
@@ -260,23 +284,23 @@ function setUrl( params ) {
...
 		filter = id( "qunit-filter-input" ).value;
 
-	selectedModule = modulesList ?
-		decodeURIComponent( modulesList.options[ modulesList.selectedIndex ].value ) :
-		undefined;
+	for ( i = 0; i < modulesList.length; i++ )  {
+		if ( modulesList[ i ].checked ) {
+			selectedModules.push( modulesList[ i ].id );
+		}
+	}
```
gibson042: These changes should wait for gh-969 to land, and then rebase on top of it.

reporter/html.js
```diff
@@ -329,44 +353,106 @@ function toolbarLooseFilter() {
...
 	}
 
-	moduleFilter.setAttribute( "id", "qunit-modulefilter-container" );
-	moduleFilter.innerHTML = moduleFilterHtml;
+	moduleSearch.placeholder = "Select modules";
 
-	addEvent( moduleFilter.lastChild, "change", applyUrlParams );
+	label.innerHTML = "Module: ";
+
+	clearFilter.innerHTML = "< All modules ><hr/>";
```
gibson042: The `<` and `>` should be expressed as HTML entities, for safety.

reporter/html.js
```diff
@@ -329,44 +353,106 @@ function toolbarLooseFilter() {
...
+				if ( moduleText.indexOf( searchText ) > -1 ) {
+					listItems[ i ].style.display = "";
+				} else {
+					listItems[ i ].style.display = "none";
+				}
+			}
+		}
+	} );
 
+	dropDownList.insertAdjacentHTML( "beforeend", dropDownHtml );
```
gibson042: I think that's a bit too modern for our support range, specifically with respect to Safari and Android.

trentmwillis: [Browser support looks pretty solid](http://caniuse.com/#feat=insertadjacenthtml), especially if we're targeting a 2.0 release.

trentmwillis:
One general comment, take a look at the [jQuery CSS Style Guide](https://contribute.jquery.org/style-guide/css/), you have some minor issues in the CSS. Such as units for 0 values and non-hex color values.

reporter/html.js
```diff
@@ -329,44 +353,106 @@ function toolbarLooseFilter() {
...
+		clearFilter = document.createElement( "li" );
 
-	if ( !toolbar || !moduleFilterHtml ) {
+	if ( !toolbar || !dropDownHtml ) {
 		return false;
 	}
 
-	moduleFilter.setAttribute( "id", "qunit-modulefilter-container" );
-	moduleFilter.innerHTML = moduleFilterHtml;
+	moduleSearch.placeholder = "Select modules";
```
trentmwillis: `placeholder` is supported in our current browser policy. Another notch for waiting for 2.0.

reporter/html.js
trentmwillis: Better yet, drop the `<`, `>`, and `<hr/>` altogether. They seem to only be there for styling, which should be done via CSS instead.

reporter/html.js
```diff
@@ -329,44 +353,106 @@ function toolbarLooseFilter() {
...
+	dropDownContainer.style.display = "none";
+
+	dropDownList.appendChild( clearFilter );
+	moduleFilter.appendChild( label );
+	moduleFilter.appendChild( moduleSearch );
+	moduleFilter.appendChild( dropDownContainer ) ;
+
+	// Enables show/hide for the dropdown
+	addEvent( moduleSearch, "focus", function() {
+		var dropDownList = id( "qunit-modulefilter-dropdown-container" );
```
trentmwillis: Throughout this callback and the next one below, is there a reason to not use the `dropDownContainer` reference you already have in scope?

Arkni:
Hi @mlato-yahoo 

I tested your code locally and noticed that when using `module` in conjunction with `moduleId` non is run.
Note that this behaviour only happens when `moduleId` is not the same as the hash of `module`.

leobalter:
> I tested your code locally and noticed that when using module in conjunction with moduleId non is run.
Note that this behaviour only happens when moduleId is not the same as the hash of module.

This is a fair behavior.

The dropdown selector could replace the `module` param by a `moduleId` in case it gets any new selection.

Arkni:
>> I tested your code locally and noticed that when using module in conjunction with moduleId non is run.
>> Note that this behaviour only happens when moduleId is not the same as the hash of module.

> This is a fair behavior.

>The dropdown selector could replace the module param by a moduleId in case it gets any new selection.

Yeah, I forgot that any test passing all filters (`moduleId`, `testId`, `module`, `filter`) gets to run.

I was thinking that, it should run one module if `moduleId` is the same as the hash of `module`, and run the two, otherwise.

But the current behavior is OK. and seems legit.
Thanks :)

reporter/html.js
stefanpenner: this should just progressively fall back to not doing anything though right?

leobalter: Next release will be 2.0, so that won't be an issue

stefanpenner:
screenshot: https://cloud.githubusercontent.com/assets/15200965/14031788/bea08d82-f1cb-11e5-8222-97285144ef3f.png

reporter/html.js
```diff
@@ -260,20 +284,22 @@ function setUrl( params ) {
...
+	}
 
 	window.location = setUrl( {
-		module: ( selectedModule === "" ) ? undefined : selectedModule,
 		filter: ( filter === "" ) ? undefined : filter,
+		moduleId: ( selectedModules.length === 0 ) ? undefined : selectedModules,
 
-		// Remove moduleId and testId filters
-		moduleId: undefined,
+		// Remove testId filter
```
gibson042: The old code set `module` and `filter`, and cleared `moduleId` and `testId`. If we're now setting `moduleId` instead of `module`, we should clear `module`.

reporter/html.js
```diff
@@ -327,44 +353,106 @@ function toolbarLooseFilter() {
...
-		">< All Modules ></option>";
-
-	for ( i = 0; i < modulesList.length; i++ ) {
-		moduleFilterHtml += "<option value='" +
-			escapeText( encodeURIComponent( modulesList[ i ] ) ) + "' " +
-			( QUnit.urlParams.module === modulesList[ i ] ? "selected='selected'" : "" ) +
-			">" + escapeText( modulesList[ i ] ) + "</option>";
+	for ( i = 0; i < config.modules.length; i++ ) {
+		if ( config.modules[ i ].name !== "" ) {
+			dropDownHtml += "<li><input id='" + config.modules[ i ].moduleId +
```
gibson042: `name` would be a more appropriate attribute than `id`.

reporter/html.js
```diff
@@ -327,44 +353,106 @@ function toolbarLooseFilter() {
...
 		return false;
 	}
 
-	moduleFilter.setAttribute( "id", "qunit-modulefilter-container" );
-	moduleFilter.innerHTML = moduleFilterHtml;
+	moduleSearch.placeholder = "Select modules";
+
+	label.innerHTML = "Module: ";
+
+	clearFilter.innerHTML = "< All modules ><hr/>";
```
gibson042: I'm with @leobalter on removing the `<` and `>`.

reporter/html.js
gibson042: It won't work on Android<4.3 or iOS<8.4. Even [jQuery 3.0 supports](https://github.com/jquery/jquery.com/blob/jquery-3/pages/browser-support.html#L18-L22) Android 4.0+ and iOS 7+.

reporter/html.js
mlato-yahoo: I need to keep `id` so that clicking the label selects the checkbox.

gibson042: That's also possible if the input is inside the label, no `id` needed. But if you do keep id, it should be prefixed with something like "module-".

mlato-yahoo: Ahh, putting the input inside of the label worked.

mlato-yahoo:
I've fixed the major CSS issues and changed most units to use `em` instead of `px`.

mlato-yahoo:
The sizing should also be fixed - menus align properly on all browsers.

leobalter:
This looks good to go, but it doesn't need to come along with 2.0.0 or this means it doesn't bring any breaking change.

trentmwillis:
> it doesn't bring any breaking change, am I right?

I guess that would depend on if we consider the DOM public API. If anyone has extended QUnit in a way that makes assumptions about the module filter's DOM, then they would break.

I wouldn't mind seeing this implemented in 2.0.

leobalter:
2.0 changes beyond deprecated methods should be minimal to avoid complex migrations. But that doesn't seem to go with 2.0 and cause problems.

@gibson042 @jzaefferer

jzaefferer:
We've made plenty of changes to the HTML reporter over the various 1.x releases, seems fine to land this in any release.

trentmwillis:
> We've made plenty of changes to the HTML reporter over the various 1.x releases

Well that answers my question then. I'm good with this landing whenever it is ready then; let's not block 2.0 on it.

reporter/html.js
```diff
@@ -92,6 +92,30 @@ function addEvent( elem, type, fn ) {
...
+
+		// Support: IE <9
+		elem.detachEvent( "on" + type, function() {
+			var event = window.event;
+			if ( !event.target ) {
+				event.target = event.srcElement || document;
+			}
+
+			fn.call( elem, event );
+		} );
```
leobalter: I'm removing this part

leobalter:
merged at 0780127, thanks!

Krinkle:
Before and after screenshots, for historical reference. Note that the below also includes https://github.com/qunitjs/qunit/pull/989 (Richard Gibson).

* [QUnit 1.23.1 demo](https://codepen.io/Krinkle/full/QwLZWWe)
* [QUnit 2.0.0 demo](https://codepen.io/Krinkle/full/mybzddj)

Screenshot
Screenshot

PR #976, opened 2016-03-25
All: Remove deprecated features

leobalter:
- globals
- assertions on QUnit namespace
- QUnit.stop
- QUnit.init
- setup/teardown
- asyncTest
- test with expected and async parameters
- old style for logging callbacks
- duration property on testDone details

This is the first batch. It can be improved and QUnit.reset is not addressed here yet.


src/core.js
```diff
@@ -77,8 +66,19 @@ extend( QUnit, {
 
 	},
 
-	// DEPRECATED: QUnit.asyncTest() will be removed in QUnit 2.0.
-	asyncTest: asyncTest,
+	asyncTest: function() {
+		throw new Error(
+			"asyntTest is now removed from QUnit 2.0, check out our upgrade guide at " +
```
gibson042: s/asyntTest/asyncTest/. And as an aside, I notice that only some of the removals come with `throw`s... when will we fully complete the process and drop even these?

src/core.js
```diff
@@ -86,8 +86,6 @@ extend( QUnit, {
 
 	only: only,
 
-	// DEPRECATED: The functionality of QUnit.start() will be altered in QUnit 2.0.
-	// In QUnit 2.0, invoking it will ONLY affect the `QUnit.config.autostart` blocking behavior.
```
gibson042: Did you forget to remove the block below?

src/export.js
```diff
@@ -1,18 +1,18 @@
-// Deprecated
-// Extend assert methods to QUnit for Backwards compatibility
-( function() {
-	var i,
-		assertions = Assert.prototype;
+function applyDeprecated( name ) {
+	return function() {
+		throw new Error(
+			name + " is now removed from QUnit 2.0, check out our upgrade guide at " +
```
gibson042: I think all of the new errors should identify context (e.g., global {name} vs. QUnit.{name}), but especially here where it's not obvious.

src/core.js
leobalter: nope, actually the current code is dependent on the whole method. Until I do a coverage check or minimize the complexity, I'm leaving it as it is. It won't have `QUnit.stop` to work along, at least.

src/export.js
leobalter: I forgot to improve this message.

src/core.js
leobalter: I've set throws only for removed methods as asyncTest, init, stop, and the previously exposed globals.

I skipped throwing for features as setup/teardown, test signature, etc.

leobalter:
@gibson042 is there anything I should change?

@jzaefferer would you take a look on this one?

src/export.js
```diff
@@ -1,18 +1,19 @@
...
 
-	function applyCurrent( current ) {
-		return function() {
-			var assert = new Assert( QUnit.config.current );
-			current.apply( assert, arguments );
-		};
-	}
+( function() {
+	var i;
+	var assertions = Assert.prototype;
```
trentmwillis: Should use a single `var`

trentmwillis:
Overall :+1:. Still need to get that typo [mentioned here](https://github.com/jquery/qunit/pull/976#discussion_r57505231).

src/export.js
leobalter: TL;DR: yes.

that's a mess I'm making after changes that happened on our [style guide](http://contribute.jquery.org/style-guide/js). It was a suggesting a single var and then it suggested 1 var per line, stacking non initialized variables on the same line. Now there's nothing and I'm confused.

At least on of the examples the style guide is using a single var, so I'll apply the change you suggested.

trentmwillis: Okay. I personally prefer 1 `var` per line, but everywhere else in QUnit we use a single `var`.

leobalter: done! I fixed this on a rebased commit, my bad I amended to a fixup commit.

leobalter:
I'll wait for @jzaefferer to take a look on this, as he mentioned he will do it tomorrow, or today on his timezone. :)

After it's merged I'll proceed merging the other done PRs with breaking changes.

I'll probably try a pre-release so we can all test it on different suites.

src/core.js
```diff
@@ -18,17 +18,6 @@ extend( QUnit, {
 			}
 		}
 
-		// DEPRECATED: handles setup/teardown functions,
-		// beforeEach and afterEach should be used instead
-		if ( testEnvironment && testEnvironment.setup ) {
-			testEnvironment.beforeEach = testEnvironment.setup;
```
jzaefferer: This should throw an error as well, otherwise the `setup` and `teardown` properties are simply ignored, breaking tests in unexpected ways.

src/core.js
jzaefferer: This should mention `QUnit.test() with assert.async()` as replacement:

```
"asyncTest is removed in QUnit 2.0, use QUnit.test() with assert.async() instead. Details in our upgrade guide at " +
```

jzaefferer: Then apply that to any other replacements as well.

src/core.js
jzaefferer: Does it throw an error when called inside of a test context?

src/core.js
```diff
@@ -143,20 +141,6 @@ extend( QUnit, {
 		resumeProcessing();
 	},
 
-	// DEPRECATED: QUnit.stop() will be removed in QUnit 2.0.
-	stop: function( count ) {
```
jzaefferer: Shouldn't this stay, but throw an error?

test/deprecated.js
```diff
@@ -0,0 +1,88 @@
...
+			"notDeepEqual",
+			"strictEqual",
+			"notStrictEqual",
+			"throws",
+			"raises"
+		];
+
+		assertions.forEach( function( item ) {
+			assert.throws( function() {
+				item();
```
jzaefferer: This looks wrong. Calling `"expect"()` doesn't make sense, right? Could be `window[ item ]()` though.

src/core.js
leobalter: 9756e69

src/core.js
leobalter: not yet, something that wasn't so simple to remove. I need to rework the "promiseable tests"  api. I'm working to get it done.

leobalter: 5bc1982 I got it working, but I still need to optimize this.

src/core.js
leobalter: We can't throw, otherwise removing it on 2.1 would be a breaking change.

We may log a warning, `console.warn` is available on the set of supported browsers for 2.0

src/core.js
leobalter: fixed at 70b9d70

src/export.js
```diff
@@ -1,51 +1,42 @@
-// Deprecated
-// Extend assert methods to QUnit for Backwards compatibility
-( function() {
-	var i,
-		assertions = Assert.prototype;
+function applyDeprecated( name ) {
+	return function() {
+		throw new Error(
+			"`" + name + "` is removed in QUnit 2.0, you should now use the " +
```
gibson042: Let's drop the backticks; this isn't a great error message:
> \`The global ok\` is removed…

src/export.js
```diff
@@ -1,51 +1,42 @@
...
-	}
+Object.keys( Assert.prototype ).forEach( function( key ) {
+	QUnit[ key ] = applyDeprecated( "QUnit." + key );
+} );
 
-	for ( i in assertions ) {
-		QUnit[ i ] = applyCurrent( assertions[ i ] );
-	}
-}() );
+QUnit.stop = applyDeprecated( "QUnit.stop" );
```
gibson042: `stop` should be handled separately, since there is no "correct namespace".

src/export.js
leobalter: ef9c175 what do you think? Also the global message is `"The global `ok` is removed..."`

src/export.js
leobalter: 97e7e49 fixes this.

leobalter:
@jquery/qunit I believe I addressed all the reviews so far, how is it now?

src/test.js
```diff
@@ -623,25 +604,61 @@ function skip( testName ) {
...
+	pauseProcessing();
+}
+
+function internalStart() {
+
+	// If a test is running, adjust its semaphore
+	config.current.semaphore -= 1;
+
+	// If semaphore is non-numeric, throw error
+	if ( isNaN( config.current.semaphore ) ) {
```
gibson042: This is probably unnecessary now that the function is internal.

leobalter: semaphore is still exposed. Can we remove it securely? I'll check this.

gibson042:
There seems to be a little cruft from arguments handling, but I'm satisfied overall. LGTM.

src/test.js
leobalter: If so, I can do this (es5 + ie9+ support ftw):

```js
Object.defineProperty(config.current, "semaphore", { set: function() { throw ... } ));
```

gibson042: `semaphore` is not documented, so anyone using it directly is already coloring outside the lines.

leobalter: 21c254a I tried this refactoring, it's working and it avoids using the config.current using test directly whenever is possible. I'll do a inline comment to show the point I'm not satisfied

leobalter: I cannot do any inline comments because it's not close from a change I did:

You can see [here](https://github.com/jquery/qunit/pull/976/commits/21c254ac2f65b4833d80eee055aa47141c737045#diff-f8f3ba9b60ca5c3a0a1dd233f0d34122R47) I have this `resumeProcessing()` that can't be called with the current test where `done` was set in. Doing it breaks the current suite.

We can investigate this in another issue, but for me it looks we are leaking the async functionality somehow and our current tests are returning false positives.

src/test.js
```diff
@@ -338,10 +336,9 @@ Test.prototype = {
 		if ( promise != null ) {
 			then = promise.then;
 			if ( QUnit.objectType( then ) === "function" ) {
-				QUnit.stop();
-				then.call(
-					promise,
-					function() { QUnit.start(); },
+				internalStop( test );
+				promise.then(
```
gibson042: Believe it or not, it's important to keep the `then.call`: https://promisesaplus.com/#point-75

src/core.js
jzaefferer: Alright

jzaefferer:
This is looking good to me. I also tested it against https://github.com/jquery/jquery-ui/pull/1577 - didn't notice any issues with QUnit itself, but both https://github.com/JamesMGreene/qunit-composite and https://github.com/JamesMGreene/qunit-assert-close need updates to be 2.0 compatible.

src/test.js
leobalter: 195b234 lol

leobalter:
I believe it's good to go now.

jzaefferer:
👍

Issue #978, opened 2016-03-26
Make module-contextual module/test/etc. functions


gibson042:
(this is a restatement of https://github.com/jquery/qunit/pull/670#discussion_r26321413 , which itself referenced a discussion from https://github.com/jquery/qunit/pull/670#issuecomment-78513676 )

7be1d6c38cac4370bafdb17b10f0f4dee576bd12 introduced the `QUnit.module( name, callback )` signature, which immediately invokes a callback to allow for definition of nested modules. However, doing so requires using global `QUnit.module` and/or `QUnit.test` functions: https://jsfiddle.net/k9af94x7/

This is ugly, and bad for the same reasons as use of global assertions like `QUnit.equal` instead of contextual assertions like `assert.equal`. We're [fixing the assertions issue](https://github.com/jquery/qunit/pull/976), and should do the same with `module` and `test` by making them accessible from either context or arguments of the module callback, while fulfilling some goals to the maximum possible extent:
* Intuitively-readable invocations, possibly paradigm-dependent (e.g., use of BDD names)
* Maximum similarity between `module` and `test` callbacks, for future iterative convergence
* Separate concerns (e.g., no properties on `assert` that do not pertain to assertions, since we might want to allow integrating external assertion libraries)

My ideal vision would fully integrate the two functions, allowing code like:
```js
// Note the second parameter
QUnit.test("grandparent", (assert, test) => {
	// `test` is destructurable, and contains helpers like beforeEach.
	// Those that are functions have signatures analogous to `test` itself.
	// Also present is environment, so users never need deal with `this`.
	test.beforeEach((assert, test) => { test.environment.ready = true; });

	// Any test can contain assertions, even those that would traditionally be modules.
	// But only child tests interact with beforeEach/afterEach.
	assert.strictEqual(test.environment.ready, undefined);
	test("uncle", (assert, { environment }) => {
		assert.strictEqual(environment.ready, true);
	});

	// As noted in https://github.com/jquery/qunit/pull/670#issuecomment-78513676 ,
	// a significant current difference between `module` and `test` is their
	// (a)synchronicity.
	// We can make that explicit (names subject to bikeshedding, but I think
	// defaulting to async will make for the smoothest transition.
	test.immediate("parent", (assert, test) => {
		test("child", (assert, test) => {
			assert.strictEqual(test.environment.ready, true)
		});
	});
});
```

However, if the above is a bit much, then I suppose we can just add `module` and `test` properties to the argument passed to module callbacks:
```js
[...]

PR #979, opened 2016-04-02
Core: Add `test` method to `hooks` object.

jonnyreeves:
Provides an alias for `QUnit.test` to allow a more natural, nested nested api; this allows the author to write their tests as follows:

```js
QUnit.module('action/example', t => {
    t.beforeEach(/* ... */);
    t.test('it should be awesome', assert => {
    	assert.ok(true, 'rock and roll!');
    }
})
```

The downside I can see to this change is that it moves the existing convention away from the `t` / `hooks` argument being solely based around adding module hooks, however I feel there is a real benefit from not having to constantly write `QUnit` everywhere :grin: 

Thanks!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -33,7 +33,8 @@ extend( QUnit, {
 
 		moduleFns = {
 			beforeEach: setHook( module, "beforeEach" ),
-			afterEach: setHook( module, "afterEach" )
+			afterEach: setHook( module, "afterEach" ),
+			test: QUnit.test
 		};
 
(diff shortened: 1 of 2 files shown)


jonnyreeves:
Ah sorry, you are already discussing this feature in #978.  Thanks again for your hard work on this awesome library.

leobalter:
thank you, @jonnyreeves!

Hooking the test on the callback arguments is the easiest part, the complexity comes registering the children tests as part of their parents. That's the biggest challenge at #978, I assume.

PR #981, opened 2016-04-13
Events: EventEmitter logging system

cynicaldevil:
Builds upon the EventEmitter model PR ( #882 ) . Adds event details as specified in the js-reporters draft proposal. Two test files( `test/event.js` and `test/logs.js` ) have been modified so that all assertions within them pass.

### DIFF
--- a/test/logs.js
+++ b/test/logs.js
@@ -9,42 +9,50 @@ var totalTests, moduleContext, moduleDoneContext, testContext, testDoneContext, 
 	log = 0,
 	module1Context = {
 		name: "logs1",
+		childSuites: [],
 		tests: [
-			(module1Test1 = {
-				"name": "test1",
+			( module1Test1 = {
+				"suiteName": "logs1",
+				"testName": "test1",
 				"testId": "646e9e25"
 			}),
-			(module1Test2 = {
-				"name": "test2",
+			( module1Test2 = {
+				"suiteName": "logs1",
+				"testName": "test2",
 				"testId": "646e9e26"
 			})
 		]
 	},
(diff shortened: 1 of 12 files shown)


test/logs.js
```diff
@@ -272,12 +285,14 @@ QUnit.test( module2Test4.name, function( assert ) {
 
 	assert.deepEqual( testDoneContext, {
 		assertions: [],
-		module: module2Context.name,
-		name: module2Test3.name,
+		suiteName: module2Context.name,
+		testName: module2Test3.testName,
+		status: "skipped",
```
leobalter: this is a big issue for implementing the events emitter: we can't simply remove the other properties, that would represent a breaking change and would require a major update. 

In our api development process, we want to give a reasonable time before any major update to allow users to migrate from the old to the new API, otherwise upgrading a major version would be very hard in many suites.

I need to check this EventEmiter api and maybe releasing it with the new properties coming on a follow-up PR might be easier. That's also gonna need to rebased after some changes happening right now at the current code.

jzaefferer: Since it looks like the new event emitter API won't make it into 2.0.0, can we release the new API somewhere in a 2.x release, while keeping the existing callbacks intact? We could then remove the old callbacks in 3.0.

New events with old event data (or both old + new data) seems like a bad idea, so I think it would be better to add the new API with new data, then eventually remove the old API....

leobalter:
@cynicaldevil I'll continue this patch on #882, to keep the discussion there. Also, as discussed above, the event details will probably come separately onto another PR.

PR #985, opened 2016-04-18
More deprecations pre-2.0

### LINKED ISSUE #453: Remove QUnit.reset from the public API
To be deprecated (issue #354).

### LINKED ISSUE #604: Remove the string comparison signature from the `throws` assertion
In `v2.0`, we should remove the string comparison signature from the `throws` assertion type.  It is an awkward signature because it _requires_ that there is also a message string provided in order for the assertion to work correctly.

### LINKED ISSUE #660: Investigate removing QUnit.config.autorun
From a cursory glance through the codebase, I'm pretty sure that `QUnit.config.autorun` really isn't necessary if we make some other updates.

It is also not officially documented, though you can definitely find a few [blog posts](https://twoguysarguing.wordpress.com/2010/11/02/make-javascript-tests-part-of-your-build-qunit-rhino/) about people leveraging it to make their QUnit tests work on Rhino. As such, I'd suggest we leave it in till `v2.x`.

### LINKED ISSUE #974: Stop defining a default QUnit.reset
Since we're deprecating `QUnit.reset`, we should stop [relying on it](https://github.com/jquery/qunit/blob/f13038f8530da11a4c3793f35ffe4d4e1d0531a2/src/test.js#L228-L231). We could replicate the behavior almost perfectly by automatically registering a `testDone` callback that includes the [current logic](https://github.com/jquery/qunit/blob/f13038f8530da11a4c3793f35ffe4d4e1d0531a2/src/test.js#L448-L459), but executes it only when `QUnit.reset === undefined` (or even `!hasOwn.call(QUnit, "reset")`).

I say "almost perfectly" because the current code runs after all `testDone` callbacks, but this proposal would have it run _before_ them. This seems acceptable to me, but if it's a problem we can explore a promise-based remedy.

leobalter:
- QUnit.reset
- QUnit.config.autorun
- Assert#throws
- extra cleanups

Closes #453 
Closes #974 - @gibson042, would you confirm this one, please?
Closes #660 
Closes #604

### DIFF
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -47,11 +47,12 @@ module.exports = function( grunt ) {
 
 		registerEvents( QUnit, file, runEnd );
 
-		QUnit.config.autorun = false;
+		QUnit.config.autostart = false;
 
 		require( "../../" + file );
 
 		QUnit.load();
+		QUnit.start();
 	}
 
 	function registerEvents( QUnit, file, runEnd ) {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -30,9 +30,7 @@ var config = QUnit.config,
 	},
 	modulesList = [];
 
-/**
-* Escape text for attribute or text content.
-*/
+// Escape text for attribute or text content.
 function escapeText( s ) {
 	if ( !s ) {
 		return "";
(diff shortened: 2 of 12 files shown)


jzaefferer:
I looked through all commits one by one, looks good to me.

leobalter:
Thanks, @jzaefferer!

rc is coming!

Issue #988, opened 2016-04-19
Module picker is broken


gibson042:
The multi-select picker introduced in 0780127c7f64eed14596afb0b3873b0dc6a389ef provides no mechanism for committing a selection, and suffers some on styling and accessibility. I'd like to fix it before 2.0.

trentmwillis:
Styling seemed okay to me, but I did note some other issues as you did as well:

* Accessibility could be improved
* Shouldn't have to hit "go" in order to commit a selection (not intuitive)
* After closing the dropdown, there should be some indication of what has been selected
* After closing the dropdown, the previous search/filter should clear
* There should be some indication that the thing is a dropdown and not just a text input
* The labels for the options should be block-level so that users can click anywhere in the row to select an item
* Since "All modules" doesn't switch all the checkboxes to "checked", it should probably say "Clear all" or something similar

PR #989, opened 2016-04-20
HTML Reporter: Improve toolbar styles & accessibility

gibson042:
Fixes gh-988
Ref 0780127

Updates styles, enlarges interaction targets (i.e., `display:block`), adds a down-arrow and Reset/Apply buttons in the module picker, displays current module section (comma-separated) in placeholder and tooltip text, and closes the drop-down on Escape keydown. Possible more as well, but those are the key changes.
screenshot


leobalter:
this is great! Thanks!

PR #991, opened 2016-04-20
Dump: update typeOf to extract extra complex type definition

danielmiester:
Older versions of Rhino seem to have an issue with the complexity of the conditional.
Pulling it out into it's own function helps, and also makes the code intent more readable.

### DIFF
--- a/src/dump.js
+++ b/src/dump.js
@@ -35,6 +35,19 @@ QUnit.dump = ( function() {
 		return join( "[", ret, "]" );
 	}
 
+	function isArray( obj ) {
+		return (
+
+			//Native Arrays
+		toString.call( obj ) === "[object Array]" ||
+
+		// NodeList objects
+		( typeof obj.length === "number" && obj.item !== undefined ) &&
+		( obj.length ?
+		obj.item( 0 ) === obj[ 0 ] :
+			( obj.item( 0 ) === null && obj[ 0 ] === undefined ) ) );
+	}
+
 	var reName = /^function (\w+)/,
 		dump = {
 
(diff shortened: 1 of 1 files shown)


src/dump.js
```diff
@@ -62,6 +62,20 @@ QUnit.dump = ( function() {
 			},
 			typeOf: function( obj ) {
 				var type;
+
+				function isArray() {
```
gibson042: This does make sense as its own function, but it shouldn't be defined inside `typeOf`... can you give it a parameter and move it up to line 37?

src/dump.js
```diff
@@ -62,6 +62,20 @@ QUnit.dump = ( function() {
...
+					return (
+
+						//Native Arrays
+						toString.call( obj ) === "[object Array]" ||
+
+						// NodeList objects
+						( typeof obj.length === "number" && obj.item !== undefined ) &&
+							( obj.length ?
+								obj.item( 0 ) === obj[ 0 ] :
+								( obj.item( 0 ) === null &&	obj[ 0 ] === undefined ) ) );
```
trentmwillis: There's an extra tab in the middle of this line, can you change to a space?

leobalter:
LGTM. There's a few whitespace issues that can be addressed before merging. Isolating this function will help improving it in the future.

danielmiester:
@leobalter I'm a bit confused now. Do I have to do anything at this point, or were you just commenting you will do on merge?

leobalter:
It's ok and all set, I just mention a [minor whitespace issue](https://github.com/jquery/qunit/pull/991/files#diff-44aa0638208d269c757aa37892695911R41) which I can take care of before merging.

danielmiester:
Ok. Thanks!

trentmwillis:
@leobalter fixed the whitespace and merged for you. @danielmiester thanks!

danielmiester:
YW :)

PR #992, opened 2016-04-21
Assert: Deprecate assert.push

### LINKED ISSUE #986: Deprecate assert.push
Please, see the comments below.

trentmwillis:
Closes #986 (I think).

I couldn't seem to find any code for `QUnit.push`, but did find code for `assert.push` (which is also deprecated).

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -1,6 +1,7 @@
 import dump from "./dump";
 import equiv from "./equiv";
 import { internalStop } from "./test";
+import { console } from "./globals";
 
 import config from "./core/config";
 import { objectType, objectValues } from "./core/utilities";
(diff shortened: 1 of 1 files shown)


jzaefferer:
I'm against landing this. The migration guide only documents `QUnit.push` being replaced by `assert.push`: http://qunitjs.com/upgrade-guide-2.x/#replace-qunit-push-with-this-push

This alias is trivial to maintain, there's hardly any gain by removing it now.

leobalter:
My bad, I explained this better on #986 and renamed it to reflect the necessary work.

@trentmwillis we can still reuse this PR to log a warning when assert.push is used instead of pushResult, not returning any abrupt completion. Do you wanna update this?

trentmwillis:
@leobalter sounds good, though I wonder if this should build on #918 once it lands?

leobalter:
seems reasonable. It would be a good use for that API.

trentmwillis:
@leobalter updated this. Now that #918 is closed, I used a simple `console.warn` for the deprecation.

leobalter:
Thanks!

PR #994, opened 2016-05-06
Component:  Assert.push

### LINKED ISSUE #890: QUnit.reporter( reporter [, options ] );

### LINKED ISSUE #986: Deprecate assert.push

rdugue:
This prints a message to the console stating that this function is now
deprecated, and pushResult is preferred.

Fixes #986

### DIFF
--- a/test/main/assert.js
+++ b/test/main/assert.js
@@ -394,3 +394,7 @@ QUnit.test( "throws", function( assert ) {
 		"throws fail when regexp doesn't match the error message"
 	);
 } );
+
+QUnit.test("push", function( assert ) {
+	assert.push( null );
+} );
--- a/src/assert.js
+++ b/src/assert.js
@@ -51,6 +51,7 @@ QUnit.assert = Assert.prototype = {
 	// Exports test.push() to the user API
 	// Alias of pushResult.
 	push: function( result, actual, expected, message, negative ) {
+		console.log("Warning: assert.push is deprecated, assert.pushResult should be used instead");
 		var currentAssert = this instanceof Assert ? this : QUnit.config.current.assert;
 		return currentAssert.pushResult( {
 			result: result,


test/main/assert.js
```diff
@@ -394,3 +394,7 @@ QUnit.test( "throws", function( assert ) {
 		"throws fail when regexp doesn't match the error message"
 	);
 } );
+
+QUnit.test("push", function( assert ) {
+	assert.push( null );
```
gibson042: What does this demonstrate?

trentmwillis:
Thanks for the initiative @rdugue! However, our current plan is to wait until our new warnings API has landed (see #918) and reuse the original PR I had made to close the issue (see #992).

rdugue:
Wooops, should have read further.  I'll go ahead and close this then.

test/main/assert.js
rdugue: It just shows that warning prints really.

PR #996, opened 2016-05-09
Returnable

Thaina:
let `QUnit.test()` and `QUnit.module()` can return Promise so we could grab result from one test to another

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -108,7 +108,6 @@ Test.prototype = {
 
 		if ( config.notrycatch ) {
 			runTest( this );
-			return;
 		}
 
 		try {
@@ -124,11 +123,16 @@ Test.prototype = {
 			if ( config.blocking ) {
 				internalStart( this );
 			}
+			
+			if(test.reject)
+				test.reject(e);
 		}
 
 		function runTest( test ) {
 			promise = test.callback.call( test.testEnvironment, test.assert );
 			test.resolvePromise( promise );
+			if(test.resolve)
+				test.resolve(promise);
 		}
 	},
 
@@ -604,7 +608,14 @@ function test( testName, callback ) {
 		callback: callback
 	} );
 
-	newTest.queue();
+	if(!Promise)
+		return newTest.queue();
+		
+	return new Promise(function(resolve,reject){
+		newTest.resolve	= resolve;
+		newTest.reject	= reject;
+		newTest.queue();
+	});
 }
 
 // Will be exposed as QUnit.skip
--- a/src/test.js
+++ b/src/test.js
@@ -124,16 +124,16 @@ Test.prototype = {
 				internalStart( this );
 			}
 			
-			if(test.reject) {
-				test.reject(e);
+			if ( test.reject ) {
+				test.reject( e );
 			}
 		}
 
 		function runTest( test ) {
 			promise = test.callback.call( test.testEnvironment, test.assert );
 			test.resolvePromise( promise );
-			if(test.resolve) {
-				test.resolve(promise);
+			if ( test.resolve ) {
+				test.resolve( promise );
 			}
 		}
 	},
@@ -610,8 +610,8 @@ function test( testName, callback ) {
 		callback: callback
 	} );
 
-	if(typeof Promise !== "undefined") {
-		return new Promise(function(resolve,reject){
+	if ( typeof Promise !== "undefined" ) {
+		return new Promise(function(resolve,reject) {
 			newTest.resolve	= resolve;
 			newTest.reject	= reject;
 			newTest.queue();
--- a/src/test.js
+++ b/src/test.js
@@ -124,15 +124,17 @@ Test.prototype = {
 				internalStart( this );
 			}
 			
-			if(test.reject)
+			if(test.reject) {
 				test.reject(e);
+			}
 		}
 
 		function runTest( test ) {
 			promise = test.callback.call( test.testEnvironment, test.assert );
 			test.resolvePromise( promise );
-			if(test.resolve)
+			if(test.resolve) {
 				test.resolve(promise);
+			}
 		}
 	},
 
@@ -608,14 +610,15 @@ function test( testName, callback ) {
 		callback: callback
 	} );
 
-	if(typeof Promise == "undefined")
-		return newTest.queue();
-		
-	return new Promise(function(resolve,reject){
-		newTest.resolve	= resolve;
-		newTest.reject	= reject;
-		newTest.queue();
-	});
+	if(typeof Promise !== "undefined") {
+		return new Promise(function(resolve,reject){
+			newTest.resolve	= resolve;
+			newTest.reject	= reject;
+			newTest.queue();
+		});
+	}
+	
+	newTest.queue();
 }
 
 // Will be exposed as QUnit.skip
--- a/src/test.js
+++ b/src/test.js
@@ -608,7 +608,7 @@ function test( testName, callback ) {
 		callback: callback
 	} );
 
-	if(!Promise)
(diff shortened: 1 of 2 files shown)


src/test.js
```diff
@@ -604,7 +608,14 @@ function test( testName, callback ) {
 		callback: callback
 	} );
 
-	newTest.queue();
+	if(!Promise)
```
campersau: I think you need `typeof Promise == "undefined"`. Otherwise you would get a `ReferenceError`.

trentmwillis:
Can you explain what use case this is trying to fulfill? And maybe provide an example?

In general, you don't want tests to be order-specific and it seems like this feature would be used for exactly that.

leobalter:
As @trentmwillis said, this is not telling what it tries to solve.

AFAIK, it returns the result of a module's callback and it mess with the return on notrycatch.

Plus, I need you to run `npm test` locally to catch code style issue before pushing.

Thaina:
What I use in my project is. I was use QUnit to test connection to database and get data of some users from it. Scope it into one test case

Then I want those users to do the next test. And want it to be another test case block

I want something like this from QUnit
```js
QUnit.test("Get Users",function(assert){
    //// this function return Promise<UserData[]>
    GetUsersFromDataBase().then(function(users){
        users.forEach(function(user,i){
             assert.ok(user.HasThisValue,"Check for user " + i + " has value we need");
        });
        return users;
    });
}).then(function(users){
    return Promise.all(users.map(function(user,i){
        return QUnit.test("TestUser:" + i,function(assert){
            return GetUserDetailByID(user.ID).then(function(userDetail){
                assert.ok(userDetail.HasThisValue,"Check for this value");
                assert.notOk(userDetail.NoValue,"NoValue must not exist");
                return userDetail;
            });
        });
    }));
}).then(function(userDetails){
    return QUnit.test("Get UserDetail from Promise.all",function(assert){
        assert.ok(userDetails.length > 10,"Should have enough user");
        return userDetails.length;
    });
}).then(function(size){ // anything more });
```

What I want is to chain a test and pass variable from scope without the need to make variable outside test block. Just return value needed for the next test

scottgonzalez:
That's an integration test and should be written as a single test with many assertions. Why are you opposed to that?

Thaina:
@scottgonzalez Because there are many times the test is very specific and should be in its own block. From one User there could be a test for 2 user interact with each other then test about user use some feature relate to money then test about user spend money in 3 different feature etc,etc,etc

I was query users from database and want to test the get users itself. Then I just want to return users I got to do more test on other feature. Instead of mock up user or query again I want to test it all in the same flow

This is just my example but the real thing would be a lot more complicate. If you insist on the idea that integration test must be written as a single test with many assertions even it could be split into each feature. Then yes, I would opposed that. Any test should be compact and specific on each feature. If it could be split into module then it should. And if it just sharing only some data then it should just return

I think this is the same concept as functional programming. And test should be just as it is

I'm not sure why "integration test must be written as a single test with many assertions".
Isn't it because normal unit test cannot return so you just need to do like that?

scottgonzalez:
> Because there are many times the test is very specific and should be in its own block.

While tests create blocks, many other things do too.

> Instead of mock up user or query again I want to test it all in the same flow

I never said mock again or query again. You want this all in the same flow, which means all in the same test.

> Any test should be compact and specific on each feature.

And yet you want to work from side effects of previous tests. You can't have both.

> I think this is the same concept as functional programming.

Functional programming would have no dependencies on previous tests. The functions would be pure.

> I'm not sure why "integration test must be written as a single test with many assertions".
> Isn't it because normal unit test cannot return so you just need to do like that?

No, it's because you limit your test based on the scope of what you're testing. Units are much smaller than integrations. An integration test by nature has many related steps.

Thaina:
@scottgonzalez The reason of me using unittest is to report test result in each block cleanly. Same flow don't need to be a same block

What you say is like if I need to make house with all room connected I need to have a house with one big room. Instead of many room with door and some hallway

Thaina:
> And yet you want to work from side effects of previous tests. You can't have both.

I can have both with the feature I was requesting so your argument is invalid

Thaina:
> Functional programming would have no dependencies on previous tests. The functions would be pure.
 Pure function does not mean it cannot get parameter returned from another function. Actually that's the point of function. It work by get parameter from another system and do something then return a result for the next and next and so on until the flow end

scottgonzalez:
> What you say is like if I need to make house with all room connected I need to have a house with one big room. Instead of many room with door and some hallway

Analogies like this never help. I will now withdraw from this conversation.

leobalter:
Using a single test or not, this is not a case for a specific return in modules. Each test can still return a promiseable object.

Even though, it's not a good practice to declare new tests (using `QUnit.test`) after another test is executed. For that, what you probably want is to define the promises inside the test, as @scottgonzalez pointed, and make the best use for the test's own assert instance. That will give you a nice summary for the test's successful and failing assertions.

I am not sure if that fits as a good practice, but experimentally you can also use nested modules to give scoped access to some async return.

leobalter:
This branch needs basic improvements to go forward. As the rationale is still not clear, test cases for these changes are fundamental, as I insist you use `npm test` before pushing anything. I'll close this PR for now, please reopen when it's ready for more review with these changes.

Thaina:
@leobalter for return in module. That is what I just added for that module could be chain like I do with a test

Something like
```js
QUnit.module("A",function(hooks){
    return QUnit.test("1",function(assert){
        assert.ok(TestSomething,"AssertSomething");
        return "ResultOfA";
    });
}).then(function(resultA /* "ResultOfA" */){
    QUnit.module("B",function(hooks){
        return QUnit.test("1",function(assert){
            assert.ok(TestAnotherThing,"AssertAnotherThing");
            return "ResultOfB";
        });
    });
}).then(function(resultB /* "ResultOfB" */) {  });
```

Thaina:
@leobalter I think to make a test after test is the cleanliest way to create block in QUnit

If it possible I would like to do a nested block. But I'm not sure why it don't work with async. It may be like you said, QUnit is not work if create a test after one was executed, so it can't generate nested module asynchronously after it execute the test

That's why I try to request this feature

I don't really need you to pull this request but I want to ask some way to display test asynchronously in block. This pull request is the solution I use and it works fine in QUnit on chrome browser

And what I want to do is actually normal when writing in synchronous. In synchronous we can cache variable and dynamically generate test case just fine
I just need to do the same in asynchronous

leobalter:
That's a use case for module hooks and nested modules.

> This pull request is the solution I use and it works fine in QUnit on chrome browser

Still, I need tests for QUnit. 

> I just need to do the same in asynchronous

You want to change QUnit to behave the custom way you use it. @scottgonzalez use QUnit with async operations, other developers use QUnit with async operations. I tried to help you with other ways to use QUnit on async operations. You are insisting on your way, and even you don't provide enough content to justify it, that would be the tests, or even a branch where you used `npm test` locally before committing your changes.

Thaina:
@leobalter OK please tell me

When we have array in synchronous code we can loop to each element and generate dynamic test block
```js

QUnit.test("TestSomething",function(assert){ assert.ok(true,"AssertSomething"); });

var users = [user0,user1,user2];
users.forEach(function(user,i){
    QUnit.test("User:" + i,function(assert){ assert.ok(user.OK,"User is OK"); });
});

users = users.filter(function(user) { return user.OK });
users.forEach(function(user,i){
    QUnit.test("User:" + user.Name,function(assert){
        assert.ok(user.CanSee,"User can see");
        assert.ok(user.CanDo,"User can do");
    });
});
```

How can you do this if you need to get users in asynchronous?

PR #997, opened 2016-05-09
Tests: Test on all Node.js releases supported by upstream

mgol:
The browserstack job is run only on Node.js 6. Upstream support for Node.js 5
ends in about a month so I didn't add it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
 sudo: false
 language: node_js
 node_js:
+  - "0.10"
+  - "0.12"
   - "4"
+  - "6"
 env:
   - NPM_SCRIPT=ci
-  - NPM_SCRIPT=browserstack
 matrix:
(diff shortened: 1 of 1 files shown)


mgol:
This is a better version of #887.

trentmwillis:
Seems good to me. Do we want to move Browserstack out of the "allowed failures" category?

mgol:
BrowserStack still fails from time to time (see [this recent run](https://travis-ci.org/jquery/qunit/builds/128878211)), I guess that's why failures in it are allowed.

I didn't change this setting, though, I just added more Node versions to the `ci` tests so maybe that's a separate discussion. :)

trentmwillis:
> BrowserStack still fails from time to time

Yeah, that's why I added the allowed failures to begin with. Let's keep it as is for now then.

PR #998, opened 2016-05-14
All: Decouple Core fixture reset from HTML reporter

trentmwillis:
Addresses #974. Will have a corresponding PR for documenting `QUnit.config.fixture`.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -150,19 +150,6 @@ extend( QUnit, {
 
 registerLoggingCallbacks( QUnit );
 
-function storeFixture() {
-
-	// Avoid overwriting user-defined values
-	if ( hasOwn.call( config, "fixture" ) ) {
-		return;
-	}
-
-	var fixture = id( "qunit-fixture" );
-	if ( fixture ) {
-		config.fixture = fixture.innerHTML;
-	}
-}
-
 function begin() {
 	var i, l,
 		modulesLog = [];
@@ -185,9 +172,6 @@ function begin() {
 				tests: config.modules[ i ].tests
 			} );
 		}
-
-		// Store fixture HTML for resetting later
-		storeFixture();
 
 		// The test run is officially beginning now
 		runLoggingCallbacks( "begin", {
--- a/src/core.js
+++ b/src/core.js
@@ -150,6 +150,19 @@ extend( QUnit, {
 
 registerLoggingCallbacks( QUnit );
 
+function storeFixture() {
+
+	// Avoid overwriting user-defined values
+	if ( hasOwn.call( config, "fixture" ) ) {
+		return;
+	}
+
+	var fixture = id( "qunit-fixture" );
+	if ( fixture ) {
+		config.fixture = fixture.innerHTML;
+	}
+}
+
 function begin() {
 	var i, l,
 		modulesLog = [];
@@ -172,6 +185,9 @@ function begin() {
 				tests: config.modules[ i ].tests
 			} );
 		}
+
+		// Store fixture HTML for resetting later
+		storeFixture();
 
 		// The test run is officially beginning now
 		runLoggingCallbacks( "begin", {
--- a/src/export.js
+++ b/src/export.js
@@ -25,12 +25,23 @@ QUnit.stop = function() {
 	);
 };
 
-QUnit.reset = function() {
-	throw new Error(
-		"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
-		"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
-	);
-};
+Object.defineProperty( QUnit, "reset", {
+	get: function() {
+		return function reset() {
+			throw new Error(
+				"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
+				"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
+			);
+		};
+	},
+	set: function() {
+		throw new Error(
+			"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
+			"Custom reset methods will no longer be invoked.\n" +
+			"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
+		);
+	}
+} );
 
 if ( defined.document ) {
 	if ( window.QUnit ) {
--- a/src/export.js
+++ b/src/export.js
@@ -25,22 +25,18 @@ QUnit.stop = function() {
 	);
 };
 
+function resetThrower() {
+	throw new Error(
+		"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
+		"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
+	);
+}
+
 Object.defineProperty( QUnit, "reset", {
 	get: function() {
-		return function reset() {
-			throw new Error(
-				"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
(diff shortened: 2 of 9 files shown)


src/core.js
```diff
@@ -150,6 +150,19 @@ extend( QUnit, {
...
 registerLoggingCallbacks( QUnit );
 
+function storeFixture() {
+
+	// Avoid overwriting user-defined values
+	if ( hasOwn.call( config, "fixture" ) ) {
+		return;
+	}
+
+	var fixture = id( "qunit-fixture" );
```
gibson042: Seeing this here, I'm pretty sure I was wrong to suggest moving it. `storeFixture` really does belong in reporter/html.js, but still benefits from the new `hasOwn` guard.

src/test.js
```diff
@@ -465,9 +465,11 @@ Test.prototype = {
 
 // Resets the fixture DOM element if available.
 function internalReset() {
-	var fixture = defined.document && document.getElementById &&
-			document.getElementById( "qunit-fixture" );
+	if ( config.fixture === null ) {
```
gibson042: This should be a loose equality check.

src/test.js
```diff
@@ -465,9 +465,11 @@ Test.prototype = {
 
 // Resets the fixture DOM element if available.
 function internalReset() {
```
gibson042: This code also belongs in reporter/html.js, which I think eliminates the need to define `id` twice. I like your [`testStart` suggestion](https://github.com/jquery/qunit/issues/974#issuecomment-201966604).

src/export.js
```diff
@@ -25,12 +25,23 @@ QUnit.stop = function() {
...
+Object.defineProperty( QUnit, "reset", {
+	get: function() {
+		return function reset() {
+			throw new Error(
+				"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
+				"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
+			);
+		};
+	},
+	set: function() {
```
gibson042: This would be simpler (and protected from copypasta divergence) as e.g.
```js
function resetThrower() {
	throw new Error(
		"QUnit.reset is removed in QUnit 2.0 without replacement.\n" +
		"Details in our upgrade guide at https://qunitjs.com/upgrade-guide-2.x/"
	);
}
Object.defineProperty( QUnit, "reset", {
	get: function() {
		return resetThrower;
	},
	set: resetThrower
} );
```

test/main/test.js
```diff
@@ -24,6 +24,21 @@ QUnit.test( "basics", function( assert ) {
...
+	document.getElementById( "qunit-fixture" ).innerHTML = "foobar";
+} );
+
+QUnit.test( "basics for null reset", function( assert ) {
+	assert.equal(
+		document.getElementById( "qunit-fixture" ).innerHTML,
+		"foobar",
+		"did not reset after last test"
+	);
+	QUnit.config.fixture = "test markup";
```
gibson042: This is missing tests for `QUnit.config.fixture = userSpecifiedContent`, and is also a bit ugly (admittedly a preexisting condition). Could we use beforeEach/afterEach to make it more robust and (arguably) a little cleaner?
```js
QUnit.module( "fixture", function( hooks ) {
	var failure = false,
		values = [ /* initial value (see unshift below), */ "<b>ar</b>", undefined ],

		// Use values as a "not-set" sentinel
		originalValue = values;
	hooks.beforeEach( function() {
		if ( originalValue === values ) {
			originalValue = QUnit.config.fixture;
			values.unshift( originalValue );
		}
	} );

	// Set QUnit.config.fixture for the next test, propagating failures to recover the sequence
	hooks.afterEach( function( assert ) {
		if ( failure || values.length === 0 ) {
			assert.ok( false, "prior failure" );
			QUnit.config.fixture = originalValue;
		} else {
			QUnit.config.fixture = values.shift();
		}
	} );

	QUnit.test( "setup", function( assert ) {
		document.getElementById( "qunit-fixture" ).innerHTML = "foo";
		assert.equal( values.length, 3, "proper sequence" );
		failure = failure || values.length !== 3;
	} );
	QUnit.test( "automatically reset", function( assert ) {
		var contents = document.getElementById( "qunit-fixture" ).innerHTML;
		assert.equal( contents, "test markup" );
		assert.equal( values.length, 2, "proper sequence" );
		failure = failure || values.length !== 2 || contents !== "test markup";
	} );
	QUnit.test( "user-specified", function( assert ) {
		var contents = document.getElementById( "qunit-fixture" ).innerHTML;
		document.getElementById( "qunit-fixture" ).innerHTML = "baz";
		assert.equal( contents, "<b>ar</b>" );
		assert.equal( values.length, 1, "proper sequence" );
		failure = failure || values.length !== 1 || contents !== "<b>ar</b>";
	} );
	QUnit.test( "disabled", function( assert ) {
		var contents = document.getElementById( "qunit-fixture" ).innerHTML;
		assert.equal( contents, "baz" );
		assert.equal( values.length, 0, "proper sequence" );
		failure = failure || values.length !== 0 || contents !== "baz";
	} );
} );
```

gibson042:
@trentmwillis I just figured out why my opinion changes every time I consider this topic... conflation of the HTML reporter with the HTML _runner_. Could you move those capture/reset callbacks into something like fixture.js, to be loaded before any reporter?

trentmwillis:
@gibson042 I split the fixture cache/reset out into a new directory `runner`, to eventually use with the storage code as well. However, after seeing the final result, I am uncertain if there is benefit to not just putting this into the reporter, as the reporter is the only context in which the fixture stuff makes sense.

It seems to me like the original goal was two fold: stop coupling Core to HTML Reporter and allow users to disable the fixture resets. Both of which can be accomplished with this same code in the reporter.

test/main/test.js
```diff
@@ -9,19 +9,53 @@ QUnit.test( "expect query and multiple issue", function( assert ) {
...
-	assert.equal(
-		document.getElementById( "qunit-fixture" ).innerHTML,
-		"test markup",
-		"automatically reset"
-	);
+	QUnit.test( "automatically reset", function( assert ) {
+		var contents = document.getElementById( "qunit-fixture" ).innerHTML;
+		assert.equal( contents, "test markup" );
+		assert.equal( values.length, 2, "proper sequence" );
+		failure = failure || values.length !== 2 || contents !== "test markup";
```
gibson042: Both of these `"test markup"` strings should be `originalValue` (I _knew_ I was missing something).

gibson042:
> the reporter is the only context in which the fixture stuff makes sense… the original goal was two fold: stop coupling Core to HTML Reporter and allow users to disable the fixture resets

I think the case is more clear for also pulling out `sessionStorage`, but it is already conceivable to imagine using fixture resets while ignoring (or even excluding) the HTML reporter. In other words, progress towards the tertiary goal of a much more passive reporter—reading only configuration and events, writing only form-exposed configuration and commands.

At any rate, this looks good to me.

trentmwillis:
> progress towards the tertiary goal of a much more passive reporter

That seems reasonable to me. Would like @leobalter to review before merging.

leobalter:
>  it is already conceivable to imagine using fixture resets while ignoring (or even excluding) the HTML reporter

This is true.

I look to the runner folder and ask myself if that's right. I wonder if it should be placed under the src folder, but that's enough to create a bikeshed, and at the end it only exposes my bad decision to set the reporter folder outside of src.

The code and tests look good! I'm excited to see the fixture is finally decoupled from core. The legacy leaves us in a long path for modularity, so each improvement is awesome.

:shipit:

Issue #1000, opened 2016-05-16
urlParams does not decode "+" character in query params


bantic:
The [`getUrlParams` function](https://github.com/jquery/qunit/blob/f13038f8530da11a4c3793f35ffe4d4e1d0531a2/reporter/urlparams.js#L69) does not decode the plus character (`+`) in query parameters, which can cause QUnit to fail to match modules or test names when the `module` or `filter` query parameters include spaces encoded as plus characters.

For example, if a test named "my test" exists, and the tests page is at "http://localhost:4200/tests", then the URL "http://localhost:4200/tests?filter=my+test" will fail to run the test because it does not decode the parameter "my+test" to "my test".

The test- and module-filtering still works in most cases because QUnit's [url-generating function `setUrl`](https://github.com/jquery/qunit/blob/347a8381c803a2122279e93af51892c4351f841d/reporter/html.js#L211) encodes spaces in the query parameters as `%20`, and `getUrlParams` *does* decode `%20` as a space.

We have a use case where a uri-normalization library is running on the test page and it replaces the `%20` with `+` in the url's query parameters before QUnit runs `getUrlParams`.

According to the W3, spaces in query parameter names and values (mime type application/x-www-form-urlencoded) [should be replaced with `+`, not `%20`](https://www.w3.org/TR/REC-html40/interact/forms.html#h-17.13.4.1), so it seems like the correct behavior would be:
  * for QUnit's HTML Reporter UI to use plus characters instead of spaces when generating query params
  * for QUnit's `getUrlParams` to decode plus characters as spaces

rwjblue:
@trentmwillis - FWIW, this is the result of the bug/issue that I pestered you about in slack a week or two ago.  Many thanks to @bantic for tracking down what is going on (and his continued efforts to fix many encoding/decoding related issues in Ember's route-recognizer).

trentmwillis:
@rwjblue, I thought this might be related to that. Seems like a reasonable ask to support `%20` and `+` as space-encodings for params since URIs should support them anyway.

gibson042:
I don't think we need to go to the trouble of complicating `setUrl`, but I approve of more liberal decoding.

bantic:
neat — this is issue number 1000 and it was closed by commit number 1000. :)

leobalter:
https://twitter.com/qunitjs/status/732620820049235968

PR #1001, opened 2016-05-16
Core: Decode "+" to " " (space) in url params

bantic:
This fixes the decoding part of issue gh-1000.

Ref gh-1000

### DIFF
--- a/test/reporter-urlparams.js
+++ b/test/reporter-urlparams.js
@@ -1,23 +1,26 @@
 if ( !location.search ) {
 	location.replace( "?implicit&explicit=yes&array=A&array=B&escaped%20name&" +
-		"module=urlParams&notrycatch&custom&customArray=a&customArray=b" );
+		"module=urlParams+module&notrycatch&custom&customArray=a&customArray=b&" +
+		"filter=urlParams%20module" );
 }
 
 QUnit.config.urlConfig.push( "custom", "customArray" );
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Thanks for working on this @bantic!

trentmwillis:
This looks good to me.

@leobalter would you mind taking a look before I merge? While reasonable, I want to make sure you don't see any issues with this landing in the pending 2.0 release.

test/reporter-urlparams.js
```diff
@@ -1,23 +1,26 @@
 if ( !location.search ) {
 	location.replace( "?implicit&explicit=yes&array=A&array=B&escaped%20name&" +
-		"module=urlParams&notrycatch&custom&customArray=a&customArray=b" );
+		"module=urlParams+module&notrycatch&custom&customArray=a&customArray=b&" +
+    "filter=urlParams%20module" );
```
gibson042: The spacing is off here. We allow up to 100 characters per line (where tab width is 4), so it might fit on one. But if not, please align it with the above.

leobalter:
+1. With the whitespace fixed this looks good to land.

bantic:
@leobalter @gibson042 Fixed the spacing in the test (and noticed I had the spacing wrong for the `decodeQueryParam` function, also fixed now). thanks!

trentmwillis:
Thanks @bantic!

Issue #984, opened 2016-04-14
Confirm proper async isolation


gibson042:
https://github.com/jquery/qunit/pull/976#discussion_r59634666 suggests that some state may be leaking.

platinumazure:
What sort of state might be leaking? Is this something that needs further investigation, or do we already have a good idea where/why this is happening?

gibson042:
I haven't investigated the why, but the where is in the function returned by `assert.async`—we _should_ be able to pass `test` to `resumeProcessing`, but making that change breaks the test suite. So something is amiss and needs analysis.

PR #1002, opened 2016-05-24
All: Prevent async tests from having multiple resume timeouts

trentmwillis:
Addressing #984.

The root issue is that it was possible for a single async test to have multiple `resumeProcessing` timeouts. [This parallel assert.async calls test](https://github.com/jquery/qunit/blob/master/test/main/async.js#L58-L71) is probably the easiest case to see why that would be a problem.

1. `done1` would get called and queue up a `resumeProcessing` timeout.
2. `done2` would get called and queue up another `resumeProcessing` timeout. `test.semaphore` is now at `0`.
3. The first timeout executes and hits `begin()` to resume the test suite
4. The second timeout executes and hits `begin()` as well, which causes issues because we're already running

The reason this worked without passing in `test` before is that `config.current` would update between the time 3 and 4 happened above which would prevent `begin()` from getting called a second time.

My solution is to check to see if `test` is already waiting to resume when passed into `resumeProcessing`.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -44,7 +44,7 @@ QUnit.assert = Assert.prototype = {
 
 			test.semaphore -= 1;
 			popped = true;
-			resumeProcessing();
+			resumeProcessing( test );
 		};
 	},
 
--- a/src/core.js
+++ b/src/core.js
@@ -229,25 +229,22 @@ function resumeProcessing( test ) {
 
 	// A slight delay to allow this iteration of the event loop to finish (more assertions, etc.)
 	if ( defined.setTimeout ) {
-		if ( test && test.resuming ) {
-			return;
-		}
-
-		var resume = setTimeout( function() {
+		setTimeout( function() {
 			var current = test || config.current;
-			if ( current && current.semaphore > 0 ) {
+			if ( current && ( current.semaphore > 0 || current.resumed ) ) {
 				return;
 			}
+
 			if ( config.timeout ) {
 				clearTimeout( config.timeout );
 			}
 
+			if ( current ) {
+				current.resumed = true;
+			}
+
 			begin();
 		}, 13 );
-
-		if ( test ) {
-			test.resuming = resume;
-		}
 	} else {
 		begin();
 	}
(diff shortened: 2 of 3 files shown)


src/assert.js
```diff
@@ -44,7 +44,7 @@ QUnit.assert = Assert.prototype = {
 
 			test.semaphore -= 1;
 			popped = true;
-			resumeProcessing();
+			resumeProcessing( test );
```
gibson042: :clap:

src/core.js
```diff
@@ -229,7 +229,11 @@ function resumeProcessing( test ) {
 
 	// A slight delay to allow this iteration of the event loop to finish (more assertions, etc.)
 	if ( defined.setTimeout ) {
-		setTimeout( function() {
+		if ( test && test.resuming ) {
```
gibson042: This should guard the `begin()`, not the `setTimeout`. I believe the logic here will break down when a test's `assert.async()`s fail to resolve within 13ms of each other, e.g.
```js
QUnit.test( "fast and slow", function( assert ) {
	var done1 = assert.async(),
		done2 = assert.async();
	setTimeout( function() {
		assert.ok( true );
		done1();
	} );
	setTimeout( function() {
		assert.ok( true );
		done2();
	}, 100 );
} );
```

src/core.js
```diff
@@ -240,6 +244,10 @@ function resumeProcessing( test ) {
 
 			begin();
 		}, 13 );
+
+		if ( test ) {
+			test.resuming = resume;
```
gibson042: I'd prefer a boolean—and per the above, it's probably `test.resumed` (this property doesn't leak to consumers, does it?).

src/core.js
trentmwillis: Ah true. I'll be sure to add a similar test to the suite as well.

src/core.js
trentmwillis: Nope, because we don't expose the actual test object at any point in time (that I'm aware of at least).

trentmwillis:
@gibson042 updated.

trentmwillis:
@gibson042 / @leobalter would appreciate a review so I can close this out.

gibson042:
Sorry about dropping this. LGTM!

leobalter:
:shipit:

Issue #787, opened 2015-03-17
Pendings tests


jzaefferer:
Something that came out of a discussion with Yehuda: During a big refactoring they have hundreds of failing tests. They turned them off using `QUnit.skip`, but that doesn't tell them which of the skipped tests starting passing again at some point. To quickly test changes in passing tests, they ended up with this:

``` js
QUnit.skip = QUnit.test;
```

There's probably a better way to deal with that situation. One idea is to have something like a `QUnit.pending` method, that has the same signature as `QUnit.test()`, but will reverse the result. So if all assertions pass, the test fails, if at least one fails, the test passes.

The HTML reporter could use a similar marker as `QUnit.skip`.

Thoughts?

---

_Update 2015-04-15_:

After some discussion, we came up with this:

``` js
// This test will show up with a "todo" badge, with a similar style as the "skipped" badge
// It will show as passed when at least one assertion fails, or fail when all assertions pass
QUnit.todo( "this isn't implemented, yet" function( assert ) {
  assert.ok( featureTest(), "waiting for feature" );
} );

QUnit.testDone( function( details ) {
  console.log( details.name, details.todo );
  //> "this isn't implemented, yet" true
} );

QUnit.log( function( details ) {
  console.log( details.message, details.todo );
  //> "waiting for feature" true
} );
```

wycats:
That sounds great to me. We'd use this!

wycats:
It could also be useful for tests that we expect to fail in all browsers, but which mark code that could be removed once browsers started implementing a feature. If we start to see the pending test "fail" in Saucelabs, it's time to investigate using the browser feature.

jzaefferer:
That's an interesting idea. In both cases this would be considered a "temporary" test, so I still think some visual highlight makes sense.

gibson042:
This was touched on in the conversation that led to `QUnit.skip`.

Relevant excerpts:
https://github.com/jquery/qunit/issues/637#issuecomment-53917404
> [[TAP](http://testanything.org/tap-specification.html)] **TODO**: Expects failure (i.e., a reversal of normal operation—passes are interesting instead of failures, and failures do not effect pass→fail state changes in the containing run/module/suite/test). Reporters must take care to separate TODO failures from non-TODO failures, and should highlight TODO passes.

https://github.com/jquery/qunit/issues/637#issuecomment-53944341
> todo means "run, but don't let failures fail the containing group"

[...]

PR #1006, opened 2016-06-16
Core: Implement QUnit.todo

trentmwillis:
As discussed in #787.

Will need updates to the test runner in order to not fail on CI, but figured I would get feedback on the implementation/API first.

One potential issue to note: In the aggregate cases (e.g., `moduleDone` / `done`) it is not possible to tell if a `todo` test has "failed". This case is identified at the individual test level as `details.total === details.passedUnexpectedly` but we have no knowledge of that in the aggregate scenarios. Not sure if this needs to be addressed as it is easily worked around by tracking if you have had any failures at the test-level (as is done here).

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -13,6 +13,7 @@ QUnit.init = function() {
 var config = QUnit.config,
 	document = window.document,
 	collapseNext = false,
+	hasFailed = false,
 	hasOwn = Object.prototype.hasOwnProperty,
 	unfilteredUrl = setUrl( { filter: undefined, module: undefined,
 		moduleId: undefined, testId: undefined } ),
@@ -615,16 +616,36 @@ QUnit.done( function( details ) {
 			"</span> assertions of <span class='total'>",
 			details.total,
 			"</span> passed, <span class='failed'>",
-			details.failed,
-			"</span> failed."
-		].join( "" );
+			details.failedUnexpectedly,
+			"</span> failed"
+		];
+
+	var expectedFailures = details.failed - details.failedUnexpectedly;
+
+	if ( expectedFailures ) {
+		html = html.concat( [
+			", <span class='failed-as-expected'>",
+			expectedFailures,
+			"</span> failed as expected"
+		] );
+	}
+
+	if ( details.passedUnexpectedly ) {
+		html = html.concat( [
+			", <span class='passed-unexpectedly'>",
+			details.passedUnexpectedly,
+			"</span> passed unexpectedly"
+		] );
+	}
+
+	html.push( "." );
 
 	if ( banner ) {
-		banner.className = details.failed ? "qunit-fail" : "qunit-pass";
+		banner.className = hasFailed ? "qunit-fail" : "qunit-pass";
 	}
 
 	if ( tests ) {
-		id( "qunit-testresult" ).innerHTML = html;
+		id( "qunit-testresult" ).innerHTML = html.join( "" );
 	}
 
 	if ( config.altertitle && document.title ) {
@@ -632,13 +653,13 @@ QUnit.done( function( details ) {
 		// Show ✖ for good, ✔ for bad suite result in title
 		// use escape sequences in case file gets loaded with non-utf-8-charset
 		document.title = [
-			( details.failed ? "\u2716" : "\u2714" ),
+			( details.failedUnexpectedly ? "\u2716" : "\u2714" ),
 			document.title.replace( /^[\u2714\u2716] /i, "" )
 		].join( " " );
 	}
 
 	// Clear own sessionStorage items if all tests passed
-	if ( config.reorder && defined.sessionStorage && details.failed === 0 ) {
+	if ( config.reorder && defined.sessionStorage && details.failedUnexpectedly === 0 ) {
 		for ( i = 0; i < sessionStorage.length; i++ ) {
 			key = sessionStorage.key( i++ );
 			if ( key.indexOf( "qunit-test-" ) === 0 ) {
@@ -782,7 +803,7 @@ QUnit.log( function( details ) {
 } );
 
 QUnit.testDone( function( details ) {
-	var testTitle, time, testItem, assertList,
+	var testTitle, time, testItem, assertList, todo,
 		good, bad, testCounts, skipped, sourceName,
 		tests = id( "qunit-tests" );
 
@@ -795,7 +816,7 @@ QUnit.testDone( function( details ) {
 	assertList = testItem.getElementsByTagName( "ol" )[ 0 ];
 
 	good = details.passed;
-	bad = details.failed;
+	bad = details.failedUnexpectedly !== 0 || (details.todo && details.passedUnexpectedly === details.total);
 
 	// Store result when possible
 	if ( config.reorder && defined.sessionStorage ) {
@@ -806,7 +827,7 @@ QUnit.testDone( function( details ) {
 		}
 	}
 
-	if ( bad === 0 ) {
+	if ( !bad ) {
 
 		// Collapse the passing tests
 		addClass( assertList, "qunit-collapsed" );
@@ -823,8 +844,8 @@ QUnit.testDone( function( details ) {
 	// The testItem.firstChild is the test name
 	testTitle = testItem.firstChild;
 
-	testCounts = bad ?
-		"<b class='failed'>" + bad + "</b>, " + "<b class='passed'>" + good + "</b>, " :
+	testCounts = details.failed ?
+		"<b class='failed'>" + details.failed + "</b>, " + "<b class='passed'>" + good + "</b>, " :
 		"";
 
 	testTitle.innerHTML += " <b class='counts'>(" + testCounts +
@@ -841,7 +862,20 @@ QUnit.testDone( function( details ) {
 			toggleClass( assertList, "qunit-collapsed" );
 		} );
 
-		testItem.className = bad ? "fail" : "pass";
+		if ( bad ) {
+			testItem.className = "fail";
+			hasFailed = true;
+		} else {
+			testItem.className = "pass";
+		}
+
+		if ( details.todo ) {
+			testItem.className += " todo";
+			todo = document.createElement( "em" );
+			todo.className = "qunit-todo-label";
+			todo.innerHTML = "todo";
+			testItem.insertBefore( todo, testTitle );
+		}
 
 		time = document.createElement( "span" );
 		time.className = "runtime";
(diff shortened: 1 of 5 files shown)


leobalter:
> In the aggregate cases (e.g., moduleDone / done) it is not possible to tell if a todo test has "failed". This case is identified at the individual test level as details.total === details.passedUnexpectedly

This `passedUnexpectedly` looks like a regular fail condition. The `todo` API concept is a bit inverted from regular tests, but I believe we should keep what we have for passed and failed.

I would try forcing a regular pass and fail results before adding any details to the callbacks, even if they are weird using the `todo` method.

trentmwillis:
> This `passedUnexpectedly` looks like a regular fail condition.

That's because it is. See [this comment](https://github.com/jquery/qunit/issues/787#issuecomment-223149694) for my reasoning. It doesn't impact backwards compatibility at all though, as the passed and failed values are still the same as before.

> I would try forcing a regular pass and fail results before adding any details to the callbacks, even if they are weird using the todo method.

This approach doesn't really work though. Consider the following:

```js
todo('some todo', function(assert) {
  assert.ok(true);
  assert.ok(false);
});
```

This should pass, but if we try to force the existing pass/fail model it isn't possible. This is because our current set up assumes all the assertions for a test must have the same result (passing) in order to be "good", but for todo tests we assume that a test can pass with assertions having different results (a mix of passing and failing).

leobalter:
I'm resistant with adding more details to the log callback. 

A passing condition for a .todo is not the same concept as in a .test. The example you gave is a passing condition, and it's weird.

What we can do is to not add `.todo` runs as _passed_ tests, but adding they to _failed_ only when they fail for their expected behavior (with every expected assertions passing).

They might count as _skipped_ tests or we should create a new _todo_ count for them. 

I believe some reporters rely on the number of failed tests to tell something went wrong.

I believe the following would be fine:

```js
details: {
  ...
  passed: 42, // Referring to passing .test only
  failed: 2, // any failures, including global errors
  skipped: 4 // skipped tests, does not count as passing today
  todo: 10 // .todo tests, maybe only those that haven't failed, good for reporters
}
```

reporter/html.js
```diff
@@ -615,30 +616,50 @@ QUnit.done( function( details ) {
...
-			"</span> failed."
-		].join( "" );
+			details.failedUnexpectedly,
+			"</span> failed"
+		];
+
+	var expectedFailures = details.failed - details.failedUnexpectedly;
+
+	if ( expectedFailures ) {
+		html = html.concat( [
```
gibson042: Nit: `html.push( …, … )` is slightly better.

reporter/html.js
```diff
@@ -615,30 +616,50 @@ QUnit.done( function( details ) {
...
+		id( "qunit-testresult" ).innerHTML = html.join( "" );
 	}
 
 	if ( config.altertitle && document.title ) {
 
 		// Show ✖ for good, ✔ for bad suite result in title
 		// use escape sequences in case file gets loaded with non-utf-8-charset
 		document.title = [
-			( details.failed ? "\u2716" : "\u2714" ),
+			( details.failedUnexpectedly ? "\u2716" : "\u2714" ),
```
gibson042: What motivates use of `details.failedUnexpectedly` vs. `hasFailed`?

reporter/html.js
```diff
@@ -795,7 +816,7 @@ QUnit.testDone( function( details ) {
 	assertList = testItem.getElementsByTagName( "ol" )[ 0 ];
 
 	good = details.passed;
-	bad = details.failed;
+	bad = details.failedUnexpectedly !== 0 || (details.todo && details.passedUnexpectedly === details.total);
```
gibson042: I think there's an edge case here where `details.total` is 0... not that we currently report that, but the second clause would be more robust as something like `details.passedUnexpectedly > 0 && details.passedUnexpectedly === details.total`.

test/logs.js
```diff
@@ -279,11 +293,41 @@ QUnit.test( module2Test4.name, function( assert ) {
...
+		passedUnexpectedly: 0,
 		skipped: true,
+		todo: false,
 		testId: module2Test3.testId
 	}, "testDone context" );
 } );
 
+QUnit.todo( module2Test5.name, function( assert ) {
+  assert.ok( false, "expected failure" );
+	assert.ok( true, "unexpected pass" );
```
gibson042: Weird spacing here.

reporter/html.js
```diff
@@ -615,30 +616,50 @@ QUnit.done( function( details ) {
...
 		document.title = [
-			( details.failed ? "\u2716" : "\u2714" ),
+			( details.failedUnexpectedly ? "\u2716" : "\u2714" ),
 			document.title.replace( /^[\u2714\u2716] /i, "" )
 		].join( " " );
 	}
 
 	// Clear own sessionStorage items if all tests passed
-	if ( config.reorder && defined.sessionStorage && details.failed === 0 ) {
+	if ( config.reorder && defined.sessionStorage && details.failedUnexpectedly === 0 ) {
```
jzaefferer: Same as above, should use `hasFailed` here?

leobalter:
This is one of my favorite features to land next. I believe the holding point here is on the passedUnexpectedly and failedUnexpectedly. The other comments should be easily covered.

@jquery/qunit All your feedback here is really important to support @trentmwillis and I. Regardless the way we decide to move this on, we need to land a strong API as we won't be able to remove it easily later.

jzaefferer:
Which part of the API needs to be discussed?

leobalter:
following the [discussion above](https://github.com/jquery/qunit/pull/1006#issuecomment-228751036) whether we wanna roll with `passedUnexpectedly` and `failedUnexpectedly`.

I am not convinced it is good, we could solve it with the existing details' properties.

If the team believes these new properties are good, I'm gonna accept them. Trent is doing a great work and I feel bad for hold this for long, that's why I called for more feedback.

jzaefferer:
```
QUnit.todo("wip", function(assert) {
  assert.ok(false)
})
> testEnd { passed: 0, failed: 0, passedUnexpectedly: 0, failedUnexpectedly: 1 }
```
Is this what's currently implemented?

What's the alternative? This?
```
> testEnd { passed: 1, failed: 0 }
```

leobalter:
There's a log test for it:

```js
QUnit.todo( module2Test5.name, function( assert ) {
	assert.ok( false, "expected failure" );
	assert.ok( true, "unexpected pass" );
} );

...
	assert.deepEqual( testDoneContext, {
		...
		failed: 1,
		passed: 1,
		total: 2,
		failedUnexpectedly: 0,
		passedUnexpectedly: 1,
		...
	}, "testDone context" );
...
```

leobalter:
With @trentmwillis's approach, a todo test will report like a regular test in the amount of failed and passed assertions.

Reviewing this I'm probably changing my previous request with the following statements: 

- yes, it makes sense to preserve passed/failed. I've got it.
- the *Unexpectedly properties are still weird. Maybe: can we expose only a single property with the test result? 

This would be helpful to improve the logic on any reporter, including the built-in html replacing this:

```js
bad = details.failedUnexpectedly !== 0 || (details.todo && details.passedUnexpectedly === details.total);
```

Bringing this logic to the test result, we would assert that same log as:

```diff
assert.deepEqual( testDoneContext, {
        ...
        failed: 1,
        passed: 1,
        total: 2,
-        failedUnexpectedly: 0,
-        passedUnexpectedly: 1,
+        result: true,
        ...
    }, "testDone context" );
```

That would require less logic from reporters and keep the failed/passed in the same way.

jzaefferer:
Adding a boolean property to indicate if the test passed or not seems fine to me, though it doesn't map well to the `Test` structure in js-reporters: https://github.com/js-reporters/js-reporters#event-data (scroll down to "Test"). I think the closest equivalent is the `status` property, which can be "passed", "failed" and "skipped".

In this PR, there's also a `todo` boolean property, right? I'm not sure if that can be mapped to one of the existing js-reporters `Test` properties, or if it should be added. @flore77 thoughts?

platinumazure:
@jzaefferer I'm not familiar with the maturity level of the js-reporters spec, so forgive me if this is a dumb question. Could we augment the status enum to also include "pending"/"todo" (and perhaps, "ready" or "implemented" for those cases when all assertions pass)?

My personal preference would be for the assertions to be represented as passing/failing as normal (i.e., don't invert their meaning, don't create passedUnexpectedly/failedUnexpectedly assertion arrays). Instead, the semantics would be that the overall test status would be something besides passing, to indicate the pending/todo nature of the test.

flore77:
> here's also a todo boolean property

@jzaefferer I think this can be mapped also on to the `status` property. From what I have seen I think it is denoting if a test is a `todo test`. Am I right @trentmwillis or I didn't get it? And how also @platinumazure has suggested, I think we can introduce a new status or more.

>  include "pending"/"todo"

@platinumazure I would prefer `todo` or something else, because `pending` is used in other testing frameworks (Mocha) for skipped tests.

platinumazure:
@flore77 :+1: I'm not attached to any particular name for this new status-- I just want a new status, so we don't have to muck around with the assertion count properties and can continue to have "passed" mean "passed", and "failed" mean "failed".

jzaefferer:
Wouldn't you then need both `todo-passed` and `todo-failed`? Unlike `skipped`, a todo test can still pass/fail.

flore77:
@jzaefferer yes, or since it is a special type of test, the reporter can check the `assertions` and `errors` properties to establish if it is a pass or a fail, but this would require extra logic in the reporter and I don't find any advantage in this.

So I think it's better only by looking at the status and to know if it is a fail or a pass, so something like `todo-passed` and `todo-failed` sounds good.

leobalter:
@trentmwillis I'll be in CA this week for the TC39 meeting, but I'll have some free time this Friday and I would like to spend it on some work force to get this done. I'm looking forward to see this landing, it's a great work and an anticipated feature.

gibson042:
I just looked over this again, and found nothing that hadn't already been commented on. Per https://github.com/jquery/qunit/pull/1006#issuecomment-239470608 , though, we should explicitly decide whether the new details properties should quantify unexpected passes & failures (e.g., a failing non-todo test reports `failedUnexpectedly: N` [N > 0]), or expected ones (e.g., `failedAsExpected: 0` for the same case). I don't strongly favor one over the other, but like @leobalter said, we won't be able to change it later.

trentmwillis:
Sorry for letting this sit for so long. Work got busy and then I didn't feel like reading through all the comments haha.

Anyway, now that I am caught up and have thought about the issue some more. I think most of the confusion comes from the fact that I was trying to get the assertion counts to communicate what was happening, but in reality, we should be concerned with the test counts. In other words, we care if a test has failed and not if an individual assertion has failed.

With that in mind, I think an approach similar to @leobalter [outlined here](https://github.com/qunitjs/qunit/pull/1006#issuecomment-231770516) makes the most sense. I'll work on updating this branch with that approach and also will submit a PR to the js-reporters repo to  discuss how I think it can integrate there.

trentmwillis:
Just opened #1080 to supersede this. I wanted to preserve this implementation and discussion, but consider it to be substantially different from the new approach.

Issue #1004, opened 2016-05-31
Skipped tests runtime should be undefined


flore77:
## Problem
While working on the [js-reporter](https://github.com/js-reporters/js-reporters) project I have discovered that also the `runtime` of `skipped tests` is a `Number`. 

We on the `js-reporters` think that the `runtime` should be `undefined`, because the test does not run, it is skipped, so it cannot have a runtime.

I looked over the docs about the `runtime` prop on the `testDone` callback, where it is described:

> The total runtime in millseconds of the test, including beforeEach and afterEach

But, also all `hooks` for a skipped test, are also [skipped](https://github.com/jquery/qunit/blob/44662651017be2fedd550685227d6916767ba16b/src/test.js#L188-L190).

## Solution

To achieve this, we should only [modify the runtime](https://github.com/jquery/qunit/blob/44662651017be2fedd550685227d6916767ba16b/src/test.js#L231) of the object that is passed as param to the callback on `testDone`, something like:

```js
runLoggingCallbacks( "testDone", {
  ...
  runtime: (!!this.skip) ? undefined : this.runtime,
  ... 
} );
```

I have done this change locally and all tests are passing. I want also to write some tests to check the tests runtime for a `good coverage`. Here is a [gist](https://gist.github.com/flore77/106860feffdeaa24c8bad12705430943) about how I think to test it, but I don't know how to simulate a test that is failing, so that  I can get its runtime.

leobalter:
I'm not sure how the type is a concern, as the expected value for skipped tests is 0 or close (I hope) and replacing it with undefined would be a breaking change.

Rather than reading the runtime, js-reporters should ignore the value on skipped tests, as it is irrelevant as an undefined.

trentmwillis:
I agree with @leobalter the value shouldn't matter too much.

@flore77 could you link to an explanation (or explain yourself), why using `undefined` is preferable over an actual number? In my mind, it makes sense to report a number (even `0`), for more accuracy in understanding what the test suite is doing.

jzaefferer:
How about making sure that runtime is always 0 for skipped tests? That should work for js-reporters as well.

leobalter:
> How about making sure that runtime is always 0 for skipped tests? That should work for js-reporters as well.

Works for me. 👍

jzaefferer:
And I guess if we set it to 0, we don't really need to add tests for that...

flore77:
[...]

PR #1007, opened 2016-06-21
Core: Ensure runtime for skipped tests is 0

trentmwillis:
Addressing #1004.

This ensures that the `runtime` property during `testDone` logging for skipped tests is alway `0`. I think a future state where we remove the property all together is possible, but since that would be a breaking change I am not inclined to do that at this point in time.

cc @flore77 @jzaefferer @leobalter

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -206,9 +206,11 @@ Test.prototype = {
 		}
 
 		var i,
+			skipped = !!this.skip,
 			bad = 0;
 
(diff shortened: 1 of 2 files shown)


leobalter:
> ...  since that would be a breaking change I am not inclined to do that at this point in time.

+1

Legacy is hard to maintain, but it's the right thing to do.

Issue #1008, opened 2016-06-23
HTML Reporter breaks with Symbol values


leobalter:
Here is the line causing this bug: https://github.com/jquery/qunit/blob/master/reporter/html.js#L747

If I have a failing assertion, the HTML Reporter will try to call .indexOf on top of the expected value. A symbol type has no indexOf and that will break the tests.

```js
QUnit.test("different symbols", function(t) {
  var s1 = Symbol("QUnit");
  var s2 = Symbol("QUnit");
  t.equal(s1, s2);
});
```

This fix will require an extra investigation to see if it breaks further, but it's still an easy one to solve.

PR #1011, opened 2016-06-25
Dump: Add parser for Symbol

trentmwillis:
Addressing #1008. The root issue is that `dump.parse` didn't have a parser defined for object's of type `symbol` and thus it didn't return a `string`. This fixes that and adds a test to catch regressions.

Here are some examples of the fix in action. The first test is essentially `Symbol("QUnit") === Symbol("QUnit")` and the test shows `Symbol("QUnit1") === Symbol("QUnit2")`.

screen shot 2016-06-24 at 9 28 48 pm

### DIFF
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -1892,6 +1892,6 @@ QUnit[ hasES6Symbol ? "test" : "skip" ]( "regular checks", function ( assert ) {
 	assert.equal( QUnit.equiv( a, a ), true, "Same symbol is equivalent" );
 	assert.equal(
 		QUnit.equiv( a, b ), false,
-		"Not equivalent to another similar symbol built build on the same token"
+		"Not equivalent to another similar symbol built on the same token"
 	);
 } );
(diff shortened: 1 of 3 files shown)


test/main/dump.js
```diff
@@ -1,3 +1,5 @@
+/* globals Symbol:false */
+
```
leobalter: I believe we could set `esversion: 6` option on the JSHint config file. This inline setting wouldn't be necessary.

trentmwillis: Since Symbol still isn't supported by all browsers a we support, I think it's beneficial to still call it out specifically.

test/main/deepEqual.js
```diff
@@ -1892,6 +1892,6 @@ QUnit[ hasES6Symbol ? "test" : "skip" ]( "regular checks", function ( assert ) {
 	assert.equal( QUnit.equiv( a, a ), true, "Same symbol is equivalent" );
 	assert.equal(
 		QUnit.equiv( a, b ), false,
-		"Not equivalent to another similar symbol built build on the same token"
+		"Not equivalent to another similar symbol built on the same token"
```
leobalter: fixing this typo helped me confirming we have this exact test, as QUnit.equiv does not return true for different symbols with the same description.

leobalter:
This is great, thanks!

PR #1014, opened 2016-06-26
HTML Reporter: Fix apply/reset button disappearance

### LINKED ISSUE #1013: HTMLReporter: apply/reset buttons, in module dropdown, disappear when they shouldn't be
**QUnit version:** 2.0.0

**Steps to reproduce:**

Let's say I have 3 modules: `rules`, `messages` and `aria`.
* Select the module `rules` from the dropdown: the buttons apply/reset will appear (_expected_)
* Select another module (`messages` for example): the buttons still appear (_expected_)
* Uncheck the module `rules`: the buttons will disappear (_unexpected_)

Instead they should appear because the dropdown is still dirty and I didn't commit my choices yet.

See attached screen cast:
qunit-anim

shivamdixit:
Sometimes apply/reset buttons in module dropdown disappear when they shouldn't. This commit fixes the behaviour of these buttons.

Fixes #1013

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -411,7 +411,7 @@ function toolbarModuleFilter () {
 
 	// Processes selection changes
 	function selectionChange( evt ) {
-		var i,
+		var i, item,
 			checkbox = evt && evt.target || allCheckbox,
 			modulesList = dropDownList.getElementsByTagName( "input" ),
 			selectedNames = [];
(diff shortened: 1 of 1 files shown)


trentmwillis:
Looks good, would you mind caching the result of `modules[ i ]` since we use it frequently in that loop now?

shivamdixit:
@trentmwillis Thanks, updated the PR

leobalter:
Thank you!

Issue #1016, opened 2016-06-29
before + beforeEach = hang of test


vogdb:
```js
QUnit.module('Sample tests', {
  before: function() {
    return new Promise(function(resolve) {
      setTimeout(function() {
        console.log('`before` complete');
        resolve()
      }, 200);
    });
  },
  beforeEach: function() {
    return new Promise(function(resolve) {
      setTimeout(function() {
        console.log('`beforeEach` complete');
        resolve();
      }, 100);
    });
  }
});

QUnit.test('Sample test', function(assert) {
  console.log('Test started!');
  assert.ok(true);
});
```

`'Test started!'` never appears to console. Qunit version is 0.2.0. Resolving through `assert.async()` and `done` produces the same problem.

vogdb:
Does not happen on 1.23.1

trentmwillis:
@vogdb thanks for reporting, will look into it as I believe it has to do with some changes I made to handling async

vogdb:
Additional note. It worked on 1.23.1 cause `before` never was executed. It seems that there are no such hook as `before`.

PR #1017, opened 2016-06-29
Core: Improve internal async handling by removing resumed

trentmwillis:
The fixes introduced in #1002 actually wound up also causing a regression in #1016 & #1018.

After digging into it for a while, I think the solution shipped in #1002 was actually the wrong approach to take. It worked for simple cases, but for tests with multiple async hooks (as reported in #1016) it was too brittle due to race conditions of having multiple timeouts for resuming.

This approach, prevents multiple timeouts from being triggered by managing the number of async callbacks within the test. There is still the possibility of some resumes leaking when doing "waterfall" calls, but this can be worked around by simply making sure we're resuming the current test.

I added several test cases to ensure this is more robust.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -26,6 +26,7 @@ QUnit.assert = Assert.prototype = {
 			acceptCallCount = 1;
 		}
 
+		test.asyncCallCount += acceptCallCount;
 		test.semaphore += 1;
 		test.usedAsync = true;
 		pauseProcessing( test );
@@ -37,13 +38,21 @@ QUnit.assert = Assert.prototype = {
 					sourceFromStacktrace( 2 ) );
 				return;
 			}
+
 			acceptCallCount -= 1;
+			test.asyncCallCount -= 1;
+
 			if ( acceptCallCount > 0 ) {
 				return;
 			}
 
 			test.semaphore -= 1;
 			popped = true;
+
+			if ( test.asyncCallCount > 0 ) {
+				return;
+			}
+
 			resumeProcessing( test );
 		};
 	},
--- a/src/core.js
+++ b/src/core.js
@@ -230,17 +230,21 @@ function resumeProcessing( test ) {
 	// A slight delay to allow this iteration of the event loop to finish (more assertions, etc.)
 	if ( defined.setTimeout ) {
 		setTimeout( function() {
-			var current = test || config.current;
-			if ( current && ( current.semaphore > 0 || current.resumed ) ) {
+			var current;
+
+			// Ensure we are resuming the current test
+			if ( test && test === config.current ) {
+				current = test;
+			} else {
+				current = config.current;
+			}
+
+			if ( current && ( current.semaphore > 0 ) ) {
 				return;
 			}
 
 			if ( config.timeout ) {
 				clearTimeout( config.timeout );
-			}
-
-			if ( current ) {
-				current.resumed = true;
 			}
 
 			begin();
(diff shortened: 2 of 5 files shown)


trentmwillis:
@leobalter I think this is good to go. I am wondering at this point if it might be worthwhile to evaluate the async patterns we support since some of the code here feels really brittle. In particular the idea of "waterfall" calling async seems bad to me.

src/assert.js
```diff
@@ -26,6 +26,7 @@ QUnit.assert = Assert.prototype = {
 			acceptCallCount = 1;
 		}
 
+		test.asyncCallCount += acceptCallCount;
```
gibson042: What does this new property provide that isn't covered by `semaphore`?

src/core.js
```diff
@@ -230,19 +230,23 @@ function resumeProcessing( test ) {
...
-			var current = test || config.current;
-			if ( current && ( current.semaphore > 0 || current.resumed ) ) {
+			var current;
+
+			// Ensure we are resuming the current test
+			if ( test && test === config.current ) {
+				current = test;
+			} else {
+				current = config.current;
+			}
```
gibson042: I shudder a little every time I look at this code. You're absolutely right to call it brittle.

src/core.js
```diff
@@ -230,19 +230,23 @@ function resumeProcessing( test ) {
 	// A slight delay to allow this iteration of the event loop to finish (more assertions, etc.)
 	if ( defined.setTimeout ) {
 		setTimeout( function() {
-			var current = test || config.current;
-			if ( current && ( current.semaphore > 0 || current.resumed ) ) {
+			var current;
+
+			// Ensure we are resuming the current test
+			if ( test && test === config.current ) {
+				current = test;
```
gibson042: If I understand correctly, the problem is that we were using `resumed` to mark progression from one asynchronous phase to another (when `semaphore` dips to zero), but treating it as a _test_-level property. So rather than replacing the `test` passed in to `resumeProcessing` here, couldn't we either make `resumed` phase-aware, or explicitly clear it in the synchronized callbacks?

Then again, this feels like a band-aid on that synchronization anyway, so meh. We really should get to the deeper issue within the next couple of months.

src/assert.js
trentmwillis: Since `semaphore` could potentially be modified elsewhere, this helps us keep track of the number of `assert.async` calls. Unsure if that is a legitimate concern or not, as it was difficult for me to follow all the paths where `semaphore` might be modified.

src/core.js
trentmwillis: That was my first assumption as well. This falls apart however when you have waterfall calls that resolve close together. Consider:

```js
test('waterfall', function(assert) {
  let done1 = assert.async();
  setTimeout(() => {
    done1();
    let done2 = assert.async();
    setTimeout(done2);
  });
});
```

The resolution order can potentially go like this:
1. `done1` --> `resumeProcessing` --> sechdule `setTimeout` for 13ms
2. `setTimeout` in `test` --> `done2` --> `resumeProcessing` --> schedule `setTimeout` for 13ms
3. `setTimeout` from `done1` triggers and resumes processing
4. `resumed` is cleared as next `test` begins
5. `setTimeout` from `done2` triggers and attempts to resume processing since it doesn't know that it has already been `resumed`

trentmwillis:
Closing in favor of #1020.

Issue #990, opened 2016-04-20
How should global error handlers resume the test runner in 2.0?


platinumazure:
We're using QUnit and RequireJS. We've defined a `require.onError` handler as follows:

``` js
define([], function () {
    require.onError = function (e) {
        if (QUnit && QUnit.config) {
            if (QUnit.config.current) {
                var currentTest = QUnit.config.current;

                currentTest.pushFailure("Died on test #" + (currentTest.assertions.length + 1) + " " +
                    currentTest.stack + ": " + (e.message || e));
            }

            // Restart the tests if they're blocking
            if (QUnit.config.blocking) {
                QUnit.start();
            }
        }
    };
});
```

This is heavily inspired by how the global error handler used to handle these issues. However, `QUnit.start()` itself now throws an error (in 2.0.0-rc1), so I need a new way to resume the test runner. What is the best way to do that?

platinumazure:
For what it's worth, I would be okay with a new API, maybe something like `Test#pushFailureAndResume()`, which couples the semaphore decrement with the error being pushed. That way people can only use it to add a global failure, rather than skirting the `assert.async()` mechanism.

leobalter:
that's not async control, that's an error control we didn't have QUnit 1.x designed for it. It just worked in a way allowing the leakage from inside and outside a test context.

On the previous version, QUnit.start could be used to control any context of any test from anywhere, even from another test.

This resume method is a feature we never had properly designed on QUnit and we need to understand what is cause the blocking state to better handle it. I guess the promise-able tests might help finding a good path, but I would like to leave this issue open to discuss the best approach before shipping anything we can't remove until another major version.

platinumazure:
Fine by me.

For what it's worth, promise-able tests won't work in the RequireJS case because RequireJS does not return a promise, and its callback functions are executed asynchronously (here meaning not on the same call stack). My `require.onError` function is similarly unable to support the injection of a local test context in order to somehow return a promise.

Please let me know if I can provide further information to aid in this discussion.

platinumazure:
Oh, you said you wanted to know what causes the blocking state. That's an easy one:

```js
QUnit.test("This is a test", function (assert) {
    var done = assert.async();     // Causes config.blocking to be set to true, as it should

    require(["some/dependency"], function (dependency) {
        // This callback is invoked asynchronously. Do stuff as normal.

        assert.ok(dependency);

        done();
    });
});
```

[...]

PR #1020, opened 2016-07-08
Core: Improve internal async handling

gibson042:
Fixes gh-1016
Ref 1f00d5fa3b7064bdecec7b984fe9c15c38df2113

@trentmwillis: Sorry for essentially redoing gh-1017, but investigation of https://github.com/jquery/qunit/pull/1017/files#r69368544 got out of hand and I ended up doing a large chunk of the async cleanup we talked about. Let's call it a half rewrite, but it does resolve gh-1016 without introducing new properties (and in fact still removes `resumed`).

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -26,7 +26,6 @@ QUnit.assert = Assert.prototype = {
 			acceptCallCount = 1;
 		}
 
-		test.asyncCallCount += acceptCallCount;
 		test.semaphore += 1;
 		test.usedAsync = true;
 		pauseProcessing( test );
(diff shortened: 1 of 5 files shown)


leobalter:
I'm fine with the overall changes. It's great to see this is getting a substantial fix and a collaborative work. :)

Waiting for @trentmwillis.

trentmwillis:
Excited to see this, will try to do an actual review tonight.

trentmwillis:
Looks good to me. Just to make sure I followed everything correctly: async control is now handled by `internalStop()` and later calling `resume()` except in recovery cases where we need to call `internalRecover()`. There shouldn't be any need to call `begin()` or `internalStart()` directly, right?

Also, would we want to consider exposing `recover` to help with stuff like #990?

gibson042:
> Looks good to me, just a couple questions real quick. Just to make sure I followed everything correctly: async control is now handled by `internalStop()` and later calling `resume()` except in recovery cases where we need to call `internalRecover()`. There shouldn't be any need to call `begin()` or `internalStart()` directly, right?

Correct.

> Also, would we want to consider exposing `recover` to help with stuff like #990?

Possibly. The public function would do something like `internalRecover( config.current )`.

trentmwillis:
Merged. @leobalter can you release a patch version?

Issue #707, opened 2014-12-04
Discussion: Move to es6 modules to build QUnit


jzaefferer:
We should consider using es6 modules to build QUnit, moving from source concatenation to a standard format. The biggest issue with transpiling to es6 are dependencies, but since QUnit bundles its dependencies anyway, we don't have to care about that. There shouldn't be a difference for users either, since the built files would look mostly the same as they do now.  If that works out, we could also use other es6 features in the QUnit source. 

Currently our tests don't need a build step, which we may want to keep. If compilation is fast enough, we could move those to es6 as well, if there are obvious benefits.

Some background and examples: https://medium.com/@brianleroux/es6-modules-amd-and-commonjs-c1acefbe6fc0 (which mostly uses a transpiler *and* browserify, not exactly what we'd want)

jzaefferer:
This should use https://www.npmjs.com/package/rollup for the bundling.

JamesMGreene:
Why not just switch wholesale to Babel?

mzgol:
There's still a question of bundling. Babel can transpile ES6 modules to SystemJS, AMD or CommonJS but you still need to somehow bundle it all up and it's better to choose a method that doesn't introduce a lot of overhead. rollup seems to gain momentum now.

mikesherov:
At Behance, we use es6, with karma, Babel, and webpack to run our test suite. 

Not sure if QUnit is architected in a way to allow preloaders for code, but perhaps that's where you start: 

1. building infra that allows QUnit to use webpack + Babel or even Babel directly when loading test suites.
1. Use that infra to begin authoring QUnit and its test suite in ES6.
1. Release both.

mikesherov:
Why rollup over webpack?

mikesherov:
FYI, webpack has a decent comparison chart of all the different options: https://webpack.github.io/docs/comparison.html

They're the only one who followed @scottgonzalez's recommendation that "if you're going to exist, have a comparison chart explaining why you're existence is necessary".

trentmwillis:
I'm planning to pursue actually doing this. After looking at webpack and rollup (and some others), I'm going to explore using rollup as I think it most closely fits our current setup.

It looks like it'll probably be a large amount of work to get everything setup to work as modules because even though `src` is split up into files, they don't really act like modules with boundaries right now.

Issue #1024, opened 2016-07-26
Introduce CLI tool


trentmwillis:
Lately, we've been doing a lot of work on the reporter interface for QUnit and I think that once we start shipping a standard TAP (or other) reporter, it'll lower the barrier to adoption of QUnit for non-browser projects (specifically thinking about Node projects).

However, I think one thing that would complement that effort and make it trivial to use QUnit in those cases would be to introduce a command-line tool similar to what Mocha provides. In particular, this tool should make running QUnit tests in Node easy without having to provide a custom script.

There are some tools currently available (such as node-qunit or even mocha-qunit-ui), but it would be great to have one backed by the project that doesn't require additional overhead from the project maintainers.

As an aside: I would also very much like to see this happen so that we can unify the testing story in the Ember community, where most developers use QUnit in the browser for applications, but Mocha in node for build tools or other things.

jzaefferer:
Based on previous conversations with @leobalter, I think he's very much on board with this. I like it in general as well. I'd like to make sure we get a full picture of everything involved though.

A CLI tool needs:

1. A CLI API to specify files to run, with some reasonable default (look up files in `test/*` or something)
2. Direct support for file watching and rerunning tests, or good third-party integration, like `nodemon`
3. Multiple reporters like tap, JUnit-style XML, whatever else there is

I like how `ember test --server` does 1. and 2., while the reporters are rather bad (at least in `ember-cli: 2.4.3`), especially the failure output is hard to parse.

Speaking of ember, would ember-cli switch from testem to a native QUnit CLI?

I'm sure I'm still missing a lot here...

trentmwillis:
> I'd like to make sure we get a full picture of everything involved though.

Most definitely. I think what you've outlined above is a solid candidate for a minimum viable product and what most users would expect. I'll try to put together something a bit more formal that outlines our expected API and usage behavior/constraints.

> Would ember-cli switch from testem to a native QUnit CLI?

Doubtful as I imagine we won't want to support running/interfacing with browsers, something that Testem does well and is required for web apps. Rather, I imagine that you could set up QUnit as a "launcher" for Testem in the event that you have code that can be tested in Node ([see here for an example for Mocha](https://github.com/testem/testem#running-tests-in-node-and-custom-process-launchers)).

jzaefferer:
> I'll try to put together something a bit more formal that outlines our expected API and usage behavior/constraints.

Sounds good. Maybe you could also do some research on existing modules that we can use?

[...]

PR #1025, opened 2016-07-30
All: Use Rollup and Babel to build

trentmwillis:
Addressing #707. This will enable us to use ES6 via Babel and build our application via Rollup, meaning no more reliance on specified concat order.

Some interesting things to note:

* The code in `src` should mostly be free from `QUnit` expressions, other than in `src/core` and `src/export` which is where the definition and export of the global `QUnit` object occurs.

* JSHint now runs on the individual source files and is more strict to ensure that we are explicit about dependencies.

* Anything in `reporter`/`runner` does not depend on any code in `src` other than `src/core` which exports `QUnit` and `src/globals` which provides access to global objects.

* There are some files in `reporter`/`runner` that have their contents wrapped in an IIFE to allow them to return early. These files should be converted to export functions that execute similar code. This will eventually make it easy to ship QUnit separately from the `runner` and `reporter` which are specific to the browser.

Improvements for future PRs:

* Remove all IIFEs, these shouldn't be needed anymore

* Breakup some of the larger modules to simplify dependency graph (e.g., `src/test` and `src/html`)

* Move all source code into `src`. This might be contentious, but I'd like to see a structure similar to the following:

   ```
   qunit/
      src/
        qunit.js
        qunit.css
        core/
          index.js
        reporter/
          index.js
        runner/
          index.js
```

   The `qunit.js` file will import `core/index.js`, `reporter/index.js`, and `runner/index.js` and those will then be responsible to act as entry points for the functionality encompassed within. `index.js` could also be something like `main.js` or `exports.js`. The idea here is to have just one directory for our source code that can also mirror how we'd like to think about our dependency graph.

### CODE REVIEW

**Comment by trentmwillis** [comment]
Test failure is due to Node 0.10 not having `Map` support 😢  will fix in a bit

**Comment by leobalter** [comment]
I want to prioritize this as we're in a moment we can release a new version without any new features in the API or bug fixes.

This will give us some time to check everything.

File: Gruntfile.js [hunk 1]
```diff
@@ -28,42 +22,31 @@ grunt.initConfig( {
...
-				"src/outro.js",
-				"runner/fixture.js",
-				"reporter/intro.js",
-				"reporter/urlparams.js",
-				"reporter/html.js",
-				"reporter/diff.js",
-				"reporter/outro.js"
-			],
+			src: "dist/qunit.js",
 			dest: "dist/qunit.js"
```
leobalter: it seems we only need this for the process part. I'm ok to keep it for now, but we should replace this for anything more appropriate in a future. If we merge it this way, let's remind to open an issue to address this.

rollup.config.js
```diff
@@ -0,0 +1,35 @@
...
+	banner: "\
+/*!\n\
+ * QUnit @VERSION\n\
+ * https://qunitjs.com/\n\
+ *\n\
+ * Copyright jQuery Foundation and other contributors\n\
+ * Released under the MIT license\n\
+ * https://jquery.org/license\n\
+ *\n\
+ * Date: @DATE\n\
```
leobalter: we basically need to "process" @VERSION and @DATE, which can be easily done here. That replaces the need for the grunt concat on the js file, I guess.

leobalter:
> Test failure is due to Node 0.10 not having Map support :cry: will fix in a bit

With this and the comments I've mentioned above getting fixed I believe we'll be good to roll.

rollup.config.js
trentmwillis: Not quite, there's also [one instance in `src/core`](https://github.com/jquery/qunit/blob/b7a00ace99f60a1073a7fcf47312cc80edc556ca/src/core.js#L5), which is why I didn't already drop the concat step.

Gruntfile.js
trentmwillis: This is correct. I'll look into a replacement, but since we need to do it for both `qunit.js` and `qunit.css` I don't think this is too bad if we wind up leaving it.

trentmwillis:
So apparently [Rollup doesn't support Node 0.10](https://github.com/rollup/rollup/blob/master/.travis.yml#L4-L6). Even running with the `--harmony` flag doesn't seem to work properly. It also looks like [Webpack only supports `>0.12` versions of Node](https://github.com/webpack/webpack/blob/master/.travis.yml#L7-L10).

This only really affects the build pipeline. So maybe we can revamp Travis to build using a supported version and then use `nvm` to change to `0.10` to test against? Ideally I'd like to just drop `0.10` of Node, but [LTS doesn't end until October](https://github.com/nodejs/LTS).

Would like your thoughts @leobalter

leobalter:
I'm ok to drop support for 0.10 in this case if that won't represent any problem for other projects using QUnit. This might require a new major version which we could attach the PR dropping QUnit.push so we something else for a new 3.0.0 version.

trentmwillis:
I think we should definitely slate `0.10` (and probably `0.12` as well) for dropping in QUnit 3.0, but doing that now seems a bit pre-mature since QUnit 2.0 is still new.

Instead I updated the Travis configuration to build against a recent version of Node but to still test against our supported versions. We shouldn't need to support all the same versions for our development/build environment as we do for the distributed version of QUnit.

mgol:
> Instead I updated the Travis configuration to build against a recent version of Node but to still test against our supported versions.

Note that this means you can't have any native dependencies or devDependencies as then you can't install packages with one Node & run them with another one. You're also now using a different npm version to install packages and a different one to run `npm test`. This seems risky but at the same time I don't see another easy way: you'd basically have to `npm install` with a new Node, build QUnit, switch to the old Node, re-install all dependencies but retain the build QUnit files and then run the tests.

trentmwillis:
> you can't have any native dependencies or devDependencies

By native, you mean global? Regardless, I don't foresee this being a problem, but good to keep in mind.

> you'd basically have to npm install with a new Node, build QUnit, switch to the old Node, re-install all dependencies but retain the built QUnit files and then run the tests.

I would actually be okay with this, _except_ running `npm install` also runs `prepublish` which would attempt to re-build the files causing an error and I don't see a good way around that.

jzaefferer:
Having both `src/qunit.js` and `src/core/index.js` seems strange. Otherwise I like the suggestion to move all source files into a `src` folder.

As for dropping node 0.10, I wouldn't worry too much about it. Whoever is still on 0.10 is probably not going to touch a (current) QUnit release. October isn't far away, might as well drop that immediately. Also, http://qunitjs.com/ only mentions browser support, not node.

leobalter:
> Whoever is still on 0.10 is probably not going to touch a (current) QUnit release

The only issue here is that will require a major release (by dropping support) and @trentmwillis pointed it 2.0 has got released recently.

What we can do is land this without a release, focus on the cli (#1024) with a built in reporter and release 3.0.0 when we get all of them done.

Patches (2.0.x) might still have progress for bug fixes, but we would stall QUnit for any minor version (2.1.0+).

This strategy will require us to avoid much further breaking changes, like removing something from the API, allowing us to remote the only method we already have deprecated, .push

---

If this plan works, we can expect a new major version being released around October, otherwise, we need keep the restraints with 0.10.

jzaefferer:
I don't understand how dropping 0.10 in our build pipeline is equivalent to a major, breaking change. A few more thoughts there:

* Did we ever officially support 0.10? If not, we're not dropping support
* Is there any actual compat issue with 0.10 with the changes we're discussing? It seems like dropping 0.10 on Travis makes it easier to build/test on all other platforms, but that's it
* Holding off on releases to sprint to 3.0 sounds like a bad idea. I'd like to see js-reporters land as part of 2.x, and that takes time
* Patch releases for old versions can also eat a lot of time. Better to be able to always release from master.

leobalter:
@gibson042, can you give your feedback on this one?

From the main projects I take in consideration for support, we have lodash, which [still supports node 0.10](https://lodash.com/#support). This might be a PITA support to keep, but I would also prefer having more feedback from @jdalton.

jdalton:
I'm cool with whatever you all need to do. Lodash will drop Node 0.10 support in v5 (Jan 2017). I only ever use the QUnit dist build though so changes to your build requirement don't effect my use.

leobalter:
Ok, the way @trentmwillis fixed the CI process, it looks good to me landing this PR.

> Having both src/qunit.js and src/core/index.js seems strange. Otherwise I like the suggestion to move all source files into a src folder.

I like the `src/qunit.js` as the starting point. We can move the `src/core/index.js` to a better naming in a follow up.

trentmwillis:
> Is there any actual compat issue with 0.10 with the changes we're discussing? It seems like dropping 0.10 on Travis makes it easier to build/test on all other platforms, but that's it

@jzaefferer exactly. This "drop" was primarily talking about our build/testing processes. The distributed library itself should still work in Node 0.10.

@leobalter I would like to hammer out what we want our Node support policy to be. We currently have a "rolling" support policy for browsers and I tend to think we should probably adopt something similar with Node. Ex: we support Node LTS releases, but any QUnit released after LTS ends is not guaranteed to be compatible (similar to how we don't test against all versions of Chrome since 2.x was released, we only care about the 2 most recent). That way, when 0.10 LTS ends in October we can drop it without having to do a major version bump.

As for the file structure, we can debate that when I open the PR to make the change.

leobalter:
surprisingly we have no specific support mentioning Node 0.10. We can't say we actually had any support as some previous versions of QUnit didn't work on Node at all.

Similar to the browsers I want to say we should support the last 2 LTS versions if more than one is active. That would target us to Node 4 and 6. At some point this will make more sense.
 
The alternative is to say we support the current Node LTS versions from the time of each release.

mgol:
> Similar to the browsers I want to say we should support the last 2 LTS versions if more than one is active. That would target us to Node 4 and 6.

Currently that would target 0.12 & 4, not 4 & 6. Node 6 is not an LTS, it will become an LTS at the time of Node 7 release.

https://github.com/nodejs/LTS

platinumazure:
There's only one LTS and one stable at a time, right? That's two versions, similar to Chrome's last two versions support. So Node 4 and 6, right?

I don't know if we really need to support two LTS versions plus a stable, do we? (Honestly not sure, not trying to persuade folks.)

mgol:
@platinumazure No, there are more supported versions. Just click the link I provided, it's all explained there.

rollup.config.js
```diff
@@ -0,0 +1,35 @@
...
+	format: "iife",
+	exports: "none",
+	plugins: [
+		babel( {
+			presets: [ "es2015-rollup" ]
+		} )
+	],
+
+	// jscs:disable disallowMultipleLineStrings
+	banner: "\
```
gibson042: This would look better as a `` `…` `` template literal.

src/globals.js
```diff
@@ -0,0 +1,9 @@
+import global from "global";
+
+export const window = global.window;
+export const console = global.console;
+export const document = global.document;
```
gibson042: I think this should be `window && window.document`, since `window` is not necessarily `global`. Likewise for `navigator` and probably `sessionStorage`.

gibson042:
A couple of nits, but on the whole this looks great! :clap:

trentmwillis:
Updated to address @gibson042's comments.

I think we should land this. Since we have no official Node support policy as of this moment, I don't see a need to do a major version bump. Additionally, we're only dropping Node 0.10 from our build environment which should not concern our consumers.

I do think we should figure out an official policy for Node support, but we should likely do that in a separate issue.

trentmwillis:
Woops, looks like Node 0.10 and 0.12 don't like the template literal in the rollup config. Changing it back...

jdalton:
Curious why the choice to go with Rollup? I've found Webpack to be [super flexible](https://github.com/lodash/lodash-webpack-plugin) for builds.

trentmwillis:
@jdalton honestly, we could have gone with either. After looking at both Webpack and Rollup, they both support everything that we need. I wound up choosing Rollup because it seemed to more closely align with the architecture of QUnit, in that we only have one real module we distribute. We hope to change that in the future so that Node users don't need all the Browser-specific code, but even then we're still only likely to have a small handful of modules. Rollup also seemed a bit easier to get up and running.

At the end of the day, the primary goal is to support ES6 and start using a standard module syntax, so if needed we could switch to Webpack (or some other tool) in the future.

gibson042:
I have no objection to landing this.

trentmwillis:
@leobalter would like your :+1: before landing.

leobalter:
👍 let's do this!

leobalter:
For a heads up on the rollup/webpack, I agree with the answer from @trentmwillis. I believe it should not be any problem to switch it later. I have some experience with webpack but I'm interested to try rollup for now.

trentmwillis:
Merged, but build error'd. Not sure why, investigating now...

trentmwillis:
Update: there was a back-compat issue in Babel that messed up the preset for es2015. Find out more in this issue: https://github.com/rollup/rollup-plugin-babel/issues/72

Gotta hang tight until a new release is cut. At which point I'll kick the build again.

PR #1026, opened 2016-08-02
Adopt js reporters standard

flore77:


platinumazure:
Is it just me or do you have conflict markers?

leobalter:
is this only reusing old commits from #882?

flore77:
This wanted to be a pr only my fork, but I made the pr into the upstream...anyway this should have arrived later, so I guess it is ok. 

@platinumazure yes there were, I made the rebase multiple times.

> is this only reusing old commits from #882?
@leobalter yes

I will ping back when I will manged to get it properly done, i.e tests will pass, or if I will get stuck.

flore77:
The rebase is now complete, there is still work to do, but I think it is a good start :smiley:

jzaefferer:
I still think this needs to wait for #1025 to land, which is pretty close.

flore77:
Ok.

In the meanwhile what would be the plan, to deprecate the current API, like adding an warning in 2.0, then implement the standard and to make the 3.0 release? Or to implement the API along with the old one and release a 2.x version and then on 3.0 to get rid of the old one?

leobalter:
I'm a +1 for deprecation, but I would wait for a minimal number of reporters to adapt for the new api, including the HTML Reporter (which might be covered here but I'm writing this in the go).

The new API can be released in a 2.x version, we avoid releasing new features in a major version, just features removals.

We should not worry about the warnings message in this PR, let it be done in a follow up step so we can have an easy process for each part.

flore77:
Sounds good.

reporter/html.js
```diff
@@ -580,30 +580,30 @@ function appendTest( name, testId, moduleName ) {
...
-
-	// Initialize QUnit elements
-	appendInterface();
-	appendTestsList( details.modules );
-	tests = id( "qunit-tests" );
-	if ( tests && config.hidepassed ) {
-		addClass( tests, "hidepass" );
-	}
+QUnit.on( "runStart", function( details ) {
+  var i, moduleObj, tests;
```
jzaefferer: Looks like the indentation got messed up. `QUnit` uses tabs, not spaces.

trentmwillis:
+1 to what @leobalter said. Once this has landed I'd also like to see us introduce a non-browser reporter with the EventEmitter API (e.g., a TAP reporter).

leobalter:
I'm planning to rebase the tap reporter on top of this.

I'll do a proper review for this PR this week. Meanwhile, can you help us fixing the CLA issue? Maybe you'll just need to sign it and the problem is solved.

flore77:
Yes, I will sign the CLA. I will also rebase this again since #1025  has now landed into master.

build/tasks/test-on-node.js
```diff
@@ -59,24 +59,24 @@ module.exports = function( grunt ) {
...
-		QUnit.begin( function() {
+		QUnit.on( "runStart", function() {
 			grunt.log.ok( "Testing " + file + " ..." );
 		} );
-		QUnit.testStart( function() {
+		QUnit.on( "testStart", function() {
 			testActive = true;
 		} );
-		QUnit.log( function( details ) {
+		QUnit.on( "assert", function( details ) {
```
jzaefferer: This is kinda interesting. js-reporters doesn't define an `assert` event. To render individual assertions in progress, its necessary.

I think its fine for QUnit to add another event, but the object should follow the spec for `Assertion`.

reporter/html.js
jzaefferer: In other words, needs to convert from spaces back to tabs here.

src/core/logging.js
```diff
@@ -1,43 +1,105 @@
...
-		callbackNames = [ "begin", "done", "log", "testStart", "testDone",
+// DEPRECATED: Register old logging callbacks
+// This will be removed on QUnit 2.0.0+
+// From now on use the QUnit.on( eventType, callback ) format
+export function registerLoggingCallbacks( QUnit ) {
+	var i, l, key;
+	var callbackNames = [ "begin", "done", "log", "testStart", "testDone",
 			"moduleStart", "moduleDone" ];
+	var dictionary = {
+			"begin": "runStart",
```
jzaefferer: Looks like there's too much indent here.

jzaefferer:
This currently only adopts the event names, not the event properties, right? I think I mentioned this before: Going from old events to new events, but with old properties is not a good approach. New events need to have new event properties. Makes backwards-compat harder (need old events with old props + new events with new props), but that seems the only option to me.

build/tasks/test-on-node.js
flore77: Personally, I don't find it a good idea, because then everyone could add something. make a little change etc.

Also that's why we have introduced the test `assertions` property on the `testEnd` event , which is specifically designed for this.

reporter/html.js
flore77: Yes, indeed. I need to change my editor config to meet QUnit coding style.

flore77:
> This currently only adopts the event names, not the event properties, right?

@jzaefferer yes

> need old events with old props + new events with new props

@jzaefferer I will try that, also now from the rebase on top of es6-modules, it requires some more refactoring, so I will try to make all work

@leobalter @trentmwillis could we use the the [TAP reporter](https://github.com/js-reporters/js-reporters/blob/master/lib/reporters/TapReporter.js) from the js-reporters repo? If it requires some changes I will happily do it.

flore77:
Also I signed the CLA, and I am pretty sure that it matches the name and email from git config. What should I do? Sign it again ?

platinumazure:
@flore77 The name that shows up on the commit is "Florentin", which is one word and which will trigger the CLA bot to fail the check and require "manual verification". I don't know if that means one of the other QUnit folks can validate the CLA anyway, or if you need to change your Git config to have first/last name, `git commit --amend --reset-author`, and then force push.

flore77:
@platinumazure shouldn't I change the author for all commits, not only for the last one ?

platinumazure:
Yes... You'll want to search on Stack Overflow, there's a way to iterate
over all relevant commits with a shell script and reprocess them en masse
but I can never remember how to do it.

flore77:
Ok. Thanks!

flore77:
@platinumazure thanks. I sorted it out, but without iterating, there were only a few commits so I have done a rebase and edit each commit.

leobalter:
> could we use the the TAP reporter from the js-reporters repo?

Yes. 

My Tap reporter PR is stalled as I am busy with everything + I was waiting for 2.0 + ES Modules.

In any case, I like small steps for PRs, so let's finish this one and set the TAP reporter in a new PR, that will allow more focused reviews.

flore77:
Guys can you help me in adding `js-reporters` as a dependency? 

I have tried to concat the sources (what is currently in the pr), locally the `test-on-node` task is working, but on travis and appveyor is not.

I have tried to add it through the rollup process, with help of [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve), but I am receiving:
image

I am not sure how to add it, because also js-reporters is bundled with rollup.

jzaefferer:
Can we import js-reporters source files instead of the bundled files? Then rollup can build the final package.

src/core.js
```diff
@@ -8,9 +8,12 @@ import exportQUnit from "./export";
 
 import config from "./core/config";
 import { defined, extend, objectType, is, now } from "./core/utilities";
-import { registerLoggingCallbacks, runLoggingCallbacks } from "./core/logging";
+import { on, emit, registerLoggingCallbacks,
+	runLoggingCallbacks } from "./core/logging";
 import { sourceFromStacktrace } from "./core/stacktrace";
 
+import jsReporters from "js-reporters";
```
jzaefferer: Can we import individual modules directly? Importing the whole thing is likely to include all the adapters in the QUnit bundle, which we don't want.

src/core/config.js
```diff
@@ -1,3 +1,5 @@
+import jsReporters from "js-reporters";
```
jzaefferer: Can we do `import { Suite } from "js-reporters"` here? Same reason as above.

flore77:
Fixed it :tada:

src/core.js
flore77: We must do some changes in `js-reporters` for this.

src/core.js
```diff
@@ -8,9 +8,12 @@ import exportQUnit from "./export";
 
 import config from "./core/config";
 import { defined, extend, objectType, is, now } from "./core/utilities";
-import { registerLoggingCallbacks, runLoggingCallbacks } from "./core/logging";
+import { on, emit, registerLoggingCallbacks,
+	runLoggingCallbacks } from "./core/logging";
 import { sourceFromStacktrace } from "./core/stacktrace";
 
+import {Suite} from "js-reporters/lib/Data";
```
jzaefferer: 👍

src/core.js
```diff
@@ -106,6 +133,8 @@ extend( QUnit, {
 
 	only: only,
 
+	on: on,
```
jzaefferer: Since we can now use ES6 (right?), this can be shortened to just `on`,

src/test.js
```diff
@@ -331,6 +377,15 @@ Test.prototype = {
 
 		runLoggingCallbacks( "log", details );
 
+		this.jsRepTest.assertions.push( new jsRepAssertion(
+				details.result,
+				details.actual,
+				details.expected,
+				details.message,
+				undefined
```
jzaefferer: `details.source`?

jzaefferer:
If I'm reading this correctly, this now leaves the existing callback completely alone, and adds the event emitter as its own (sub)system. Is that right? In 3.0 we'd then remove the old system?

flore77:
>  this now leaves the existing callback completely alone, and adds the event emitter as its own (sub)system.

@jzaefferer  Yes.

I added [tests](https://github.com/flore77/qunit/blob/79e4397a716695a6fca2d862fd0a5d25c5a8532c/test/events-logs.js) to test only that the events are emitted and the order is correct (it will not always be source order, it depends on how QUnit executes the tests, but this can be treated in another pr).

Also about this tests can you explain me the [setTimeout](https://github.com/flore77/qunit/blob/79e4397a716695a6fca2d862fd0a5d25c5a8532c/test/events-logs.js#L38), I have seen this is done also in testing the [old logging callbacks](https://github.com/flore77/qunit/blob/master/test/logs.js#L300). The timeout is also to big, this tests didn't have the chance to run. I am not sure why they are influencing the other tests if I don't put them in a timeout.

To test the emitted data I must run the tests from `js-reporters`. Any suggestion how to do it ?

I will have another look through the code (to add comments, ... etc.) and then I will require review from the whole team :smiley:

src/test.js
flore77: For passed assertions the `stack` prop is `undefined`.

src/core.js
flore77: Yes, but I would leave it like this, to be equal to the others props, this can be changed in another pr.

flore77:
I have quite finished the adoption of the events and added tests to test that they are emitted and the emitting order is corect.

One problem has arise out of the [nested suites testsuite](https://github.com/flore77/qunit/blob/315c9b33aa5ba7b679b31792545f83e2237b235a/test/events/nested-suites.js).

Here is how QUnit executes the testsuite and emits on the old callbacks:
* module:  module1 > module2  starts
* test:  test2-1  starts
* test:  test2-1  ends
* module:  module1 > module2  ends
* module:  module1 > module2 > module3  starts
* test:  test3  starts
* test:  test3  ends
* module:  module1 > module2 > module3  ends
* module:  module1 > module2  starts
* test:  test2-2  starts
* test:  test2-2  ends
* module:  module1 > module2  ends

@leobalter @trentmwillis @jzaefferer how you can observe from the above description the start and end of module `module1 > module2` is emitted twice. Is this considered correct ?

trentmwillis:
Sorry for not keeping up with this.

> the start and end of module `module1 > module2` is emitted twice. Is this considered correct ?

I don't believe so. At the least, I do not think it should be correct. The module should only start and end once, any nested modules should be a sub-set of that module, which means that their start and end is "within" the start and end of the parent module.

flore77:
>  The module should only start and end once

I agree with you.

> any nested modules should be a sub-set of that module, which means that their start and end is "within" the start and end of the parent module

I think this is due to QUnit's flat style and that nested modules were introduced later. This is how the js-reporters standard is and I think I have [achieved this](https://github.com/flore77/qunit/blob/315c9b33aa5ba7b679b31792545f83e2237b235a/src/test.js#L108-L127). So I don't think it is necessary to do it also for the old callbacks, since they will be removed.

flore77:
The implementation is almost finished, I think there should be only a little bit of work on the emitted data.

To test the emitted data it would be really good to run the [tests](https://github.com/js-reporters/js-reporters/tree/master/test/integration) from the js-reporters repo. But, I am not sure how to do this. What I have thought about is to pull out the tests from js-reporters and to build something like a cli-tool that would test any framework.

To achieve this, the tool would need 2 external things:
* a test fixture, which would mimic exactly this ones: https://github.com/js-reporters/js-reporters/tree/master/test/fixtures
* a way to run the framework against the test fixture, like this: https://github.com/js-reporters/js-reporters/blob/master/test/integration/adapters-run.js

I found that from this solution all testing frameworks can benefit, also futures ones, with only a little effort they can test their reporting and know that they are compliant to the spec.

What do you think about this? 

This is only an idea, so I am open to any other solution.

gibson042:
I'm in favor of pursuing such a tool.

flore77:
I will start soon to develop the tool, but I need a name for it, I came up with this ones:

* standard-reporter
* standard-js-reporter
* js-reporter-compliant

Which do you prefer? Or do you have other suggestions.

leobalter:
I like standard-reporter, the other names are fine too

flore77:
I also like standard-reporter, it is shorter than the others :smiley:

trentmwillis:
This has been superseded by other PRs. Thanks for the initial work in spiking this out though! It was much appreciated; I am sorry that this PR wound up stalling for so long.

Issue #1029, opened 2016-08-11
urlParams mishandles Object.prototype property collisions


gibson042:
`Object.prototype` properties are treated as prior values when processing URL parameters of matching names: https://output.jsbin.com/venicolabe?toString=string

PR #1030, opened 2016-08-11
HTML Reporter: Handle URL params named like Object.prototype properties

gibson042:
Fixes gh-1029

### DIFF
--- a/reporter/urlparams.js
+++ b/reporter/urlparams.js
@@ -1,6 +1,5 @@
 import QUnit from "../src/core";
 import { window } from "../src/globals";
-import { hasOwn } from "../src/core/utilities";
 
 ( function() {
 
@@ -74,7 +73,7 @@ QUnit.begin( function() {
 
 function getUrlParams() {
 	var i, param, name, value;
-	var urlParams = {};
+	var urlParams = Object.create( null );
 	var params = location.search.slice( 1 ).split( "&" );
 	var length = params.length;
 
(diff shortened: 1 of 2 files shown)


reporter/urlparams.js
```diff
@@ -85,7 +86,7 @@ function getUrlParams() {
 			// Allow just a key to turn on a flag, e.g., test.html?noglobals
 			value = param.length === 1 ||
 				decodeQueryParam( param.slice( 1 ).join( "=" ) ) ;
-			if ( urlParams[ name ] ) {
+			if ( hasOwn.call( urlParams, name ) ) {
```
trentmwillis: I think another possible approach would be to simply initialize the object like:

```js
var urlParams = Object.create(null);
```

Then we don't have to worry about any property collisions as there are none to inherit. Thoughts?

gibson042: It's an old compatibility habit. But since we're now on IE9+, I'll make the change.

reporter/urlparams.js
```diff
@@ -85,7 +85,7 @@ function getUrlParams() {
 			// Allow just a key to turn on a flag, e.g., test.html?noglobals
 			value = param.length === 1 ||
 				decodeQueryParam( param.slice( 1 ).join( "=" ) ) ;
-			if ( urlParams[ name ] ) {
+			if ( name in urlParams ) {
```
leobalter: We can also use `Object#hasOwnProperty` here, I'm not a big fan of the in operator and we don't need to search the property in the urlParams prototype, as it is null.

If you prefer to keep the in operator, I'm ok with it, there's nothing wrong with it.

reporter/urlparams.js
leobalter: HA! Now I just saw this comment...

Well, the Object.create(null) is great, but we can stick with the hasOwn :)

leobalter: btw, spec wise, hasOwn will take less steps until it returns a value than the in operator. Performance wise I can't tell which is faster.

PR #1033, opened 2016-08-23
HTML Reporter: Fix expanding failed tests when collapse is false.

yagni:
Looks like QUnit.config.collapse = false stopped working with commit 5333de89dda55691e1634081e6fe97033185f836. This fixes that regression.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -816,14 +816,16 @@ QUnit.testDone( function( details ) {
 
 		// Collapse the passing tests
 		addClass( assertList, "qunit-collapsed" );
-	} else if ( bad && config.collapse && !collapseNext ) {
+	} else if ( config.collapse ) {
+		if ( !collapseNext ) {
 
-		// Skip collapsing the first failing test
-		collapseNext = true;
-	} else {
+			// Skip collapsing the first failing test
+			collapseNext = true;
+		} else {
 
-		// Collapse remaining tests
-		addClass( assertList, "qunit-collapsed" );
+			// Collapse remaining tests
+			addClass( assertList, "qunit-collapsed" );
+		}
 	}
 
 	// The testItem.firstChild is the test name


platinumazure:
Can this be unit-tested at all?

leobalter:
I fixed the commit reference (GH needs 7 chars to create the link, it was missing one)

leobalter:
> Can this be unit-tested at all?

The short answer is yes, but in general we lack some tests for the HTML Reporter so that's a work for structuring good tests for all the features.

---

LGTM, but I would like @trentmwillis to sign this off with another +1 as he has a current PR on the same lines (#1006). In anyway, this still seems fine.

leobalter:
@yagni ping me if you need any help with the CLA

yagni:
@leobalter CLA seems to need manual verification because I'm using yagni as a name. I guess I'll watch for an email?

leobalter:
Thanks @yagni, I think this should be ok.

@jzaefferer can you help me with this? Also, am I able to do this manual verification? how?

trentmwillis:
:+1: looks good to me. Should be okay to merge once the CLA is done; I'll rebase my other PR.

> Can this be unit-tested at all?

We need a better testing story around the HTML reporter. I have thoughts on that, but those should probably be saved for another time as it is beyond the scope of this PR.

leobalter:
@dmethvin ping. I need help on this manual verification.

leobalter:
@yagni [there was a change to verify](https://github.com/jquery/jquery-license/commit/582e274b1141d2d9063f71ff419ab2fbc6f5180a) just for a name, without the min 2 words.

In order to re-check this, would you mind pushing a rebased commit? This way we can confirm the CLA.

leobalter:
Oh, @jzaefferer has got this! Thanks!

Issue #928, opened 2016-02-03
Add a Stop Execution button up top


tundal45:
This might be just me but I feel that there are times where you want to run all the tests until their is a particular type of failure that you want to focus on. However, getting to that error to click rerun while the tests are running is impossible. So I was wondering if it would be useful to have a button that stops test execution and retains the results view up until that point.
1. Is this possible?
2. Will people who are not me find this useful?
3. Where should I begin (assuming 1 & to are affirmative)?

trentmwillis:
In my personal experience, when dealing with a large test suite this is definitely desirable. Especially when dealing with debugging a flaky test that is dependent on test order.

I know Ember currently supports this with a [`pauseTest` helper that returns a non-resolving promise](http://emberjs.com/api/classes/Ember.Test.html#method_pauseTest), but this only works when invoked from within a test. I don't think it would be difficult to implement a similar feature that is invoked from the UI.

That said, we'd need to establish a more rigorous definition of what this behavior is: do we pause immediately or do we wait for the current test and associated hooks to finish? Do we allow tests to be unpaused? Will this pause count against the test suite's timeout limit?

tundal45:
> do we pause immediately or do we wait for the current test and associated hooks to finish?

For now, I would be happy with a stop execution that just stops everything and the user can re-run the test if they want to. Maybe a pause could be implemented but I was envisioning the "simpler" just stop the world with this one.

gibson042:
I could see doing this with something like `QUnit.config.queue.length = 0`, which would let the active test finish but abort all pending tests. The display elements would also need updating, e.g. to reveal the hidden `<li>` elements. PRs considered, I suppose.

trentmwillis:
> I was envisioning the "simpler" just stop the world with this one.

Seems good to me. Depending on implementation, I don't think it would be hard to extend this to "pause" in the future.

> let the active test finish but abort all pending tests.

I think this makes the most sense as far as stopping behavior is concerned.

@tundal45 would you want to take a crack at implementing?

tundal45:
@trentmwillis @gibson042 Yes, I can take a look.

rwjblue:
I was just discussing this with @raytiley earlier this week.  I think this would be a wonderfully useful feature...

trentmwillis:
@rwjblue / @raytiley PRs welcome 😃

tundal45:
[...]

PR #1040, opened 2016-09-16
HTML Reporter: Add a stop button

raytiley:
This adds a stop button that will set the queue length to 0.
This will allow the current test to finish executing,
but then finish the test run.

Fixes gh-928

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -96,13 +96,14 @@ function id( name ) {
 	return document.getElementById && document.getElementById( name );
 }
 
-function stopTests() {
-  var stopButton = id( "qunit-stop-tests-button" );
-  if ( stopButton ) {
-    stopButton.innerHTML = "Stopping";
-  }
-  QUnit.config.queue.length = 0;
-  return false;
+function abortTests() {
+	var abortButton = id( "qunit-abort-tests-button" );
+	if ( abortButton ) {
+		abortButton.disabled = true;
+		abortButton.innerHTML = "Aborting...";
+	}
+	QUnit.config.queue.length = 0;
+	return false;
 }
 
 function interceptNavigation( ev ) {
@@ -271,12 +272,12 @@ function toolbarUrlConfigContainer() {
 	return urlConfigContainer;
 }
 
-function stopTestsButton() {
-  var button = document.createElement( "button" );
-  button.id = "qunit-stop-tests-button";
-  button.innerHTML = "Stop";
-  addEvent( button, "click", stopTests );
-  return button;
+function abortTestsButton() {
+	var button = document.createElement( "button" );
+	button.id = "qunit-abort-tests-button";
+	button.innerHTML = "Abort";
+	addEvent( button, "click", abortTests );
+	return button;
 }
 
 function toolbarLooseFilter() {
@@ -463,7 +464,6 @@ function appendToolbar() {
 		toolbar.appendChild( toolbarUrlConfigContainer() );
 		toolbar.appendChild( toolbarModuleFilter() );
 		toolbar.appendChild( toolbarLooseFilter() );
-    toolbar.appendChild( stopTestsButton() );
 		toolbar.appendChild( document.createElement( "div" ) ).className = "clearfix";
 	}
 }
@@ -487,7 +487,8 @@ function appendBanner() {
 
 function appendTestResults() {
 	var tests = id( "qunit-tests" ),
-		result = id( "qunit-testresult" );
+		result = id( "qunit-testresult" ),
+		controls;
 
 	if ( result ) {
 		result.parentNode.removeChild( result );
@@ -499,7 +500,14 @@ function appendTestResults() {
 		result.id = "qunit-testresult";
 		result.className = "result";
 		tests.parentNode.insertBefore( result, tests );
-		result.innerHTML = "Running...<br />&#160;";
+		result.innerHTML = "<div id=\"qunit-testresult-display\">Running...<br />&#160;</div>" +
+			"<div id=\"qunit-testresult-controls\"></div>" +
+			"<div class=\"clearfix\"></div>";
+		controls = id( "qunit-testresult-controls" );
+	}
+
+	if ( controls ) {
+		controls.appendChild( abortTestsButton() );
 	}
 }
 
@@ -617,7 +625,7 @@ QUnit.begin( function( details ) {
 QUnit.done( function( details ) {
 	var banner = id( "qunit-banner" ),
 		tests = id( "qunit-tests" ),
-    stopButton = id( "qunit-stop-tests-button" ),
+		abortButton = id( "qunit-abort-tests-button" ),
 		html = [
 			"Tests completed in ",
 			details.runtime,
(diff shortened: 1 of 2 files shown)


raytiley:
@trentmwillis Let me know if this approach flies and I'll figure out how to add a test. /CC @tundal45 @rwjblue

reporter/html.js
```diff
@@ -108,6 +108,15 @@ function id( name ) {
 	return document.getElementById && document.getElementById( name );
 }
 
+function stopTests() {
+  var stopButton = id( "qunit-stop-tests-button" );
+  if ( stopButton ) {
+    stopButton.innerHTML = "Stopping";
```
leobalter: Where does it says it stopped? This is probably the place for this.

reporter/html.js
```diff
@@ -108,6 +108,15 @@ function id( name ) {
...
+  }
+  QUnit.config.queue.length = 0;
+  return false;
```
leobalter: Can you fix the indentation? It should use tabs instead of spaces.

leobalter:
It also needs to remove the stop button when the tests are done.

raytiley:
I'll fix the indentation, not sure why my editoring config didn't work. So is the expectation that we add / remove the button during a test run, or should it always be int he bar but enabled / disabled? Is the toolbar the correct place for the button?

reporter/html.js
raytiley: It says `Stop` when the button is created and added to the toolbar, and then set again to `Stop` when the tests are done.

leobalter:
> So is the expectation that we add / remove the button during a test run, or should it always be int he bar but enabled / disabled? Is the toolbar the correct place for the button?

Good questions.

I like add / remove. Even a disabled button might be a distraction for me when the tests are done, but I won't complain if enabled / disabled is preferred here.

I think @trentmwillis might give a better answer for where we should place it, I would say yes to have in the toolbar if it better blends with ember-qunit. Otherwise, the toolbar seems ok for it.

platinumazure:
What happens in terms of remaining tests when the stop button is clicked? Do the remaining tests disappear, or are they "skipped" or something else?

Personally I'd want to make sure that the tests are still visible, so a user can know that the test was known by QUnit but wasn't executed.

Also, what happens in terms of logging callbacks? Can we ensure that the correct callbacks are still run (including end of module and end of test run)?

trentmwillis:
For location/appearance, I think it would make the most sense to put this in the `#qunit-testresult` area. My reasoning for this is that we display the current status there (e.g., "Running: some test") so it would make sense to have the button show up there when tests are running and then remove it when the tests have stopped. We could potentially even turn it into a toggle, where "stop" becomes "start". Thoughts?

raytiley:
@trentmwillis so looking into this putting it in `#qunit-testresult` is kinda problematic as that element gets `innerHTML` set on it for each test. Do you think going setting up / tearing down the event handler for each event is worth it... I'm open to making anything work, I just don't want to make a drastic change and get it rejected as too much.

Thanks for the guidance.

rwjblue:
>  putting it in #qunit-testresult is kinda problematic as that element gets innerHTML set on it for each test

Could we change this around a tad and make `#qunit-testresult` be something like:

```html
<div id="qunit-testresult">
  <div id="qunit-testresult-display"></div>
  <div id="qunit-testresult-controls"></div>
</div>
```

The code that currently sets `innerHTML` on `#qunit-testresult` would set it on `#qunit-testresult-display`, and the button would be stable in `#qunit-testresult-controls`.  

RE: design of this (which we haven't really talked about yet), the `#qunit-testresult-display` could be setup to be ~ 90ish% width and the controls would be the rest (with a `float: right`).

@trentmwillis / @leobalter - How does that sound to y'all?

trentmwillis:
@raytiley what @rwjblue suggests is more-or-less what I had in mind.

leobalter:
>Could we change this around a tad and make #qunit-testresult be something like: ...

this is good, @rwjblue. I like it

@raytiley how does it sound to you? Can we pursue this structure?

raytiley:
Hey Everyone... I will for sure take care of this... Just short on time this week due to a unexpected funeral :/  I should be able to handle it this weekend.

leobalter:
I'm sorry, @raytiley. Take it easy.

leobalter:
Hi @raytiley. I imagine a weekend is never really enough, but just for a heads up if you prefer we can help addressing the review points.

trentmwillis:
@raytiley any chance you'll get to this soon? If not, mind if I pick it up?

raytiley:
@trentmwillis et al. Super sorry for the delay on this. I did some more work on it tonight to bring it more inline with what people were asking for. I also did some additional work so it's super clear that the test run was aborted (see screenshot).

I took a quick look at existing tests, but wasn't exactly sure how to go about testing this.

qunit example 2016-11-27 22-29-44

reporter/html.js
```diff
@@ -610,14 +637,38 @@ QUnit.done( function( details ) {
...
+		}
+	}
 
-	if ( banner ) {
+	if ( banner && ( !stopButton || stopButton.disabled === false ) ) {
 		banner.className = details.failed ? "qunit-fail" : "qunit-pass";
 	}
 
+	if ( stopButton ) {
+    stopButton.parentNode.removeChild( stopButton );
```
rwjblue: This needs to be two tabs (looks like spaces ATM)

reporter/html.js
```diff
@@ -610,14 +637,38 @@ QUnit.done( function( details ) {
...
+  }
```
rwjblue: Should be one tab here (looks like 2 spaces ATM)

reporter/html.js
```diff
@@ -599,6 +625,7 @@ QUnit.begin( function( details ) {
 QUnit.done( function( details ) {
 	var banner = id( "qunit-banner" ),
 		tests = id( "qunit-tests" ),
+    stopButton = id( "qunit-stop-tests-button" ),
```
rwjblue: spaces -> tab

trentmwillis:
I'd like to make sure the messaging is consistent, if we're aborting tests, then the button should say "abort". To me, if it says "stop", then that means potential for there to be a "start" coming after it; it's a "recoverable" action if you will.

To that end, I would actually rather we just stop the tests and not update the remaining tests. My thinking is two-fold:

1. This makes it easier for us to introduce a "re-start" feature in the future without altering behavior.
2. If you abort early on in a large test suite (several thousand tests for example), then all the updates can be annoying.

leobalter:
@raytiley do you mind replacing Stop with Abort as suggested by @trentmwillis?

I'm +1 to land this after this small part is fixed.

raytiley:
@trentmwillis @leobalter.... I'll plan on wrapping this up tonight. Can we clarify.

  ## Option A
 - Stop button changes to abort. Everything else remains the same

 ## Option B
 - Stop button stays stopped
 - Unrun tests are NOT shown
 - The status area updates to stay that tests were stopped (not aborted)

trentmwillis:
@raytiley sorry for the delay, let's just do Option A. If we want to add `Start/Stop` functionality in the future, we can do it independent of the `Abort` functionality.

raytiley:
@trentmwillis no worries. Option A done.

trentmwillis:
This seems good to me. The styling of the button could use a bit of work, but we can address that in a follow up.

@leobalter let me know if we're good to merge.

PR #1041, opened 2016-09-17
Test: Properly handle Promises rejected without an error

### LINKED ISSUE #1039: QUnit crashes if promise is rejected without arguments
In case of a failing test when a promise gets rejected, QUnit crashes internally with a  null reference.

This can be reproduced using this snippet:

```js
	QUnit.test("QUnit shall handle failing tests properly", function(assert) {
		var resolve = function(){};
		var reject = function(){};
		var promise = new Promise(function(_resolve, _reject) {
			reject = _reject;
		});
		reject(undefined);
		return promise;
	});
```

The crash happens here:
```js
	resolvePromise: function( promise, phase ) {
		var then, resume, message,
			test = this;
		if ( promise != null ) {
			then = promise.then;
			if ( QUnit.objectType( then ) === "function" ) {
				resume = internalStop( test );
				then.call(
					promise,
					function() { resume(); },
					function( error ) {
						message = "Promise rejected " +
							( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
							" " + test.testName + ": " + ( error.message || error );
						test.pushFailure( message, extractStacktrace( error, 0 ) );
						// Else next test will carry the responsibility
						saveGlobal();
						// Unblock
						resume();
					}
				);
			}
		}
	},
```

The error object is undefined, hence error.message will fail.

Best, Thomas

trentmwillis:
Fixes #1039. Adds tests for the correct behavior as well.

cc @platinumazure @tbiesemann

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -378,7 +378,8 @@ Test.prototype = {
 					function( error ) {
 						message = "Promise rejected " +
 							( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
-							" " + test.testName + ": " + ( error.message || error );
+							" \"" + test.testName + "\": " +
+							( ( error && error.message ) || error );
 						test.pushFailure( message, extractStacktrace( error, 0 ) );
 
 						// Else next test will carry the responsibility
--- a/test/main/promise.js
+++ b/test/main/promise.js
@@ -1,13 +1,18 @@
 // NOTE: Adds 1 assertion
-function createMockPromise( assert ) {
+function createMockPromise( assert, reject, value ) {
+	if ( arguments.length < 3 ) {
+		value = {};
+	}
 
 	// Return a mock self-fulfilling Promise ("thenable")
 	var thenable = {
-		then: function( fulfilledCallback /*, rejectedCallback */ ) {
+		then: function( fulfilledCallback, rejectedCallback ) {
 			assert.strictEqual( this, thenable, "`then` was invoked with the Promise as the " +
 				"context" );
 			setTimeout( function() {
-				return fulfilledCallback.call( thenable, {} );
+				return reject ?
+					rejectedCallback.call( thenable, value ) :
+					fulfilledCallback.call( thenable, value );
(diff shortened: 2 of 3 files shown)


src/test.js
```diff
@@ -378,7 +378,7 @@ Test.prototype = {
 					function( error ) {
 						message = "Promise rejected " +
 							( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
-							" " + test.testName + ": " + ( error.message || error );
+							" \"" + test.testName + "\": " + ( error && error.message );
```
platinumazure: In case of a rejection with a string, should this fall back to `error` like the old logic had done? Something like `( ( error && error.message ) || error )` maybe?

test/main/promise.js
```diff
@@ -152,3 +157,31 @@ QUnit.test( "fulfilled Promise with non-Promise async assertion", function( asse
 	// Adds 1 assertion
 	return createMockPromise( assert );
 } );
+
+QUnit.module( "Proise-aware return values with rejections", {
```
platinumazure: Typo: s/Proise/Promise

test/main/promise.js
```diff
@@ -152,3 +157,31 @@ QUnit.test( "fulfilled Promise with non-Promise async assertion", function( asse
...
+	afterEach: function( assert ) {
+		assert.test.pushFailure = this.pushFailure;
+	}
+} );
+
+QUnit.test( "rejected Promise with undefined value", function( assert ) {
+	this.pushFailure = assert.test.pushFailure;
+	assert.test.pushFailure = function( message ) {
+		assert.equal(
```
platinumazure: Should this be strictEqual? (I tend to favor strictEqual for string compares, so ignore this if repo conventions say use equal for string comparisons)

src/test.js
trentmwillis: Yeah, it should. Late night coding mistake when thinking about how short-circuit evaluation works.

Issue #964, opened 2016-03-18
Remove sessionStorage interactions from HTML Reporter


gibson042:
The HTML Reporter is largely an independent subcomponent, but as long as it interacts with `sessionStorage`, it must remain synchronized with `Test#queue` (which needs the ability to [prioritize previously failed tests](https://github.com/jquery/qunit/blob/e47a7c3db1572387e62846e2e4f39acf7db528f2/src/test.js#L268-L270)). However, it does very little with `sessionStorage`:
* test start: [display alternate "running" text for previously failed tests](https://github.com/jquery/qunit/blob/85bc66ec879bce05626689fcc46bab3733c0b463/reporter/html.js#L654)
* test end: [store result](https://github.com/jquery/qunit/blob/85bc66ec879bce05626689fcc46bab3733c0b463/reporter/html.js#L766-L770)
* suite end: [clear (if no assertions failed)](https://github.com/jquery/qunit/blob/85bc66ec879bce05626689fcc46bab3733c0b463/reporter/html.js#L610-L615)

The first is unnecessary (but if important, could be handled by adding a property to the testStart callback details), and the others really belong near the prioritization logic in test.js. An argument could be made for the smaller change of moving prioritization into the HTML Reporter, but `QUnit.config.reorder` isn't reporter-specific. And further, even though `sessionStorage` is essentially a browser capability, _we_ only care about support of the relevant [`Storage` interface](https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface)—there's no reason why we couldn't expose a settable `QUnit.config.storage` property and thereby enable reordering in non-browser environments once the reporter yields this responsibility.

Therefore, we should do just that: move `sessionStorage` calls into runner code, where they will be invoked against a `QUnit.config` property defaulting to the well-known value.

PR #1042, opened 2016-09-17
HTML Reporter: Decouple from sessionStorage reordering logic

trentmwillis:
Addresses #964.

I opted to add a `previousFailure` property to the `testStart` callback details. I think displaying the alternative text when running a previously failed text is beneficial for users that are new to QUnit and may not be aware of the reordering logic by default. @jzaefferer does this conflict at all with the js-reporters effort?

cc @gibson042

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,4 +1,4 @@
-import { window, setTimeout, console } from "./globals";
+import { window, setTimeout, console, sessionStorage } from "./globals";
 
 import equiv from "./equiv";
 import dump from "./dump";
@@ -257,7 +257,7 @@ export function process( last ) {
 }
 
 function done() {
-	var runtime, passed;
+	var runtime, passed, i, key;
 
 	internalState.autorun = true;
 
@@ -270,6 +270,16 @@ function done() {
 		total: config.stats.all,
 		runtime: runtime
 	} );
+
+	// Clear own sessionStorage items if all tests passed
+	if ( config.reorder && defined.sessionStorage && config.stats.bad === 0 ) {
+		for ( i = 0; i < sessionStorage.length; i++ ) {
+			key = sessionStorage.key( i++ );
+			if ( key.indexOf( "qunit-test-" ) === 0 ) {
+				sessionStorage.removeItem( key );
+			}
+		}
+	}
 }
 
 function setHook( module, hookName ) {
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -1,5 +1,5 @@
 import QUnit from "../src/core";
-import { window, sessionStorage, navigator } from "../src/globals";
+import { window, navigator } from "../src/globals";
 
 // Escape text for attribute or text content.
 export function escapeText( s ) {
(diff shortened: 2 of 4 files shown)


leobalter:
> I opted to add a previousFailure property to the testStart callback details.

I think it's gonna be ok to have it. We should document it.

For the js-reporters, afaik, it won't break anything, but it might not be consumed.

> I think displaying the alternative text when running a previously failed text is beneficial for users that are new to QUnit and may not be aware of the reordering logic by default.

Agreed. It's good to me as well, I always get lost with these reorderings.

LGTM, needs rebasing.

src/core.js
```diff
@@ -284,6 +284,16 @@ function done() {
 		total: config.stats.all,
 		runtime: runtime
 	} );
+
+	// Clear own sessionStorage items if all tests passed
+	if ( config.reorder && defined.sessionStorage && config.stats.bad === 0 ) {
+		for ( i = 0; i < sessionStorage.length; i++ ) {
+			key = sessionStorage.key( i++ );
```
gibson042: Redundant increment.

src/core.js
```diff
@@ -1,4 +1,4 @@
-import { window, setTimeout, console } from "./globals";
+import { window, setTimeout, console, sessionStorage } from "./globals";
```
gibson042: This is well on its way, but I'd still like to see global `sessionStorage` become e.g. `QUnit.config.storage`. Do you think that should instead be a followup PR?

src/core.js
jzaefferer: Since that bug was present before, I wonder if the whole implementation is flawed. Removing items from an object while iterating over it also seems fishy.

jzaefferer:
Overall looking good to me. I don't see a problem with the additional property.

src/core.js
leobalter: this for loop can finally be replaced by forEach or even better `[].reduce`.

> Removing items from an object while iterating over it also seems fishy.

Yes, it's fishy. But this PR is just moving a pre-existent code, so I believe we can address it later.

@trentmwillis I want to confirm you agree, in this case can you open an issue to keep track of this?

src/core.js
leobalter: > Do you think that should instead be a followup PR?

I always prefer followup PRs, more simple to review, and less responsibility on each single merges.

This other PR should also include docs.

trentmwillis: Yeah, we should tackle that as a separate PR.

src/core.js
trentmwillis: I think we can address this in another PR. If the implementation is busted, then we should also investigate getting a failing test.

trentmwillis: Opened https://github.com/jquery/qunit/issues/1054

trentmwillis:
Rebased this. I think we should follow up `QUnit.config.storage` and the potential bugfix in separate PRs.

gibson042:
> Rebased this. I think we should follow up `QUnit.config.storage` and the potential bugfix in separate PRs.

Works for me. I think this is good to go.

Issue #977, opened 2016-03-26
Allow multiple before/beforeEach/after/afterEach


gibson042:
It's inconsistent that every callback passed to `QUnit.test` is preserved, but `beforeEach` and `afterEach` overwrite preexisting values: https://jsfiddle.net/k9af94x7/

We should let the module hooks queue just like the tests do.

**EDIT**: …including `before` and `after`.

trentmwillis:
Since `test` is a function of the `QUnit` global while `beforeEach` / `afterEach` are functions of the local `hooks` object I, personally, don't find it that inconsistent, but acknowledge how similar signatures could lead one to believe they are.

That said, is there a use case for needing the hooks to queue? Given that there is no equivalent queuing for non-callback modules, I am hesitant to want to introduce it here. Could we potentially rename the hook functions to be `setBeforeEach` and `setAfterEach` to clear up the inconsistency?

gibson042:
> Since `test` is a function of the `QUnit` global while `beforeEach` / `afterEach` are functions of the local `hooks` object I, personally, don't find it that inconsistent, but acknowledge how similar signatures could lead one to believe they are.

Exactly. Further, the global-context `test` is likely [on its way out](https://github.com/jquery/qunit/issues/978), which will further emphasize the similarity.

> That said, is there a use case for needing the hooks to queue? Given that there is no equivalent queuing for non-callback modules, I am hesitant to want to introduce it here. Could we potentially rename the hook functions to be `setBeforeEach` and `setAfterEach` to clear up the inconsistency?

I can't think of a pressing case, but module nesting already allows for use of multiple hooks. And I'm not too worried about _non_-callback modules because a) the limitation is inherent to the `{ [name]: fn }` structure, and b) I'm trying to phase them out. It's true that this issue is motivated entirely by a desire for consistency, but that's not necessarily a bad thing (I know, I know... cue [reference](https://en.wikipedia.org/wiki/Self-Reliance)).

leobalter:
I am not a good supporter of nested tests, different than nested modules.

Said that, I like how our api design works with es6 modules bindings and
destructuring bindings:

~~~js
import { module, test } from './qunitjs';

module( 'foo', { beforeEach } => {
  ...
});
~~~

(Typed on my phone, sorry if the code shows up as a mess)

This is our current api, and I like to explore it on this way.

If this convo ends up on ditching out the hooks object argument from
module, we could use the return of module calls to set the hooks in a
chainable way:

~~~js
[...]

PR #1043, opened 2016-09-17
Test improvements

trentmwillis:
This improves some various tests. Most notably it adds tests to verify the order of execution for logging callbacks and module hooks (as raised in https://github.com/jquery/api.qunitjs.com/issues/136).

While writing these tests, 2 questions came up:

1. Should the "done" category of callbacks (`done`, `moduleDone`, and `testDone`) be FIFO or LIFO? I know we had a lot of discussion about a similar topic in #977.
2. Should `moduleDone` fire when starting a nested module? I don't believe it should, since the parent module is not technically finished yet.

### DIFF
--- a/test/callbacks.js
+++ b/test/callbacks.js
@@ -1,0 +1,120 @@
+var invokedHooks = [];
+var done = false;
+
+function callback( name ) {
+	return function() {
+		if ( done ) {
+			return;
+		}
+
+		invokedHooks.push( name );
+	};
+}
+
+QUnit.begin( callback( "begin1" ) );
+QUnit.begin( callback( "begin2" ) );
+QUnit.moduleStart( callback( "moduleStart1" ) );
+QUnit.moduleStart( callback( "moduleStart2" ) );
+QUnit.testStart( callback( "testStart1" ) );
+QUnit.testStart( callback( "testStart2" ) );
+
+QUnit.log( callback( "log1" ) );
+QUnit.log( callback( "log2" ) );
+
+QUnit.testDone( callback( "testDone1" ) );
+QUnit.testDone( callback( "testDone2" ) );
+QUnit.moduleDone( callback( "moduleDone1" ) );
+QUnit.moduleDone( callback( "moduleDone2" ) );
+QUnit.done( callback( "done1" ) );
+QUnit.done( callback( "done2" ) );
+
+QUnit.done( function() {
+	if ( done ) {
+		return;
+	}
+
+	done = true;
+
+	QUnit.test( "verify callback order", function( assert ) {
+		assert.deepEqual( invokedHooks, [
+			"begin1",
+			"begin2",
+			"moduleStart1",
+			"moduleStart2",
+			"testStart1",
+			"testStart2",
+			"module1 > before",
+			"module1 > beforeEach",
+			"module1 > test1",
+			"log1",
+			"log2",
+			"module1 > afterEach",
+			"testDone1",
+			"testDone2",
+			"moduleDone1",
+			"moduleDone2",
+			"moduleStart1",
+			"moduleStart2",
+			"testStart1",
+			"testStart2",
+			"module2 > before",
+			"module1 > beforeEach",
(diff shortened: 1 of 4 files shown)


platinumazure:
For the second issue, I would say we should avoid firing moduleDone until it and all nested modules are truly finished. Otherwise that event conceivably fires twice (I know we've had an issue on that).

test/callbacks.js
```diff
@@ -0,0 +1,121 @@
...
+			"module1 > before",
+			"module1 > beforeEach",
+			"module1 > test1",
+			"log1",
+			"log2",
+			"module1 > afterEach",
+			"testDone1",
+			"testDone2",
+			"moduleDone1", // This shouldn't happen, should it?
+			"moduleDone2", // This shouldn't happen, should it?
```
leobalter: why not? these moduleDone are for the module that has started on [line 43](https://github.com/jquery/qunit/pull/1043/files#diff-e6fe7aebcef218e471b7b20028f2df99R43). This seems right

test/callbacks.js
```diff
@@ -0,0 +1,121 @@
...
+			"log2",
+			"module2 > afterEach",
+			"module1 > afterEach",
+			"module2 > after",
+			"testDone1",
+			"testDone2",
+			"moduleDone1",
+			"moduleDone2",
+			"moduleStart1", // This shouldn't happen, should it?
+			"moduleStart2", // This shouldn't happen, should it?
```
leobalter: same thing here, this seems right, related to the modules done at lines 86/87

test/callbacks.js
trentmwillis: The module that started on line 43 doesn't actually end until line 86/87.

leobalter:
> Should moduleDone fire when starting a nested module? I don't believe it should, since the parent module is not technically finished yet.

IMHO, I believe it's ok to fire this event. That respective module is finished, not the parent module. The provided details should identify the module that is done.

Besides the comments above, this looks great. It's nice to have specific tests for the _execution order_ (which is also a good band name).

trentmwillis:
@leobalter I suppose that depends on how you view modules. I consider nested modules to be part of their parent module. In other words, given:

```javascript
module('parent', function() {
  test('1', function() {});
  module('child', function() {
    test('1', function() {});
  });
  test('2', function() {});
});
```

I would expect the order of events to be:

```
parent -> moduleStart
child -> moduleStart
child -> moduleEnd
parent -> moduleEnd
```

Which reflects the structure of the code as well. What we have *currently* is:

```
parent -> moduleStart
parent -> moduleEnd
child -> moduleStart
child -> moduleEnd
parent -> moduleStart
parent -> moduleEnd
```

Which, in my opinion, should translate to the following code:

```javascript
module('parent', function() {
  test('1', function() {});
});
module('child', function() {
  test('1', function() {});
});
module('parent', function() {
  test('2', function() {});
});
```

Essentially, I don't think a module should be considered "done" until all of it's children (whether tests or other modules) are also done.

test/callbacks.js
leobalter: I finally got it. It's interesting to see how the module1's `after` hook isn't triggered, at least.

leobalter:
```
parent -> moduleStart
parent -> moduleEnd
child -> moduleStart
child -> moduleEnd
parent -> moduleStart
parent -> moduleEnd
```

geez.

yes, this is not good, thanks for clarifying this part.

leobalter:
It's easy to see the callbacks are following when a specific module is running, but the moduleDone is indeed expected right after it's respective `after` hook.

platinumazure:
For what it's worth, as a QUnit user, I would much prefer @trentmwillis' interpretation.

leobalter:
> For what it's worth, as a QUnit user, I would much prefer @trentmwillis' interpretation.

and that's where I agree and even consider the current way it works is a bug.

trentmwillis:
Okay, I think we can merge this test and follow up on improving the `moduleDone` behavior in a separate PR/issue.

PR #1045, opened 2016-09-21
All: Fix some outstanding indentation issues in files

platinumazure:
Now that we're using ESLint and now that ESLint's [`indent`](http://eslint.org/docs/rules/indent) rule is getting some love, I decided to use it to find indentation issues in the repository.

There are still some false positives due to the lack of rule options we need for jQuery-style repos, so I'm not turning on the rule in general yet. Hopefully, though, all the changes I've made here are fairly non-controversial.

Let me know if I'm missing anything or using the wrong commit message. Thanks!

### DIFF
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -100,22 +100,23 @@ export function objectType( obj ) {
 		type = match && match[ 1 ];
 
 	switch ( type ) {
-		case "Number":
-			if ( isNaN( obj ) ) {
-				return "nan";
-			}
-			return "number";
-		case "String":
-		case "Boolean":
-		case "Array":
-		case "Set":
-		case "Map":
-		case "Date":
-		case "RegExp":
-		case "Function":
-		case "Symbol":
-			return type.toLowerCase();
+	case "Number":
+		if ( isNaN( obj ) ) {
+			return "nan";
+		}
+		return "number";
+	case "String":
+	case "Boolean":
+	case "Array":
+	case "Set":
+	case "Map":
+	case "Date":
+	case "RegExp":
+	case "Function":
+	case "Symbol":
+		return type.toLowerCase();
 	}
+
 	if ( typeof obj === "object" ) {
 		return "object";
 	}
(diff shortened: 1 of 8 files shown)


src/core/utilities.js
```diff
@@ -100,22 +100,23 @@ export function objectType( obj ) {
...
-		case "Boolean":
-		case "Array":
-		case "Set":
-		case "Map":
-		case "Date":
-		case "RegExp":
-		case "Function":
-		case "Symbol":
-			return type.toLowerCase();
+	case "Number":
```
platinumazure: Switch statement cases should be aligned with the switch keyword and everything else goes from there. See ["Switch" statements](https://contribute.jquery.org/style-guide/js/#switch-statements) section of the jQuery style guide.

platinumazure:
Oops, I forgot to run the full ESLint configuration (I'd been suppressing all rules except "indent" to make my hunt easier).

Two of my changes resulted in max-len violations. For now, I'm suppressing them with an ESLint config comment-- let me know if I should instead undo the changes in those areas.

test/main/async.js
```diff
@@ -487,8 +487,8 @@ QUnit.test( "multiple `done` calls, final `done` is called BEFORE assertion", fu
 	} );
 } );
 
-QUnit.test( "cannot allow assertions between first `done` call and second `assert.async` call",
-		function( assert ) {
+// eslint-disable-next-line max-len
+QUnit.test( "cannot allow assertions between first `done` call and second `assert.async` call", function( assert ) {
```
trentmwillis: I'd rather avoid the eslint overrides, why not just leave this on two separate lines? Same with the other override instance.

platinumazure:
Ack, ESLint prefers upstream branches for dev team members. Stinks to have
conflicting practices on different repos. I'll try to remember.

platinumazure:
Closing this PR so I can move the branch to my fork (I'll also be altering its history locally anyway).

Issue #1044, opened 2016-09-20
Improve moduleStart and moduleDone behavior in nested modules


trentmwillis:
As discussed in #1043, `moduleDone` should only fire after the entirety of a module has finished, including it's child modules. Additionally, `moduleStart` should only fire once when the first test of the module is about to begin.

leobalter:
fixed by gh-1047

PR #1047, opened 2016-09-23
Core: Improve moduleDone and moduleStart behavior

trentmwillis:
Addressing #1044. This ensures that `moduleDone` and `moduleStart` are fired exactly once per module.

Also, removes `config.previousModule` and `config.moduleStats`, these aren't documented anywhere and seem to have just been there to pass information around QUnit internals, so it seems beneficial to get them off the global. @leobalter hope you can verify if it is safe to remove these.

A couple additional things noticed in this cleanup:

* Modules currently have a `childModule` property, this contains a reference to a single module. This likely means that modules with multiple nested children are broken. We should fix that.

* `moduleDone` reports the `runtime` of the module, but this is done by taking a "start" value when `moduleStart` is emitted. This is likely to produce odd and not very useful results when tests within a module are not run serially (e.g., when reordering after a failure or using the `seed` option). We should likely change the `runtime` to be a summation of the individual test run times.

If the above suggestions seem okay, I can make those changes in a separate PR (to prevent this one being too complex).

### DIFF
--- a/test/callbacks.js
+++ b/test/callbacks.js
@@ -51,8 +51,6 @@ QUnit.done( function() {
 			"module1 > afterEach",
 			"testDone1",
 			"testDone2",
-			"moduleDone1",
-			"moduleDone2",
 			"moduleStart1",
 			"moduleStart2",
 			"testStart1",
--- a/test/callbacks.js
+++ b/test/callbacks.js
@@ -68,8 +68,6 @@ QUnit.done( function() {
 			"testDone2",
 			"moduleDone1",
 			"moduleDone2",
-			"moduleStart1",
-			"moduleStart2",
 			"testStart1",
 			"testStart2",
 			"module1 > beforeEach",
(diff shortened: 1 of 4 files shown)


test/callbacks.js
```diff
@@ -70,8 +68,6 @@ QUnit.done( function() {
 			"testDone2",
 			"moduleDone1",
 			"moduleDone2",
-			"moduleStart1",
-			"moduleStart2",
```
leobalter: this finally feel sane.

leobalter:
That's great!

> Modules currently have a childModule property, this contains a reference to a single module.
> This likely means that modules with multiple nested children are broken. We should fix that.

+1

> We should likely change the runtime to be a summation of the individual test run times.

I think - not sure - we've had this discussion before. It's fair to tell how long a module takes time considering it's own tests, but that also requires inside code running inside a module function.

> I can make those changes in a separate PR (to prevent this one being too complex).

Please! Simple PRs are way easier to review.

PR #1049, opened 2016-09-23
Build: Update dependencies

leobalter:
`grunt-qunit-istanbul` still requires grunt 0.4.x as a peerDep, but it still works

The motivation is to bump grunt to 1.0.1, and grunt-qunit-istanbul still needs some love from @asciidisco or we'll need to find another way to cover our tests.


leobalter:
grunt-contrib-qunit works, without coverage. I don't know where is the best path from here

trentmwillis:
Any chance we can convince @asciidisco to let us take over ownership of the project? Doesn't look like it's being maintained any longer.

@leobalter is there any super urgent need to upgrade?

leobalter:
Not exactly urgent, but I can't run the tests on phantomjs anymore since I installed Mac OS Sierra. It's annoying. 

I have a second machine with Windows and a vps on aws to use while I can't get rid of this issue.

asciidisco:
Sorry for the late reply. I would be happy to give away the ownership of `grunt-qunit-istanbul` as I have no time maintaining it since ... ... ...a long time now.

leobalter:
@asciidisco can I have it, please? I'm willing to update it rather than creating some other library.

I'm @leobalter on github + npm. Thanks!

leobalter:
This is up to date now. PhantomJS is acting weird. Sometimes it does not get QUnit loaded properly. I need to check. In the meanwhile I removed grunt-concurrent as it's not helpful to debug.

trentmwillis:
@leobalter, I'll look into it, this looks like the error I thought I fixed yesterday

trentmwillis:
Been digging into this for a couple hours. Basically, I think there is a race condition that is now biting us: https://github.com/asciidisco/grunt-lib-phantomjs-istanbul/blob/bfdb8d078c0b33a9116ae83c63e3c20f391549bc/phantomjs/main.js#L123-L124 Unsure what to do/how to fix.

leobalter:
I'll try to fix this asap

leobalter:
The problem is there, but I couldn't fix the race condition so far.

I'm +1 to remove coverage to at least maintain the tests running. We can make further plans to reintroduce coverage instrumenting files pre-phantomjs, that should be good even with the Node tests.

leobalter:
I'm merging this for now while it allows me to have a minimum dev environment.

PR #1050, opened 2016-09-24
Core: Support multiple nested modules hooks properly

trentmwillis:
As discussed in #1047:

> Modules currently have a childModule property, this contains a reference to a single module. This likely means that modules with multiple nested children are broken. We should fix that.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -733,10 +733,16 @@ function internalStart( test ) {
 }
 
 function numberOfTests( module ) {
-	var count = module.tests.length;
-	while ( ( module = module.childModule ) ) {
-		count += module.tests.length;
+	let count = module.tests.length,
+		modules = [ ...module.childModules ];
+
+	// Do a breadth-first traversal of the child modules
+	while ( modules.length ) {
+		let nextModule =  modules.shift();
+		count += nextModule.tests.length;
+		modules.push( ...nextModule.childModules );
 	}
+
 	return count;
 }
 
--- a/test/callbacks.js
+++ b/test/callbacks.js
@@ -68,6 +68,23 @@ QUnit.done( function() {
 			"testDone2",
 			"moduleDone1",
 			"moduleDone2",
+			"moduleStart1",
+			"moduleStart2",
+			"testStart1",
+			"testStart2",
+			"module3 > before",
+			"module1 > beforeEach",
+			"module3 > beforeEach",
+			"module3 > test1",
+			"log1",
+			"log2",
+			"module3 > afterEach",
+			"module1 > afterEach",
+			"module3 > after",
+			"testDone1",
+			"testDone2",
+			"moduleDone1",
+			"moduleDone2",
 			"testStart1",
 			"testStart2",
 			"module1 > beforeEach",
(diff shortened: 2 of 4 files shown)


src/test.js
```diff
@@ -733,10 +733,16 @@ function internalStart( test ) {
 }
 
 function numberOfTests( module ) {
-	var count = module.tests.length;
-	while ( ( module = module.childModule ) ) {
-		count += module.tests.length;
+	let count = module.tests.length,
+		modules = [ ...module.childModules ];
```
platinumazure: I think `modules` can be const here.

test/callbacks.js
```diff
@@ -68,6 +68,23 @@ QUnit.done( function() {
 			"testDone2",
 			"moduleDone1",
 			"moduleDone2",
+			"moduleStart1",
```
platinumazure: I was kind of thrown by `moduleStart1` not actually relating to "module 1", but rather is the first callback invoked for module start. Any chance some of the callback text could be changed (maybe in a separate PR) to make it easier to follow when reviewing a file diff that doesn't show some of the context?

test/callbacks.js
```diff
@@ -109,6 +126,18 @@ QUnit.module( "module1", {
...
 
+	QUnit.module( "module3", {
+		before: callback( "module3 > before" ),
+		beforeEach: callback( "module3 > beforeEach" ),
+		afterEach: callback( "module3 > afterEach" ),
+		after: callback( "module3 > after" )
+	}, function() {
+		QUnit.test( "test1", function( assert ) {
+			invokedHooks.push( "module3 > test1" );
+			assert.ok( true );
```
platinumazure: Do we have an expect count for this test? Should we?

src/test.js
trentmwillis: Thought about this. I'd like for us to establish an ES6 convention for types of variable bindings before introducing `const` (since it is more restrictive than `let`).

test/callbacks.js
trentmwillis: Yeah, will try to figure out better wording in another PR.

test/callbacks.js
trentmwillis: I don't think it is needed. The verification of all the logging should verify the assertion counts as well (since `QUnit.log` is invoked per assertion).

src/test.js
leobalter: `Array.from` would be a better option here as it shows better that you want to clone this array. Even though, it's a built-in api I suspect it's not present in every supported browser. 


---

`var` is not deprecated or even close to this. `let` is made for block scoped bindings. `const` too, but it also goes forward protecting a name to change it's value. 

I don't see a reason for us to be super protective with our bindings and `const` everything that are not assigned more than once. This adds logic (to prevent reassignments) we particularly don't need in most cases.

In any way, the best answer should be a convention. @jzaefferer is there any discussion about it for the jQuery Code Style?

src/test.js
```diff
@@ -733,10 +733,16 @@ function internalStart( test ) {
...
+
+	// Do a breadth-first traversal of the child modules
+	while ( modules.length ) {
+		let nextModule =  modules.shift();
+		count += nextModule.tests.length;
+		modules.push( ...nextModule.childModules );
```
leobalter: once again, I don't think it's supported in our required envs, but `modules.concat( nextModule.childModules );` would be great here.

leobalter:
>Left a few comments, but this looks pretty good to me and I'm not sure any immediate changes are needed.

The discussion is interesting but there is no blocker to merge this. I'm +1 to this too.

src/test.js
leobalter: well, it its very well supported.

PR #1051, opened 2016-09-25
Test: Implement QUnit.failTest().

platinumazure:
This allows for error recovery in async conditions.

I've written the logic and added a basic test. I still need to write tests for a few edge cases (mostly along the lines of "what will happen if pushFailureAndResume() has been called already but other assertions come back after the test runner has moved on?"). I just wanted to put this up to get some initial feedback.

Open to feedback on the function names I've used.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -3,7 +3,7 @@ import { window, setTimeout, console, sessionStorage } from "./globals";
 import equiv from "./equiv";
 import dump from "./dump";
 import Assert from "./assert";
-import Test, { test, skip, only, pushFailure, generateHash } from "./test";
+import Test, { test, skip, only, pushFailure, failTest, generateHash } from "./test";
 import exportQUnit from "./export";
 
 import config from "./core/config";
@@ -173,6 +173,7 @@ extend( QUnit, {
 } );
 
 QUnit.pushFailure = pushFailure;
+QUnit.failTest = failTest;
 QUnit.assert = Assert.prototype;
 QUnit.equiv = equiv;
 QUnit.dump = dump;
--- a/src/test.js
+++ b/src/test.js
@@ -119,11 +119,7 @@ Test.prototype = {
 			runTest( this );
 		} catch ( e ) {
 
-			// Else next test will carry the responsibility
-			saveGlobal();
-
-			// Push assertion failure and restart the tests if they're blocking
-			this.pushFailureAndResume(
+			this.pushFailure(
 				[
 					"Died on test #",
 					this.assertions.length + 1,
@@ -134,6 +130,13 @@ Test.prototype = {
 				].join( "" ),
 				extractStacktrace( e, 0 )
 			);
+
+			// Else next test will carry the responsibility
+			saveGlobal();
+
+			if ( config.blocking ) {
+				internalRecover();
+			}
 		}
 
 		function runTest( test ) {
@@ -716,10 +719,8 @@ export function internalStop( test ) {
 
 // Forcefully release all processing holds.
 function internalRecover( test ) {
-	if ( config.blocking ) {
-		test.semaphore = 0;
-		internalStart( test );
-	}
+	test.semaphore = 0;
+	internalStart( test );
 }
 
 // Release a processing hold, scheduling a resumption attempt if no holds remain.
--- a/src/test.js
+++ b/src/test.js
@@ -118,16 +118,22 @@ Test.prototype = {
 		try {
 			runTest( this );
 		} catch ( e ) {
-			this.pushFailure( "Died on test #" + ( this.assertions.length + 1 ) + " " +
-				this.stack + ": " + ( e.message || e ), extractStacktrace( e, 0 ) );
 
 			// Else next test will carry the responsibility
 			saveGlobal();
 
-			// Restart the tests if they're blocking
-			if ( config.blocking ) {
-				internalRecover( this );
-			}
+			// Push assertion failure and restart the tests if they're blocking
+			this.pushFailureAndResume(
+				[
+					"Died on test #",
+					this.assertions.length + 1,
+					" ",
+					this.stack,
+					": ",
+					e.message || e
+				].join( "" ),
+				extractStacktrace( e, 0 )
+			);
 		}
 
 		function runTest( test ) {
@@ -495,15 +501,34 @@ Test.prototype = {
 };
 
 export function pushFailure() {
+	return internalPushFailure( Array.prototype.slice.call( arguments ), false );
+}
+
+export function failTest() {
+	return internalPushFailure( Array.prototype.slice.call( arguments ), true );
+}
+
+function internalPushFailure( args, abandonThisTest ) {
 	if ( !config.current ) {
-		throw new Error( "pushFailure() assertion outside test context, in " +
-			sourceFromStacktrace( 2 ) );
+		throw new Error(
+			[
+				abandonThisTest ? "failTest()" : "pushFailure()",
+				"assertion outside test context, in",
+				sourceFromStacktrace( 2 )
+			].join( " " )
+		);
 	}
 
 	// Gets current test obj
 	var currentTest = config.current;
 
-	return currentTest.pushFailure.apply( currentTest, arguments );
+	var result = currentTest.pushFailure.apply( currentTest, args );
+
+	if ( abandonThisTest ) {
+		internalRecover( currentTest );
+	}
+
+	return result;
 }
 
 // Based on Java's String.hashCode, a simple but not
@@ -674,8 +699,7 @@ export function internalStop( test ) {
 	if ( config.testTimeout && defined.setTimeout ) {
 		clearTimeout( config.timeout );
 		config.timeout = setTimeout( function() {
-			pushFailure( "Test timed out", sourceFromStacktrace( 2 ) );
-			internalRecover( test );
+			failTest( "Test timed out", sourceFromStacktrace( 2 ) );
 		}, config.testTimeout );
 	}
 
@@ -692,8 +716,10 @@ export function internalStop( test ) {
 
 // Forcefully release all processing holds.
 function internalRecover( test ) {
-	test.semaphore = 0;
-	internalStart( test );
+	if ( config.blocking ) {
+		test.semaphore = 0;
+		internalStart( test );
+	}
 }
 
 // Release a processing hold, scheduling a resumption attempt if no holds remain.
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -83,6 +83,7 @@ grunt.initConfig( {
 			"test/index.html",
 			"test/deprecated.html",
 			"test/autostart.html",
+			"test/errorRecovery.html",
 			"test/startError.html",
 			"test/reorderError1.html",
 			"test/reorderError2.html",
--- a/test/errorRecovery.html
+++ b/test/errorRecovery.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Error Recovery Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+</head>
+<body>
+	<div id="qunit"></div>
+	<script src="../dist/qunit.js"></script>
+	<script src="errorRecovery.js"></script>
+</body>
+</html>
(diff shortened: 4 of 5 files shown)


src/test.js
```diff
@@ -659,8 +692,7 @@ export function internalStop( test ) {
 	if ( config.testTimeout && defined.setTimeout ) {
 		clearTimeout( config.timeout );
 		config.timeout = setTimeout( function() {
-			pushFailure( "Test timed out", sourceFromStacktrace( 2 ) );
-			internalRecover( test );
+			pushFailureAndResume( "Test timed out", sourceFromStacktrace( 2 ) );
```
platinumazure: Hmm. If we already have a `test` variable, I wonder if I should instead do `test.pushFailure(/* args */)` followed by `internalRecover( test )`? Somehow I didn't think of this until after I pushed the branch.

src/test.js
```diff
@@ -677,8 +709,10 @@ export function internalStop( test ) {
 
 // Forcefully release all processing holds.
 function internalRecover( test ) {
-	test.semaphore = 0;
-	internalStart( test );
+	if ( config.blocking ) {
```
platinumazure: My assumption is this should always hold (or rather, if `config.blocking` is falsy, then there's really nothing to do with regard to starting the test). If this is a bad assumption, let me know.

src/test.js
```diff
@@ -480,15 +486,42 @@ Test.prototype = {
...
+
+function internalPushFailure( shouldResume ) {
 	if ( !config.current ) {
-		throw new Error( "pushFailure() assertion outside test context, in " +
-			sourceFromStacktrace( 2 ) );
+		throw new Error(
+			[
+				shouldResume ? "pushFailureAndResume()" : "pushFailure()",
+				"assertion outside test context, in",
+				sourceFromStacktrace( 2 )
```
gibson042: I don't think `2` is the correct value here as of this change.

src/test.js
gibson042: I'm not sure, but it seems like there may be some edge cases affected. At minimum, such a change doesn't seem necessary for this PR and should remain independent.

src/core.js
```diff
@@ -172,6 +172,7 @@ extend( QUnit, {
 } );
 
 QUnit.pushFailure = pushFailure;
+QUnit.pushFailureAndResume = pushFailureAndResume;
```
gibson042: It's a really verbose name. I'd like to see something shorter, but won't stand too firmly on it.

jzaefferer: `QUnit.recover`?

src/test.js
jzaefferer: It sounds reasonable, though I'm lacking context to really answer this.

src/core.js
leobalter: +1 on QUnit.recover

platinumazure: I'll switch to QUnit.recover. Thanks!

src/test.js
platinumazure: Ah, I think you're right. Any way I can test this, incidentally?

src/test.js
platinumazure: I'll revert it (and add the if check back to the one consumer) and see what happens.

platinumazure:
Thanks for the early feedback, folks.

When I get back to this, I'm going to switch to QUnit.recover, and I'll edit the first commit message to do so (and thus force-push the branch). In order to keep a decent history, I'll try to put any other changes in a second commit.

platinumazure:
Actually, sorry, changed my mind a bit.

Are we sure `QUnit.recover()` is a good name? Bear in mind that the API will have the same signature as `QUnit.pushFailure()` (i.e., it will take in an error message which will show up as an assertion failure). I think the method name we choose should reflect the fact that an assertion failure is involved. I admit "pushFailureAndResume" is long, but we should try to find a name that conveys the assertion failure without being quite so long.

platinumazure:
@gibson042 @jzaefferer @leobalter In case you missed [this comment](https://github.com/jquery/qunit/pull/1051#issuecomment-249912206), I'm not so sure about `QUnit.recover()` anymore. Mostly it's because I want to make it clear that this is an abnormal failure that needs to be logged.

I've come up with an alternative name that is less verbose than pushFailureAndResume, which is `QUnit.abandonTest()`. What do you think of that? It's four characters more than `QUnit.recover()` but nine characters fewer than `QUnit.pushFailureAndResume()`, and it more clearly (in my opinion) indicates that the test should be considered a failure as a result of performing this operation. Method signature is unchanged (continues to be the same as `QUnit.pushFailure()`.

Thoughts on this new suggestion?

gibson042:
I like `abandonTest`, or even `failTest` if you want to emphasize the connection with `pushFailure`. Good idea, @platinumazure!

trentmwillis:
:+1: to `failTest`. That says to me that QUnit will fail this single test but continue running afterward.

platinumazure:
I can get behind `failTest`. I'll give it another day or two in case others want to weigh in.

leobalter:
+1 to `failTest` as well. At this point, even `fail` would be nice.

platinumazure:
My only issue with `fail` is it could be ambiguous whether that means failing the test or the test run. Sort of like how we renamed `setup` to `beforeEach` due to similar ambiguity.

leobalter:
makes total sense. So I keep my support on `failTest`. o/

platinumazure:
Awesome! Today and tomorrow are quite busy for me, but I hope to get back to this on Thursday.

src/test.js
platinumazure: @gibson042 Could I get some insight on how I can find the correct number? Is the number supposed to be how many levels in the call stack get popped off (from src/core/stacktrace.js etc.) before returning the stack trace as a string?

platinumazure:
@jquery/qunit Updated this PR (now exposing `QUnit.failTest()`). Pretty sure I've addressed everything except the `sourceFromStackTrace()` number. Please let me know if there's anything else I should change.

JamesMGreene:
Why not `assert.fail();`? Seems like a much more natural/consistent API feel to me.

platinumazure:
Because there's also assert.pushFailure. To me, fail is too confusing
between:

1. Push one assert failed;
1. Fail the test, but keep going; and
1. Fail the whole test run (error is completely unrecoverable).

I'm implementing the second, and the first exists as assert.pushFailure.

src/test.js
platinumazure: I've decided not to make this change, for now. Thanks!

src/test.js
platinumazure: This should be reverted now.

src/core.js
platinumazure: I've noted this in the main conversation, but I'm using QUnit.failTest() at this point.

JamesMGreene:
Just so I'm clear, I am interpreting those 3 scenarios a bit more verbosely as the following (with corresponding API method name suggestions):

 1. Push one assert failure but do not return/exit from the test immediately (continue running any additional assertions/code)
     * `assert.fail`
         * a.k.a. existing `assert.test.pushFailure()`
         * a.k.a. existing `QUnit.pushFailure()` \[ `=== assert.test.pushFailure()` \]

 2. Fail the current test, return/exit from it immediately (do not run any additional assertions/code), teardown (run any relevant `after`/`afterEach` hooks), proceed to execute any other remaining tests
     * `assert.test.fail()` / `test.fail()`
     * `this.fail()` (on the `test.testEnvironment`)
     * `throw new Error()`  _\*cough\*_:wink:_\*cough\*_

 3. Fail the current test, return/exit from it immediately (do not run any additional assertions/code), teardown (run any relevant `after`/`afterEach` hooks), do not execute any other remaining tests, consider the whole test run a failure (isn't that technically what any failed assertion achieves, though?)
     * `QUnit.fail()`

JamesMGreene:
**Summary of my last comment, perhaps:**
I don't think that `fail` is a confusing method name so long as it is scoped contextually.

JamesMGreene:
Anyway, I don't mean to derail the progress of this PR but `QUnit.failTest` (or any other top-level methods like this, really) leave a bit of a bad taste in my mouth since we've been progressively moving away from upper-level API exposure like that (e.g. doing all assertions via the provided `assert` argument now rather than global (`ok`), top-level (`QUnit.ok`), or static (`QUnit.assert.ok`) references).

platinumazure:
@JamesMGreene It's worth noting that this effort is to address #990, which talks specifically about handling errors that come up outside of test scope.

I could possibly get behind `config.current.fail()` (I can't remember of `config.current` is a Test instance?) if the team favors that, but this is honestly meant to be used in situations where you might not have the test or assert object in scope.

Given that motivation and context, what sort of API would you prefer to see?

platinumazure:
Also worth noting: We've had the opportunity to define APIs on module/test objects before and have not always chosen to do so.

For example:

```js
QUnit.module("With multiple hooks", function(hooks) {
    hooks.beforeEach(function () {});   // Why is this not `this.beforeEach(function () {});`?
});
```

So I guess I'm not immediately persuaded by the notion that everything should be on a module/test/assert instance, at least not without seeing other issues or PRs to fix existing APIs that don't follow that notion.

If this is a new direction we would like to go, that's great, but I'd love to see that captured in some sort of roadmap issue. Imposing it ad hoc on a PR that was designed for a different purpose entirely seems awkward.

Krinkle:
@platinumazure Hi! I'm not very familiar yet with issue #990, but just sharing a few thoughts that may help.

Our "contextual" design is primarily about lexical scope, which is not limited to the "this" object. Global methods are stateless and rely on implicit association instead of natural or explicit association. Let us consider the following example using the (legacy) QUnit v1.x interface:

```js
QUnit.asyncTest('First', function () {
  equal('foo', 'foo');
  setTimeout(function () {
    equal('bar', 'bar');
    QUnit.start();
  });
  setTimeout(function () {
    equal('baz', 'qux');
  });
});
QUnit.asyncTest('Second', function () {
    setTimeout(function () {
      equal('qux', 'quux');
      QUnit.start();
  });
});
```

This would silently perform the `"qux" equals "quux"` assertion as part of the second test, instead of the first. The new contextual `assert` interface from QUnit v1.15 solved this by moving assertion calls to a local object that is internally associated with the given test. QUnit then has a natural link from the assertion to the related test. (And it can detect if an assertion is made outside test execution, or when it was completed and another test already started.) In theory, we could even allow parallel async test execution.

Between tests, QUnit sets `QUnit.current` to undefined (as it should). So for things truly outside test execution, `QUnit.current` wouldn't work. And if not caught elsewhere, QUnit handles them via global onerror already. Perhaps we can improve that method instead.

Within test execution, there are try/catch handlers by default. Are these disabled in your use case?

JamesMGreene:
> And if not caught elsewhere, QUnit handles them via global onerror already. Perhaps we can improve that method instead.

To me, _this_ really does seem like the situation we are talking about. :+1:

platinumazure:
I wasn't aware that QUnit defined a `window.onerror` handler, actually. I'll have to look at that.

For my original problem, I would certainly love to be able to solve it as `require.onError = QUnit.onTestError` (making up a new API for this because it feels weird to point to `window.onerror` which wouldn't exist in non-browser environments). If that prevents the need for this API, that's awesome. So I'll check that out sometime in the next week or so.

platinumazure:
Update: Still investigating the possibility of using QUnit's `window.onerror` handler, but there are a couple of open PRs in that area which I want to let settle before I attempt anything.

trentmwillis:
@platinumazure any update here?

platinumazure:
@trentmwillis I haven't been able to pick this up since my last update. Summary of my current status is this: I am no longer sure this API is needed/worthwhile, because I believe creating some sort of QUnit.onError handler that can handle Node as well as browser may be good enough for my needs (and possibly better overall).

trentmwillis:
Okay, thanks for the update. If we can find a tractable solution without adding additional API I am :+1:

leobalter:
@platinumazure how do you feel about this PR? Should we use onError for this?

platinumazure:
@leobalter I'm leaning in that direction, but please see [this comment](https://github.com/qunitjs/qunit/issues/990#issuecomment-277463794) and let me know what you think about that overall approach. If people get on board with that approach, then I think this PR is redundant, yes.

platinumazure:
Closing this, since we'll use `QUnit.onError` and a config option to fix the use case this PR was trying to address.

PR #1052, opened 2016-09-25
Build: Consume eslint-plugin-qunit, use "two" configuration

platinumazure:
As some folks know, I maintain the [eslint-plugin-qunit](https://github.com/platinumazure/eslint-plugin-qunit) ESLint plugin. The purpose of that plugin is to provide QUnit-specific rules for ESLint.

I haven't had a good open-source project to dogfood the plugin, so now that we're using ESLint, I'd like to offer it to this project.

This PR enables all of the rules in the ["two" configuration](https://github.com/platinumazure/eslint-plugin-qunit#two) (i.e., rules related to proper use of QUnit 2.0 APIs). If there is interest and we want to use more rules from the project, that could be done in a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "browserstack-runner": "0.4.3",
     "commitplease": "2.3.0",
     "eslint-config-jquery": "1.0.0",
+    "eslint-plugin-qunit": "2.2.0",
     "grunt": "0.4.5",
     "grunt-cli": "0.1.13",
     "grunt-concurrent": "2.2.1",
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -1,7 +1,11 @@
 {
+	"extends": [ "plugin:qunit/two" ],
 	"env": {
 		"browser": true
 	},
+	"plugins": [
+		"qunit"
+	],
 	"globals": {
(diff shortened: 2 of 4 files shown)


jzaefferer:
Looks good to me!

leobalter:
I like this, but I have a suggestion on the plugin that maybe you discussed before: the "two" configuration could be renamed to current, or being just the regular configuration, as anything else should be legacy. I imagine most people implementing new tools as this will also be up to date with QUnit and its current version.

leobalter:
This suggestion should not block the PR. o/

platinumazure:
Thanks for merging!

Just to explain my naming logic: The plugin actually has two shared configurations right now, "recommended" (my personal recommendations for sensible rules to adopt) and "two" (rules to prevent possible runtime errors when using QUnit 2.0 or to help identify migration spots). I imagine if QUnit 3.0 introduces breaking changes, I would create a "three" configuration.

I like the idea of creating a "current" shareable configuration which dynamically uses the most current of the "version" configurations. Only thing is, any time I change what's in a shareable config, I consider that a semver-major change. Not a huge deal since QUnit would also be going semver-major at about the same time, but it's something to consider.

PR #1053, opened 2016-09-26
Test: Replace while loop with recursive [].reduce

leobalter:


src/test.js
```diff
@@ -733,17 +733,11 @@ function internalStart( test ) {
...
 	// Do a breadth-first traversal of the child modules
-	while ( modules.length ) {
-		let nextModule =  modules.shift();
-		count += nextModule.tests.length;
-		modules.push( ...nextModule.childModules );
-	}
-
-	return count;
+	return module.childModules.reduce( function( p, childModule ) {
+	    return p + numberOfTests( childModule );
```
leobalter: if we change the signature of numberOfTests to receive a offset param, we can take some advantage from Tail Call Optimization, consuming less memory in a long list of modules.

That would be like: 

```js
function numberOfTests( module, offset = 0 ) {

	return module.childModules.reduce( function( p, childModule ) {
		return numberOfTests( childModule, p );
	...
```

too bad the args order are the inverse, otherwise it would be even easier.

At least this is also better using arrow fns:

```js
function numberOfTests( module, offset = 0 ) {
	return module.childModules.reduce( ( p, childModule ) => numberOfTests( childModule, p ) );
```

leobalter: fyi, I don't want to pull the trigger using arrow fns for now. But it looks [Babel does not mess with it](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=function%20numberOfTests(%20module%2C%20offset%20%3D%200%20)%20%7B%0A%20%20%20%20return%20module.childModules.reduce(%20(%20p%2C%20childModule%20)%20%3D%3E%20numberOfTests(%20childModule%2C%20p%20)%20)%3B%0A%7D).

I can change it if you prefer.

platinumazure: @leobalter If there are certain expressions we don't want to use that, let me know what those might be and I can try to configure ESLint to restrict them. (Should be pretty easy to restrict arrow functions since they are a separate node type in ESTree.)

leobalter: I've made a fixup commit to address this, my example was not accurate but it was close. :)

src/test.js
```diff
@@ -732,18 +732,13 @@ function internalStart( test ) {
 	}
 }
 
-function numberOfTests( module ) {
-	let count = module.tests.length,
-		modules = [ ...module.childModules ];
+function numberOfTests( module, offset = 0 ) {
```
trentmwillis: I have no problem with changing the order of these arguments to make everything cleaner:

```js
function numberOfTests( offset, module ) {
  return module.childModules.reduce( numberOfTests, module.tests.length + offset);
}
```

If we don't change the order, let's using something more descriptive than `p` for the callback arg.

platinumazure:
Can someone remind me what the motivation for this change is? From what I could tell, the while loop worked perfectly fine and seemed easy enough to read.

leobalter:
@platinumazure I just wanted to get rid of this while loop. At the end we may can get some memory optimization from the recursive TCO but that's uncertain as we don't have any benchmark or performance stats.

platinumazure:
What's wrong with the loop? Is it buggy? Confusing? Hard to read? Prone to error?

leobalter:
> What's wrong with the loop?

Not much, just distracting.

> Is it buggy?

No

> Confusing?

> Hard to read?

> Prone to error?

src/test.js
```diff
@@ -732,18 +732,13 @@ function internalStart( test ) {
...
 
 	// Do a breadth-first traversal of the child modules
```
trentmwillis: This changes the algorithm to "depth-first" instead of "breadth-first".

leobalter:
@platinumazure this is just a change proposal, I started this just trying to improve the use on spread operators:

```js
modules = [ ...module.childModules ]
// to a better array cloning.
```

and

```js
modules.push( ...nextModule.childModules );
// to modules.concat( nextModule.childModules );
```

even though I've finished going way beyond the first intended change.

gibson042:
I find the original version easier to read and comprehend, but don't feel strongly about this.

src/test.js
```diff
@@ -732,18 +732,13 @@ function internalStart( test ) {
...
-	while ( modules.length ) {
-		let nextModule =  modules.shift();
-		count += nextModule.tests.length;
-		modules.push( ...nextModule.childModules );
-	}
-
-	return count;
+	return module.childModules.reduce(
+		( p, childModule ) => numberOfTests( childModule, p ),
+		module.tests.length + offset
```
jzaefferer: This is pretty confusing. I'd put the arguments on the previous line, as usual with other callback methods, and keep the body on its own line, adding curely braces as necessary.

leobalter:
I've went to far with this PR and it's trying to solve way more than my [main intention](https://github.com/jquery/qunit/pull/1053#issuecomment-250032240) and it's creating unnecessary confusion.

Closing this for simplicity. What we have works just fine, so I'll leave it to be improved when we actually have something to change in that part.

PR #1055, opened 2016-10-01
Core: Introduce config.storage option

### LINKED ISSUE #1054: Investigate clearing of sessionStorage
It appears that `sessionStorage` clearing may be busted per https://github.com/jquery/qunit/pull/1042#discussion_r80399634. We should investigate and try to add a regression test if possible.

trentmwillis:
Continuation of #1042 and #964 and fixes #1054.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,4 +1,4 @@
-import { window, setTimeout, console, sessionStorage } from "./globals";
+import { window, setTimeout, console } from "./globals";
 
 import equiv from "./equiv";
 import dump from "./dump";
@@ -269,7 +269,8 @@ export function process( last ) {
 }
 
 function done() {
-	var runtime, passed, i, key;
+	var runtime, passed, i, key,
+		storage = config.storage;
 
 	internalState.autorun = true;
 
@@ -283,12 +284,12 @@ function done() {
 		runtime: runtime
 	} );
 
-	// Clear own sessionStorage items if all tests passed
-	if ( config.reorder && defined.sessionStorage && config.stats.bad === 0 ) {
-		for ( i = 0; i < sessionStorage.length; i++ ) {
-			key = sessionStorage.key( i++ );
+	// Clear own storage items if all tests passed
+	if ( config.reorder && storage && config.stats.bad === 0 ) {
+		for ( i = 0; i < storage.length; i++ ) {
+			key = storage.key( i++ );
 			if ( key.indexOf( "qunit-test-" ) === 0 ) {
-				sessionStorage.removeItem( key );
+				storage.removeItem( key );
 			}
 		}
 	}
--- a/src/core.js
+++ b/src/core.js
@@ -286,8 +286,8 @@ function done() {
 
 	// Clear own storage items if all tests passed
 	if ( storage && config.stats.bad === 0 ) {
-		for ( i = 0; i < storage.length; i++ ) {
-			key = storage.key( i++ );
+		for ( i = storage.length - 1; i >= 0; i-- ) {
+			key = storage.key( i );
 			if ( key.indexOf( "qunit-test-" ) === 0 ) {
 				storage.removeItem( key );
 			}
--- a/src/core.js
+++ b/src/core.js
@@ -285,7 +285,7 @@ function done() {
 	} );
 
 	// Clear own storage items if all tests passed
-	if ( config.reorder && storage && config.stats.bad === 0 ) {
+	if ( storage && config.stats.bad === 0 ) {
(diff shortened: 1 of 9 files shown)


src/core.js
```diff
@@ -282,12 +283,12 @@ function done() {
...
 	} );
 
-	// Clear own sessionStorage items if all tests passed
-	if ( config.reorder && defined.sessionStorage && config.stats.bad === 0 ) {
-		for ( i = 0; i < sessionStorage.length; i++ ) {
-			key = sessionStorage.key( i++ );
+	// Clear own storage items if all tests passed
+	if ( config.reorder && storage && config.stats.bad === 0 ) {
+		for ( i = 0; i < storage.length; i++ ) {
+			key = storage.key( i++ );
```
gibson042: Are we going to let this slide again, or finally fix it? The right logic would look like
```js
for ( i = storage.length - 1; i >= 0; i-- ) {
	key = storage.key( i );
	if ( … ) {
		storage.removeItem( key );
	}
}
```

test/storage.js
```diff
@@ -0,0 +1,64 @@
...
+		return;
+	}
+
+	done = true;
+
+	mockStorage.setItem( "qunit-test-should-remove", true );
+	mockStorage.setItem( "should-not-remove", true );
+	mockStorage.setItem( "qunit-test-should-also-remove", true );
+
+	// TODO: Is there a way to test this that will show up in CI?
```
trentmwillis: In order to test the clearing logic, we need to run assertions _after_ `done` has executed. Unfortunately, this involves using a `setTimeout` which doesn't seem to be reported properly when the tests run in Phantom (seems to work properly in Node land though). So, any suggestions to improve this would be beneficial.

trentmwillis:
@gibson042 added tests and fixed that bug. This is ready for review, want some input on the tests I wrote though. Some of the behavior is unfortunately difficult to test.

test/storage.js
gibson042: Does it work if you make the late addition an async test and move the `setTimeout` into its body?

trentmwillis: Tried that, and if it fails, it shows up but the process still exits successfully. I think the exit code is determined when `done` is fired, which makes sense logically. So we may not be able to catch this easily.

gibson042: I'm surprised it works on Node; this looks like a problem there as well: https://github.com/jquery/qunit/blob/master/build/tasks/test-on-node.js#L80 . Tools just aren't built for runs to end more than once: https://github.com/browserstack/browserstack-runner/blob/f5689b8d03bc4695e2f8a536d94b3f7316bd78ca/lib/_patch/reporter.js#L20 .

Maybe we can instead invoke the file twice, the second time with a query parameter/environment variable/etc. that we look for to alter behavior?

trentmwillis: After further digging, my statement was incorrect, this doesn't work properly in Node.

Invoking the file twice would require the state to be persisted though, correct?

gibson042: It would, and that may be more trouble than it's worth. I guess we should make changes here to handle multiple `done` in Node, and then just skip those assertions in phantomjs.

trentmwillis:
@gibson042 figured out a way to test the storage clearing behavior after a test run. Only happens on the Node tests, but that should be fine IMO.

test/storage.js
```diff
@@ -0,0 +1,21 @@
+/* global module */
```
platinumazure: Wondering if this file should be in a different directory, so it's not confused as a potential test? Something like `test/fixtures` maybe?

trentmwillis: Thought about that myself. Almost thinking I might move this and the tests to a `test/node` directory. This helper could then be in `test/node/helpers` (since it's not really a fixture).

test/node/storage.js
```diff
@@ -0,0 +1,20 @@
...
+  removeItem: function( key ) {
+    if ( key in this._store ) {
+      delete this._store[ key ];
+      this.length--;
+    }
+  },
+  key: function( i ) {
+    return Object.keys( this._store )[ i ];
+  },
+  length: 0
```
gibson042: Manual record-keeping would be unnecessary if this were `get length() { return Object.keys( this._store ).length; }`.

trentmwillis:
@platinumazure thanks for the tip!

@gibson042 understandable, I've been a bit pre-occupied as well with my day job. Hoping to get back into QUnit a bit more in the coming weeks.

test/node/storage.js
trentmwillis: Good suggestion. Updating then merging.

PR #1057, opened 2016-10-02
Core: Rewrite QUnit.equiv to be breadth-first

divec:
Also:
- Redefine Map/Set equivalence to match the .has methods
- Fix recursion bugs, and add tests for failing cases
- Implement recursion for Map
- Document algorithm and rationale

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -134,18 +134,42 @@ export default ( function() {
 			return true;
 		},
 
+		// Define sets a and b to be equivalent if for each element aVal in a, there
+		// is some element bVal in b such that aVal and bVal are equivalent. Element
+		// repetitions are not counted, so these are equivalent:
+		// a = new Set( [ {}, [], [] ] );
+		// b = new Set( [ {}, {}, [] ] );
 		"set": function( a, b ) {
 			var innerEq,
 				outerEq = true;
 
 			if ( a.size !== b.size ) {
+
+				// This optimization has certain quirks because of the lack of
+				// repetition counting. For instance, adding the same
+				// (reference-identical) element to two equivalent sets can
+				// make them non-equivalent.
 				return false;
 			}
 
 			a.forEach( function( aVal ) {
+
+				// Short-circuit if the result is already known. (Using for...of
+				// with a break clause would be cleaner here, but it would cause
+				// a syntax error on older Javascript implementations even if
+				// Set is unused)
+				if ( !outerEq ) {
+					return;
+				}
+
 				innerEq = false;
 
 				b.forEach( function( bVal ) {
+
+					// Likewise, short-circuit if the result is already known
+					if ( innerEq ) {
+						return;
+					}
 					if ( innerEquiv( bVal, aVal ) ) {
 						innerEq = true;
 					}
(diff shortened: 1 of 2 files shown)


leobalter:
I don't know if I really want deep comparison on Map and Set objects. In any way I noticed Lodash's `_.isEqual` already do it.

```js
var a = new Set([1, 2]); var b = new Set([1, 3]); _.isEqual(a, b)
// false

...
var a = new Set([1, 2]); var b = new Set([1, 2]); _.isEqual(a, b)
// true
```

At this point I believe it's reasonable enough we pursue an implementation using the lodash's method.

It's good to see the code in this branch don't return false positives for WeakSet and WeakMap. See this:

```js
var o = { a: 42 };
var s1 = new Set( [o] );
var s2 = new Set( [o] );
var s3 = new WeakSet( [o] );

QUnit.equiv( s1, s2 ); // true
QUnit.equiv( s1, s3 ); // true
```

I need to finish reviewing this, but these are my initial thoughts so far.

divec:
Thanks for your comments! I think we should be careful about Lodash's _.isEqual , because it has n^2 complexity (i.e. it compares sets using nested loops - as does QUnit's Set handing code right now). I'm sure that's a good choice sometimes, but for a general comparison test in a unit testing framework it could cause frustrating unexplained freezes. The simple example below, comparing two modest sets of numbers, causes over 12 million comparisons and takes several seconds:
```
oneToFiveThousand = Array.from( 'x'.repeat(5000) ).map( function( c, i ){ return i; } );
_.isEqual(
        new Set( oneToFiveThousand ),
        new Set( oneToFiveThousand.slice().reverse() )
);
```
A real-life situation, where isEqual must recurse into non-primitive set elements, would be even slower. I don't think this should happen implicitly whenever a tested data structure contains a Set. At the very least the behaviour should need triggering explicitly e.g. by calling a special method.

For this reason, I think it's better to do strict matching on Set elements and Map keys (with .has instead of a nested loop). I think it's fine for equiv to descend into Map *values* though.

leobalter:
cc @trentmwillis this looks good to me. I'm +1 if you want to merge this

leobalter:
@divec can you update the CLA thing, please?

divec:
@leobalter Ok, I've update the CLA, and I've also changed the function checks to include object method .toString() comparison, mimicking the effect of 904e389d85232a9d6d783881e879fce9a0965b3f (which was conflicting with this commit).

trentmwillis:
Overall, this looks good to me. I would, however, prefer that it be broken up into separate commits (for each bullet point mentioned above).  I am primarily worried about the Map/Set changes potentially breaking existing test suites and so if that happens it would be nice to have a smaller set of changes to look at.

divec:
@trentmwillis Good point. I've broken it up, and I've put the Map key and Set changes in two different commits as I can imagine them breaking different test suites.

leobalter:
@trentmwillis This is good to me, for the possible breaking change, my recommendation is a manual strategy I did before: build this version to run again big repos. jQuery, jQuery UI, jQuery Mobile, Globalize, anything we can get with Ember, Lodash, etc.

If we find any breaking change from the master branch version to branch, we need to flag this for a major release.

trentmwillis:
So after thinking about this more, I'm not comfortable with the Map/Set changes (everything else is good). Allow me to explain why:

First, this is clearly a breaking change now that I look at the split up commits. For instance, look at the assertion changes made [here](https://github.com/qunitjs/qunit/pull/1057/commits/dcc7b297bdee3024a072706721eaee36f86deb1c#diff-d901125d4db3b69d066f186fe4a455b1) and [here](https://github.com/qunitjs/qunit/pull/1057/commits/eaa521123ca307330444dbcc4a9f684a863f0526#diff-d901125d4db3b69d066f186fe4a455b1).

Second, I think from a testing perspective, using the "strictness" of `.has` is not particularly valuable. When running assertions against the output of a unit of code, you are unlikely to have references to the exact objects contained within a set. For example:

```
const generatedSet = getSetOfThings();
assert.deepEqual( generatedSet, new Set( { name: 'thing1' } ));
```

It is much more likely that you will know what the expected contents of the `Set` / `Map` are, but to not have direct references to them.

I think the only real benefit gained by using `.has` is increase in speed, but I prioritize both ease of use and testability over speed for test suites; both of which I feel will be negatively impacted by this change.

divec:
@trentmwillis I understand your concerns about likely test usage (I think there's some tension between that and conceptual clarity here).

I've changed the commit stack so it keeps the existing Map/Set logic, and documents it in more detail. I also added some short-circuiting to the forEach callbacks.

leobalter:
@trentmwillis I'll wait for your thoughts on this.

divec:
Ping - anything I can do to help this along?

PR #1058, opened 2016-10-03
Core: QUnit.equiv object methods comparison

### LINKED ISSUE #914: QUnit.equiv ignores object methods
https://jsfiddle.net/xcrh9n3d/

Since we verify inherited properties anyway, we should replace the weird constructor-based "function" callback with a simple `return false` (relying solely on the strict equality check in `innerEquiv`).

kumarmj:
fixes #914

### DIFF
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -1315,7 +1315,6 @@ QUnit.test( "Instances", function( assert ) {
 	// But we decided to skip function for instances.
 	assert.equal( QUnit.equiv( b1, b2 ), true, "Same property, same constructor" );
 
-	// Added as bugfix for same property, different reference functions.
 	assert.equal( QUnit.equiv( c1, c2 ), false, "Same property with different reference and different function" );
 	assert.equal( QUnit.equiv( c1, c3 ), true, "Same property with different reference but same function" );
 
--- a/test/main/deepEqual.js
+++ b/test/main/deepEqual.js
@@ -1289,7 +1289,7 @@ QUnit.test( "Prototypal inheritance", function( assert ) {
 } );
 
 QUnit.test( "Instances", function( assert ) {
-	var a1, a2, b1, b2, car, carSame, carDiff, human;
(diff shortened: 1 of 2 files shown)


test/main/deepEqual.js
```diff
@@ -1301,12 +1301,24 @@ QUnit.test( "Instances", function( assert ) {
...
+
 	assert.equal( QUnit.equiv( a1, a2 ), true, "Same property, same constructor" );
 
 	// b1.fn and b2.fn are functions but they are different references
 	// But we decided to skip function for instances.
 	assert.equal( QUnit.equiv( b1, b2 ), true, "Same property, same constructor" );
 
+	// Added as bugfix for same property, different reference functions.
+	assert.equal( QUnit.equiv( c1, c2 ), false, "Same property differnent reference, but different functions" );
+	assert.equal( QUnit.equiv( c1, c3 ), true, "Same property different reference, same functions" );
```
trentmwillis: I would prefer the assertion messages be formatted like such:

* "Same property with different reference and different function"
* "Same property with different reference but same function"

I think it makes it a bit clearer.

test/main/deepEqual.js
```diff
@@ -1301,12 +1301,24 @@ QUnit.test( "Instances", function( assert ) {
...
+	c2 = new C( function() { return false; } );
+	c3 = new C( function() { return true; } );
+
 	assert.equal( QUnit.equiv( a1, a2 ), true, "Same property, same constructor" );
 
 	// b1.fn and b2.fn are functions but they are different references
 	// But we decided to skip function for instances.
 	assert.equal( QUnit.equiv( b1, b2 ), true, "Same property, same constructor" );
 
+	// Added as bugfix for same property, different reference functions.
```
trentmwillis: I think this comment is unnecessary. The assertion messages are enough.

gibson042:
LGTM. @trentmwillis?

trentmwillis:
Thanks @kumarmj!

PR #1060, opened 2016-10-10
Core: Support a predefined QUnit.config

### LINKED ISSUE #1059: feature: Make it possible to load QUnit after domReady
I am using QUnit together with [OpenUI5](https://github.com/SAP/openui5)

UI5 comes with its own amd syntax similar to require.js.

It would be nice to be able to load QUnit asynchronously to provide a test environment from the framework side.

How i would imagine it is that you could define a global QUnit config before QUnit itself is loaded

`
window.QUnit = {
  config: {
    autostart: false
  }
};
`

When QUnit is loaded, it takes this config and extends it with the needed defaults. Here is a full [jsbin](http://jsbin.com/fewake/1/edit?html,js,output) outlining the problem

If you like the idea i would create a pull request for it. I just wanted to discuss it upfront.

TobiasOetzel:
If a global QUnit.config object is present before QUnit is loaded,
the default config values will be extended.
The use case is to configure QUnit before it is loaded.

Closes #1059

### DIFF
--- a/src/export.js
+++ b/src/export.js
@@ -47,7 +47,9 @@ Object.defineProperty( QUnit, "reset", {
 } );
 
 if ( defined.document ) {
-	if ( window.QUnit ) {
+
+	// QUnit may be defined when it is preconfigured but then only QUnit and QUnit.config may be defined.
+	if ( window.QUnit && window.QUnit.version ) {
 		throw new Error( "QUnit has already been defined." );
 	}
 
--- a/test/preconfigured.html
+++ b/test/preconfigured.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Preconfigured Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+	<script>
+		window.QUnit = {
+			config: {
+				autostart: false,
+				reorder: false
+			}
+		};
+	</script>
+	<script src="../dist/qunit.js"></script>
+	<script src="preconfigured.js"></script>
+</head>
+<body>
+	<div id="qunit"></div>
+</body>
+</html>
--- a/test/preconfigured.js
+++ b/test/preconfigured.js
@@ -1,0 +1,21 @@
+window.addEventListener( "load", function() {
+
+	// make sure QUnit has started if autostart would be true
+	setTimeout( function() {
+		QUnit.module( "QUnit.preconfigured.asyncTests" );
+		QUnit.test( "QUnit.config should have an expected default", function( assert ) {
+			assert.ok( QUnit.config.scrolltop, "The scrolltop default is true" );
+		} );
+
+		QUnit.test( "Qunit.config.reorder default was overwritten", function( assert ) {
+			assert.notOk( QUnit.config.reorder, "reorder was overwritten" );
+		} );
+
+		QUnit.start();
+	}, 100 );
+} );
+
+QUnit.module( "QUnit.preconfigured" );
+QUnit.test( "Autostart is false", function( assert ) {
+	assert.notOk( QUnit.config.autostart, "The global autostart setting is applied" );
+} );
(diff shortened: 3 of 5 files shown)


src/core.js
```diff
@@ -11,7 +11,11 @@ import { defined, extend, objectType, is, now } from "./core/utilities";
 import { registerLoggingCallbacks, runLoggingCallbacks } from "./core/logging";
 import { sourceFromStacktrace } from "./core/stacktrace";
 
-const QUnit = {};
+var QUnit = {};
```
platinumazure: Why do we need to change this to `var`? I don't see the `QUnit` local being overwritten here, am I missing something?

TobiasOetzel: You are right, it was just my IDE fooling me, i don't need it - will revert it to const thanks!

src/core.js
```diff
@@ -13,6 +13,10 @@ import { sourceFromStacktrace } from "./core/stacktrace";
 
 const QUnit = {};
 
+// take a predefined QUnit.config and extend the defaults
+var globalConfig = window && window.QUnit && window.QUnit.config;
+extend( config, globalConfig );
```
trentmwillis: This should go in `src/core/config.js` instead. Easier to reason about what is happening that way.

src/export.js
```diff
@@ -47,8 +47,11 @@ Object.defineProperty( QUnit, "reset", {
 } );
 
 if ( defined.document ) {
-	if ( window.QUnit ) {
-		throw new Error( "QUnit has already been defined." );
+	if ( window.QUnit && window.QUnit.test ) {
```
trentmwillis: Would prefer that we check `window.QUnit.version`. I don't see `test` going away any time soon, but I think that checking `version` (which is essentially just a tag) makes more sense as it is unlikely to ever be mocked out.

src/export.js
```diff
@@ -47,8 +47,11 @@ Object.defineProperty( QUnit, "reset", {
...
+		throw new Error(
+			"QUnit has already been defined.\n" +
+			"Only a global config is allowed: QUnit = { config: { autostart: false, ...} }"
```
trentmwillis: I'm not sure this updated message is relevant. This error should only occur when `QUnit` has accidentally been loaded twice, shouldn't ever show up when someone is trying to do a custom config.

TobiasOetzel:
Thanks for the review. All points are valid and the change looks better after applying them.

Should i also add a contribution to the documentation pages? Because probably this needs to be documented in the QUnit.config section.

TobiasOetzel:
@trentmwillis are the small changes ok?
Do you need any further changes from my sinde?

test/preconfigured.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Preconfigured Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+	<script>
+		QUnit = {
```
trentmwillis: Let's be explicit here and do `window.QUnit = {`.

test/preconfigured.html
```diff
@@ -0,0 +1,21 @@
...
+			config: {
+				autostart: false,
+				reorder: false
+			}
+		}
```
trentmwillis: Should have `;` at the end.

test/preconfigured.js
```diff
@@ -0,0 +1,21 @@
+window.addEventListener( "load", function() {
+
+	// make sure QUnit has started if autostart would be true
+	setTimeout( function() {
+		QUnit.module( "QUnit.proconfigured.asyncTests" );
```
trentmwillis: Typo: `proconfigured` -> `preconfigured`

TobiasOetzel:
There was a second error thrown when overloading QUnit since there were 2 started instances working on the same reference of the stats object inside of the config.
This lead to a strange error. To get around the problem i only extend if there is not an overloaded QUnit

trentmwillis:
Looks good to me! Thanks. Would you mind addressing https://github.com/qunitjs/api.qunitjs.com/issues/140 as well?

TobiasOetzel:
I will adress it soon. Thanks for your reviews!

Issue #1062, opened 2016-10-17
catch TypeError in sync call


ffoysal:
I have tests in two files, one file

``` js
QUnit.test( "Test 1", function( assert ) {

    var obj = undefined;
    var doneFun = assert.async();
    setTimeout(function(){
        assert.equal('hell', obj.hell, 'Hello');
        doneFun();
    },100) });
```

and the other is 

``` js
QUnit.test( "Test 2", function( assert ) {
    var doneFun = assert.async();
    setTimeout(function(){
        assert.equal('hell', 'hell','hello');
        doneFun();
    },500); });
```

when I run both together I got

```
        assert.equal('hell', obj.hell, 'Hello');
                                ^

TypeError: Cannot read property 'hell' of undefined
    at null._onTimeout (/home/foysal/workspace/qa/qa-test-framework/source/tests/mytest.js:12:30)
    at Timer.listOnTimeout (timers.js:92:15)
```

want to catch any TypeError in async test.
I am running in nodejs.

Help is very much appreciated

leobalter:
> Help is very much appreciated

Likewise, always.

It's interesting this bug happens on the git version, but the current released version still catches the error: http://jsbin.com/jevutajuni/edit?html,js,console,output (using QUnit 2.0.1)

<img width="472" alt="screenshot 2016-10-17 18 46 48" src="https://cloud.githubusercontent.com/assets/301201/19458440/1774a4c2-949a-11e6-9ad9-734667e4477a.png">

The current git version from the master branch is not catching the exception:

<img width="470" alt="screenshot 2016-10-17 18 47 44" src="https://cloud.githubusercontent.com/assets/301201/19458466/389b4eda-949a-11e6-8871-dd08d5d37d43.png">

---

I'll run a git bisect to find this bug and fix it. Thanks for the report.

leobalter:
oh, this is an uncaught bug from 80a14ee5ac49c7f0981884a6cbb5786dea3a9c8b (PR #1025).

Now it's time to figure out how to fix it.

leobalter:
> I am running in nodejs.

Ok, that's another issue we need to address. 2 bugs in 1 issue. 👍

trentmwillis:
Closing this as the behavior shown by @leobalter is fixed in `2.1.0` and on the current `master` branch (either due to #1063 or as a side-effect of something else).

[...]

Issue #1061, opened 2016-10-14
Should we extract src/core/onerror.js core logic to be independent of browser?


platinumazure:
Can we extract the [core logic](https://github.com/jquery/qunit/blob/904e389d85232a9d6d783881e879fce9a0965b3f/src/core/onerror.js#L28-L38) of src/core/onerror into a module, and then change the HTML reporter (or other browser-specific orchestration code) to expose a wrapper of that code as `window.onerror`?

This would allow me to use that same logic pretty easily in similar-but-not-quite-equivalent scenarios, such as `require.onError` when using RequireJS.

As a bonus, this may prevent the need for #1051 (at least for me). Or, at the very least, it will allow us to have a purer API there (right now I'm trying to solve a few different problems with that PR).

platinumazure:
Ping @Krinkle @JamesMGreene, since both of you inspired me to look in this direction.

JamesMGreene:
Should also probably roll in handling `process.on( 'uncaughtException', function( err ) { ... } );` when running on Node.js, and any other CommonJS/Rhino/Nashorn error handling as makes sense.

JamesMGreene:
Our current `window.onerror` handler also does not attempt to utilize the modern signature, which provides 5 arguments (the last being the actual `Error` object).

leobalter:
I'm +1 to seek a solution using `process.on( 'uncaughtException', ...` and a modern event listener to replace window.onerror, addEventListener is a valid option after QUnit 2

platinumazure:
Fixed by baff17d.

PR #1063, opened 2016-10-17
Core: Load the onerror module

leobalter:
Ref gh-1062

This is not fixing #1062, but it fixes another bug I found from there.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -10,6 +10,7 @@ import config from "./core/config";
 import { defined, extend, objectType, is, now } from "./core/utilities";
 import { registerLoggingCallbacks, runLoggingCallbacks } from "./core/logging";
 import { sourceFromStacktrace } from "./core/stacktrace";
+import {} from "./core/onerror";
 
 const QUnit = {};
 


leobalter:
No, @jquerybot, it's not my first contribution to this project. :)

platinumazure:
I'd better track this one while thinking about #1061. To me, it's a bit of a smell that we are importing that file for its side effects- better that that module export an error handling function and the HTML reporter use that to create a `window.onerror` handler.

leobalter:
I believe it's time to release a new version, this one can land faster, but it shouldn't block your work on #1061

platinumazure:
No objection to releasing this in a new version, #1061 still needs
discussion and buy-in. Thanks!

src/core.js
```diff
@@ -10,6 +10,7 @@ import config from "./core/config";
 import { defined, extend, objectType, is, now } from "./core/utilities";
 import { registerLoggingCallbacks, runLoggingCallbacks } from "./core/logging";
 import { sourceFromStacktrace } from "./core/stacktrace";
+import {} from "./core/onerror";
```
trentmwillis: You be able to just do:

```
import "./core/onerror";
```

Similar to what is done in `src/qunit.js`. I prefer this pattern as it is a good indicator that you simply want the module evaluated, but don't expect exports from it.

trentmwillis:
Went ahead and made the adjustment myself and merged.

PR #1069, opened 2016-11-12
Use eslint-plugin-html to lint HTML files

platinumazure:
**This should now be safe to merge.**

This PR uses [eslint-plugin-html](https://github.com/BenoitZugmeyer/eslint-plugin-html) to run ESLint on the JavaScript snippets in our HTML files.

Some notes:

1. I edited .gitattributes to make sure our HTML files were treated as having linefeeds on checkout. (From what I can tell, all the HTML files have LF line endings in the repository, but my local Git was checking them out with CRLF due to being on Windows.) I did this to satisfy the ESLint `linebreak-style` rule.
1. Inspired by [this PR review](https://github.com/qunitjs/qunit/pull/1060#pullrequestreview-8301003), I've favored assigning to `window` instead of using implicit global var declarations.
1. There is currently a bug in eslint-plugin-html which causes bogus lint errors to appear before a closing `</script>` tag if there is another opening `<script>` tag further down. I've submitted [this PR](https://github.com/BenoitZugmeyer/eslint-plugin-html/pull/37) to fix the upstream issue.

When the plugin issue is fixed and a new patch release is cut, I'll push another commit to consume that version, and then hopefully CI will pass and this can be merged. (I pushed the PR before that so it could get early review, and I can abandon the branch if this is rejected.)

### DIFF
--- a/test/startError.html
+++ b/test/startError.html
@@ -10,20 +10,17 @@
 	<script src="../dist/qunit.js"></script>
 	<script>
 
-		/* exported autostartStartError, tooManyStartsError */
-		var autostartStartError, tooManyStartsError;
-
 		try {
 			QUnit.config.autostart = true;
 			QUnit.start();
 		} catch ( thrownError ) {
-			autostartStartError = thrownError;
+			window.autostartStartError = thrownError;
 		}
 
 		try {
 			QUnit.start();
 		} catch ( thrownError ) {
-			tooManyStartsError = thrownError;
+			window.tooManyStartsError = thrownError;
 		}
 
 		// Delete QUnit global so we can reload it
--- a/test/startError.html
+++ b/test/startError.html
@@ -9,20 +9,20 @@
 	<div id="qunit"></div>
 	<script src="../dist/qunit.js"></script>
 	<script>
+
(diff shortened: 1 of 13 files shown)


platinumazure:
This should be safe to merge now. Let me know if I need to rebase and/or squash.

test/startError.html
```diff
@@ -9,20 +9,20 @@
 	<div id="qunit"></div>
 	<script src="../dist/qunit.js"></script>
 	<script>
+
+		/* exported autostartStartError, tooManyStartsError */
 		var autostartStartError, tooManyStartsError;
```
trentmwillis: Why is this usage not changed to use `window` like the others?

platinumazure: Excellent catch, the answer is "no reason, I just missed it".

trentmwillis: Okay, let me know when you've updated and I'll merge.

platinumazure:
@trentmwillis Fixed the issue and rebased as well, please let me know if I need to do anything else.

trentmwillis:
Merged! Thanks :)

PR #1070, opened 2016-11-12
Build: Update Node support versions

### LINKED ISSUE #1067: Run tests on Node 7 (& perhaps stop on 0.10?)
Node 7 is stable so QUnit should be tested on it. OTH, upstream support for Node 0.10 has just ended so it might be good to plan when to drop it.

jQuery Core has already stopped testing on Node 0.10 but we're in a different situation than QUnit.

trentmwillis:
Includes our supported versions in package.json to clarify the current support.

This also brings our support policy in line with Rollup, so we can avoid having to switch versions for building.
    
Fixes #1067

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
   - "0.12"
   - "4"
   - "6"
+  - "7"
 env:
   - NPM_SCRIPT=ci
 matrix:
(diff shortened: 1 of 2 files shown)


platinumazure:
Can you document (or link to existing documentation for) what you mean by "brings our support policy in line with Rollup"?

trentmwillis:
Sure, previously we included `0.10` in our build matrix, but Rollup didn't support `0.10`. Now, we support the same versions, or at least the same versions they [test against](https://github.com/rollup/rollup/blob/5344665d7256c7192cb69068635ff0d3f6c6cd8b/.travis.yml#L4-L6), or newer.

platinumazure:
Ah, okay, so we only had to switch because our prepublish hook failed on
0.10 basically? Makes sense, thanks for explaining.

trentmwillis:
Also added the supported versions to our `package.json` to make it more obvious for users.

@leobalter @mgol would appreciate your input before merging.

leobalter:
It looks great

mgol:
LGTM

PR #1072, opened 2016-11-26
Core: Fix tests for equivalent functions

divec:
Update expected results to match 904e389d85232a9d6d783881e879fce9a0965b3f .


divec:
Hmm, I think I may have a Heisenbug here - investigating

Issue #1077, opened 2016-12-06
CI breaking on Node 0.12


leobalter:
Of course I broke everything again.

Travis CI is breaking on Node 0.12. This passed as a [false positive](https://travis-ci.org/qunitjs/qunit/jobs/181583215#L219-L223) from #1049.

ESLint is failing on Node 0.12...

mgol:
Hint: you need sth like the following:
https://github.com/jquery/jquery/blob/7fbab88/Gruntfile.js#L19-L24
https://github.com/jquery/jquery/blob/7fbab88/Gruntfile.js#L189-L194
https://github.com/jquery/jquery/blob/7fbab88/Gruntfile.js#L232

to make it work (basically, to turn ESLint into a noop in Node <4).

platinumazure:
When making Travis run ESLint, I intentionally made it so it only ran on Node 6 (because we don't need to lint in every configuration since ESLint basically behaves the same on Node >=4.0 and because it's not a crucial functionality test that needs to run on all supported platforms). Please let me know if you have questions on how I did that.

mgol:
@platinumazure QUnit uses Grunt and just configuring that on Travis is not enough. `Gruntfile.js` is loaded in all Node versions and this line:
https://github.com/qunitjs/qunit/blob/d3aaf9b/Gruntfile.js#L4
means the `eslint` task gets loaded which breaks in older Nodes as it's written in ES6.

I don't think it's feasible to rely purely on Travis here, QUnit needs the jQuery Core treatment I mentioned in my previous comment.

platinumazure:
Thanks. Bit surprised that what I had done before worked, then...

On Dec 6, 2016 8:39 AM, "Michał Gołębiowski" <notifications@github.com>
wrote:

@platinumazure <https://github.com/platinumazure> QUnit uses Grunt and just
configuring that on Travis is not enough. Gruntfile.js is loaded in all
Node versions and this line:
https://github.com/qunitjs/qunit/blob/d3aaf9b/Gruntfile.js#L4
means the eslint task gets loaded which breaks in older Nodes as it's
written in ES6.

I don't think it's feasible to rely purely on Travis here, QUnit needs the
jQuery Core treatment I mentioned in my previous comment.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/qunitjs/qunit/issues/1077#issuecomment-265165809>, or mute
the thread

mgol:
[...]

PR #1078, opened 2016-12-07
Build: Don't run ESLint task in old Node

trentmwillis:
Addressing #1077.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "rollup-plugin-babel": "2.6.1"
   },
   "scripts": {
-    "browserstack": "sh build/run-browserstack.sh",
+    "browserstack": "grunt build && sh build/run-browserstack.sh",
     "build": "grunt build",
     "test": "grunt"
   },
--- a/package.json
+++ b/package.json
@@ -53,8 +53,7 @@
   "scripts": {
     "browserstack": "sh build/run-browserstack.sh",
     "build": "grunt build",
-    "test": "grunt test",
-    "prepublish": "grunt build"
+    "test": "grunt"
   },
   "commitplease": {
     "components": [
--- a/package.json
+++ b/package.json
@@ -52,9 +52,8 @@
   },
   "scripts": {
     "browserstack": "sh build/run-browserstack.sh",
-    "ci": "grunt test",
-    "ci-include-lint": "grunt eslint && grunt test",
-    "test": "grunt",
+    "build": "grunt build",
+    "test": "grunt test",
     "prepublish": "grunt build"
   },
   "commitplease": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,12 @@ node_js:
   - "6"
   - "7"
 env:
-  - NPM_SCRIPT=ci
+  - NPM_SCRIPT=test
 matrix:
   fast_finish: true
   allow_failures:
     - env: NPM_SCRIPT=browserstack
-  exclude:
-    - node_js: "6"
-      env: NPM_SCRIPT=ci
   include:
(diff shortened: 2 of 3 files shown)


platinumazure:
Thanks a lot @trentmwillis for putting this together. The changes look good.

I do have one question, though. Wouldn't it make more sense for `grunt-eslint` to handle this? After all, they're a Grunt integration with ESLint and should be most aware of both sides. Shouldn't grunt-eslint detect Node version and/or ESLint version and export an empty task (possibly behind a silent failure option)? I feel like that would be the ideal way to handle this.

trentmwillis:
> Wouldn't it make more sense for `grunt-eslint` to handle this?

Not really. They actually bumped a major version (`18` -> `19`) when they moved to ESLint 3 (which is when Node 0.12 was dropped I believe).  So from that perspective, they shouldn't need to support consumers using things outside of semver ranges (IMO). Additionally, this gives us an easy way to no-op any other potential tasks that might lose 0.12 support prior to us droppping 0.12 (which should be beginning of next year).

I could alternatively, try using a version of grunt-eslint with ESLint 2 instead of adding this additional infrastructure.

platinumazure:
It just seemed like a lot of hoops to jump through in Grunt. I'm opposed to
using ESLint 2 because there have been many useful improvements in the 3.x
series, so if you believe the infrastructure you've created is the most
strategically sensible approach, I'll defer to your judgement.

trentmwillis:
Yes, definitely more hoops to jump through, but from a maintainer perspective, I wouldn't want the grunt lib to add code to support a "non-use" case.

cc/ @leobalter @mgol if this looks good to y'all, I'll merge

package.json
```diff
@@ -52,9 +52,8 @@
   },
   "scripts": {
     "browserstack": "sh build/run-browserstack.sh",
-    "ci": "grunt test",
-    "ci-include-lint": "grunt eslint && grunt test",
-    "test": "grunt",
+    "build": "grunt build",
+    "test": "grunt test",
```
leobalter: https://github.com/qunitjs/qunit/blob/master/test/index.html#L7

We still need to build the dist files for the tests, right?

leobalter: and now I look at the previous ci script and ... I'm missing something here.

trentmwillis: When `npm install` happens it runs the `prepublish` script which runs `grunt build`.

mgol: We shouldn't be using the `prepublish` hook at all, it's deprecated.

mgol: jQuery doesn't build on `grunt test` but it does on `npm test`:
https://github.com/jquery/jquery/blob/master/package.json#L63

I think it makes sense as it allows for the standard `npm install && npm test` (or `npm it` for short) workflow without diving into the project internals and without relying on a deprecated hook.

trentmwillis: Did not realize it was deprecated. Looks like it happened in npm 4. We could change it to the new `npm prepare` hook, but I think just including the build in the `test` command makes the most sense. Will update here shortly.

trentmwillis:
@mgol updated to remove the prepublish script. @leobalter do you know if we were relying on this for releases or e browserstack tests? Browserstack in particular failed 100%, but I am unsure if it is because we don't generate a build now.

leobalter:
It seems we need to add the build to the npm's `browserstack` script.

leobalter:
LGTM

leobalter:
Merged with the following commits:

- 172aa6366705d6a0a6f4497c557fd055574c4576
- 990c6985d1d27bfb348ab89847c751103cff256d
- da70f9d9e249897a6b213299e9c5c4dbbdc838de

Issue #1048, opened 2016-09-23
PhantomJS is blocking tests to run


leobalter:
```
    Running "qunit:qunit" (qunit) task
    Testing test/index.html
    PhantomJS threw an error:ERROR
    >> 0 [ '' ]
    Warning: PhantomJS exited unexpectedly with exit code null. Use --force to continue.

    Aborted due to warnings.
```

I have no idea what is causing this, some sort of ERROR is happening there where I can't reproduce in a browser test.

I guess it's due to my recent installed Mac OS Sierra.

It works on Ubuntu.

trentmwillis:
@leobalter did this ever resolve? Planning to update my Mac tomorrow, so I'll see if I can reproduce if you're still having issues.

leobalter:
I'm using my home mac for QUnit. I believe we will need to update the grunt Istanbul task by ourselves if want to fix it.

mgol:
I'm seeing the same, also in the Migrate repo.

The problem is that the task is defined in the [grunt-qunit-istanbul](https://github.com/asciidisco/grunt-qunit-istanbul) fork of the upstream [https://github.com/gruntjs/grunt-contrib-qunit](https://github.com/gruntjs/grunt-contrib-qunit) and the fork doesn't seem to be maintained anymore.

mgol:
The [grunt-qunit-istanbul](https://github.com/asciidisco/grunt-qunit-istanbul) fork is also blocking us from upgrading Grunt to version 1.x.

mgol:
...and it still relies on PhantomJS 1 while the upstream repo switched to version 2.

If the Istanbul stuff still can't be done without hacking the upstream `qunit` task we may need to maintain the fork ourselves.

trentmwillis:
I looked into this and it was a bear to figure out what was going on. But I have two PRs now that should enable us to move to Grunt 1.x and run on latest Mac OS:

- https://github.com/asciidisco/grunt-lib-phantomjs-istanbul/pull/11
- https://github.com/asciidisco/grunt-qunit-istanbul/pull/60

leobalter:
This issue is almost solved. Our current problem right now is to figure out a way to have code coverage and not being stuck on PhantomJS' race conditions.

platinumazure:
If we had an lcov reporter, could we use Istanbul?

trentmwillis:
[...]

PR #1079, opened 2016-12-09
Build: Re-introduce code coverage

trentmwillis:
Fixing the last part of #1048.

This takes a more explicit approach to coverage. It does the following:
1. Builds the library
2. Instruments the code
3. Copies the instrumented code to the build output so that it is used within the tests
4. Runs the tests
5. Restores the original build output
6. Stores the coverage and generates reports
7. Sends coverage to coveralls


trentmwillis:
@leobalter I think we may need to update the coveralls repo so that we can send the metrics (I bet it's still pointing to `jquery/qunit`).

leobalter:
closed with: 

- a46fbb53b8c25a8535cf991118c5d6a1838063aa
- 7dfe400719bb814dbd7becc5756d76fad63410b9
- 294ba438cb76edd61bfa5385002734440d1a676f
- 365f8d24524e1faaceb4c98b460d060abed172cc

Thanks for this amazing work!

dorgan:
So does this mean qunit, now has code coverage??  I am running into the same issues you are ran into, our projects use grunt-qunit-istanbul in order to do code coverage and now phantomjs doesnt work with sierra, would be great if qunit now has this natively.

leobalter:
we don't have any direct integration with coverage tools in our API. This PR only reapplied the coverage check for the own QUnit tests, testing QUnit.

leobalter:
Working on an open source base, where every maintainer have to find available free time for the project, we can't afford to ship any coverage tool right now. Some other priorities (like using QUnit in the node cli) are a bit more valuable at this current point. Coverage integration can still be done.

btw, if we have the cli ready, we could just call `istanbul qunit ...` that would be amazing

PR #1080, opened 2016-12-10
Core: Implement QUnit.todo

trentmwillis:
Supersedes #1006.

This takes a simpler approach to implementing `QUnit.todo`. We simply add a `todo` flag to `log` and `testDone` details that denotes if an assertion or a test was part of a `todo`. The reporter can then handle that scenario appropriately.

This approach keeps QUnit's internal logic simple and encourages reporters to report success/failure for a test suite based on the outcome of tests and _not_ on the outcome of assertions, which is inline with the thinking of js-reporters.

While this likely means most reporters will have to implement new logic to support `todo`, this change should not break existing reporters and provides a relatively clear path to supporting the new feature.

_Note: Assuming this implementation looks okay, I will open a PR to support `todo` in `grunt-contrib-qunit`._

### DIFF
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -14,24 +14,55 @@ module.exports = function( grunt ) {
 
 		var done = this.async();
 
-		async.series( runs, function( error, result ) {
-			var total = result.reduce( function( previous, details ) {
-					return previous + details.total;
-				}, 0 );
-			var failed = result.reduce( function( previous, details ) {
-					return previous + details.failed;
-				}, 0 );
-			var runtime = result.reduce( function( previous, details ) {
-					return previous + details.runtime;
-				}, 0 );
+		async.series( runs, function( error, stats ) {
+			var totals = stats.reduce( function( totals, stats ) {
+				totals.passed += stats.passed;
+				totals.failed += stats.failed;
+				totals.skipped += stats.skipped;
+				totals.todo += stats.todo;
+				totals.passedAssertions += stats.passedAssertions;
+				totals.failedAssertions += stats.failedAssertions;
+				totals.runtime += stats.runtime;
+				return totals;
+			}, {
+				passed: 0,
+				failed: 0,
+				skipped: 0,
+				todo: 0,
+				runtime: 0,
+				passedAssertions: 0,
+				failedAssertions: 0
+			} );
 
 			grunt.log.writeln( "-----" );
-			grunt.log.ok( total + " total assertions (in " + runtime + "ms) , " +
-				"with "  + failed + " failed assertions" );
+			grunt.log.ok( constructMessage( totals ) );
 
 			done( !error );
 		} );
 	} );
+
+	function constructMessage( stats ) {
+		var totalTests = stats.passed + stats.failed + stats.skipped + stats.todo,
+			totalAssertions = stats.passedAssertions + stats.failedAssertions;
+
+		return [
+			totalTests,
+			" tests completed with ",
+			stats.failed,
+			" failed, " +
+			stats.skipped,
+			" skipped, and ",
+			stats.todo,
+			" todo. \n" +
+			totalAssertions,
+			" assertions (in ",
+			stats.runtime,
+			"ms), passed: " +
+			stats.passedAssertions,
+			", failed: ",
+			stats.failedAssertions
+		].join( "" );
+	}
 
 	function runQUnit( file, runEnd ) {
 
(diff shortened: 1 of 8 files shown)


leobalter:
this looks amazing!

It might be a breaking change for the reporters but it's worth a major release for this, which we can bundle with a few other features, including the js-reporters.

trentmwillis:
I don't think we need a major version bump. As I said, this doesn't break any existing set ups; it should only cause problems if someone starts adding todo tests.

I actually think this gives a fairly good rollout strategy as reporters can be updated over time, rather than expecting them all to do it at once.

leobalter:
Got it! 

I'm +1 to merge it as soon we can get the updated the grunt plugin ready.

leobalter:
it is breaking in the logs file, this is weird. I need to spend a bit more time on this.

leobalter:
ok, got it. Had to remind what was happening here. I'll check how is the status on grunt-contrib-qunit

leobalter:
I asked @jzaefferer a little help in there and also write/publish access to the grunt plugin. I'll be able to update it as soon as we solve this.

leobalter:
sorry for the spam, but I finally got access to it! Thanks @tkellen!

reporter/html.js
```diff
@@ -1,6 +1,14 @@
 import QUnit from "../src/core";
 import { window, navigator } from "../src/globals";
 
+const stats = {
+	totalTests: 0,
```
gibson042: Could we document the relationships between these values? It's worth commenting that `totalTests` equals the sum of the other values, and maybe even keeping two variables (e.g., `let totalTests = 0; const testPartitions = {…}`).

build/tasks/test-on-node.js
```diff
@@ -66,31 +70,39 @@ module.exports = function( grunt ) {
...
 			var message = "name: " + details.name + " module: " + details.module +
 				" message: " + details.message;
 			grunt.log.error( message );
 		} );
-		QUnit.testDone( function() {
+		QUnit.testDone( function( details ) {
 			testActive = false;
+
+			if ( details.failed && !details.todo ) {
+				stats.failedTests++;
```
gibson042: This will fail to increment `failedTests` when a TODO test has no failing assertions. I think it should instead use the same logic as the HTML reporter.

reporter/html.js
trentmwillis: I'll remove `totalTests` from being tracked altogether. We only use it after the run has finished to display the number of tests that ran, so I just created a local variable at that point which clearly shows that it is the sum of the other categories.

reporter/html.js
```diff
@@ -823,14 +838,18 @@ QUnit.testDone( function( details ) {
...
+
 	testItem = id( "qunit-test-output-" + details.testId );
 
 	assertList = testItem.getElementsByTagName( "ol" )[ 0 ];
 
 	good = details.passed;
 	bad = details.failed;
 
-	if ( bad === 0 ) {
+	let failed = details.todo ? !details.failed : !!details.failed;
```
trentmwillis: Good suggestion. I'll updated this and dependent logic.

reporter/html.js
```diff
@@ -867,7 +887,24 @@ QUnit.testDone( function( details ) {
...
+			if ( failed ) {
+				stats.failedTests++;
+			} else {
+				stats.todoTests++;
+			}
+
+			let todoLabel = document.createElement( "em" );
+			todoLabel.className = "qunit-todo-label";
+			todoLabel.innerHTML = "todo";
+			testItem.insertBefore( todoLabel, testTitle );
```
trentmwillis: Yep, this helped simplify logic and keeps things clearer.

trentmwillis:
Updated with the suggestions from @gibson042 and opened a PR to update grunt-contrib-qunit to support `todo`: https://github.com/gruntjs/grunt-contrib-qunit/pull/137. Linked the two locally and verified that it works.

build/tasks/test-on-node.js
```diff
@@ -57,39 +88,63 @@ module.exports = function( grunt ) {
...
 			grunt.log.error( message );
 		} );
-		QUnit.testDone( function() {
+
+		QUnit.testDone( function( details ) {
 			testActive = false;
+
+			if ( details.skipped ) {
+				stats.skipped++;
+			} else if ( details.failed && !details.todo ) {
```
gibson042: Todo tests can also "fail". I think this would be clearest if it mirrored the HTML reporter:
```js
// This test passed if it has no unexpected failed assertions
var testPassed = details.failed > 0 ? details.todo : !details.todo;

if ( details.skipped ) {
	stats.skipped++;
} else if ( !testPassed ) {
	stats.failed++;
} else if ( details.todo ) {
	stats.todo++;
} else {
	stats.passed++;
}
```

trentmwillis:
Updated once more. @leobalter when you have a chance, would appreciate your help reviewing and landing this (will require a new grunt-contrib-qunit release first).

leobalter:
LGTM.

We can already include the expected grunt-contrib-qunit 1.3.0

trentmwillis:
Updated with the dependency update; tests are now passing!

reporter/html.js
```diff
@@ -861,20 +881,35 @@ QUnit.testDone( function( details ) {
...
 		testItem.insertBefore( time, assertList );
 	}
 
+	if ( !testPassed ) {
+		stats.failedTests++;
+	} else if ( details.todo ) {
+		stats.todoTests++;
+	} else {
+		stats.passedTests++;
+	}
```
gibson042: This need to be moved up inside the `else` block so skipped tests don't increment two `stats` counters.

trentmwillis: Good catch.

trentmwillis:
Updated again. Unless any other issues are found, I think this should be good to land (will open corresponding documentation PR soon)

leobalter:
:shipit: (with the docs)

trentmwillis:
Corresponding docs PR is now open: https://github.com/qunitjs/api.qunitjs.com/pull/146

@gibson042 if this seems good to you, we can merge (and finally close this feature out!)

leobalter:
Waiting for @gibson042 to confirm the changes after his feedback, LGTM.

Issue #1081, opened 2016-12-11
Not 'loading' properly on Node


leobalter:
We need to make this more friendly on Node. 

@getify helped me finding this bug.

As you can see [here](https://github.com/qunitjs/qunit/blob/master/build/tasks/test-on-node.js#L54-L55), unless we call `QUnit.load()` before `QUnit.start()`, QUnit won't run properly in Node.

We need to make `QUnit.start` a self method to _load_ and start QUnit without calling anything else. This is the expected behavior reading the api docs.

The good part is that we can remove any sights of 1.x warnings of QUnit.start along.

---

As I can't commit to any code today, I'm planning to fix this tomorrow - Dec 12 - in the evening (EST), high priority, unless someone wants to fix it before, please just add a comment here.

getify:
Will this be in a patch release (2.1.1)? Any ETA?

leobalter:
Yes, this week.

PR #1083, opened 2016-12-14
QUnit.load is now called inside QUnit.start

leobalter:
Only in non-browser environments, avoiding breaking changes.

Closes gh-1081

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -121,6 +121,10 @@ extend( QUnit, {
 			} else if ( config.autostart ) {
 				throw new Error( "Called start() outside of a test context when " +
 					"QUnit.config.autostart was true" );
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -121,6 +121,10 @@ extend( QUnit, {
 			} else if ( config.autostart ) {
 				throw new Error( "Called start() outside of a test context when " +
 					"QUnit.config.autostart was true" );
+			} else if ( !defined.document && !config.pageLoaded ) {
+
+				// Starts from Node even if .load was not previously called
+				QUnit.load();
```
trentmwillis: I'm assuming we'll refactor this later due to #1084.

leobalter: Yeah, it's not beautiful but it's a fix for now.

PR #1085, opened 2016-12-14
Reporter: move urlparams import from reporter to html

### LINKED ISSUE #1082: Test runner UI checkbox settings not synced with query parameters
When specifying values for the test runner's query params they aren't synced up to the checkboxes for the corresponding settings.  This affects the hidepassed query param more than the others, because the functionality is directly tied to the box being checked as opposed to the presence of the query param itself.

This problem can be replicated using [the basic example](http://qunitjs.com/cookbook/#automating-unit-testing) and was happening in multiple browsers on multiple platforms. After creating the HTML file, open the file in a browser. When using version 2.0.1 of QUnit, the checkboxes are synced up correctly with the query parameters. When using version 2.1.0 of QUnit, the syncing doesn't work the same way. For example, when using 2.1.0, select the 'Hide Passed Tests' checkbox, then refresh the page. The checked value is lost. Selecting the 'No try-catch' will immediately refresh the page, and will drop the checked value of that checkbox as part of the refresh.

dustinspecker:
> 80a14ee5ac49c7f0981884a6cbb5786dea3a9c8b implemented Rollup.
> Reporter is importing `html` before `urlparams`, but `urlparams` needs
> to be loaded before `html` as it was in the previous commit.
> 
> Tried reordering the imports within the file, but running the build task
> seemed to not change the output. By moving the `urlparams` import
> into `html`, the built output is correct.
> 
> Fixes #1082 

Does anyone know why reordering the imports in https://github.com/qunitjs/qunit/blob/master/reporter/reporter.js doesn't have the desired effect?

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -1,5 +1,6 @@
 import QUnit from "../src/core";
 import { window, navigator } from "../src/globals";
+import "./urlparams";
 
 // Escape text for attribute or text content.
 export function escapeText( s ) {
--- a/reporter/reporter.js
+++ b/reporter/reporter.js
@@ -1,3 +1,2 @@
 import "./diff";
 import "./html";
-import "./urlparams";


leobalter:
from a es module perspective it should not have effects or it would be a potential bad practice.

platinumazure:
Can we rewrite urlparams to export an actual module (or set of functions, or whatever) so that whoever is supposed to consume the logic just does so conventionally? It's not great that we're importing the module for side effects, honestly. Even if it just exported an "initializeUrlParamHandling" function and we called that from the HTML reporter, that would feel a lot better to me.

leobalter:
> It's not great that we're importing the module for side effects, honestly.

Exactly. It does not need to be on this PR, but for sure in a follow up.

dustinspecker:
Not trying to rush or be rude, but when does QUnit typically release bug fixes?

leobalter:
This is all volunteered work. The end of 2016 was tough and I believe the same or even other reasons happened to the other maintainers. Although, I understand it's our responsibility to be reasonably fast with our releases and I appreciate your patience and the not rush neither rude message. Feel free to ping us if it's taking too long.

I'm releasing it right now.

PR #1087, opened 2016-12-17
Core: Implement event emitter

trentmwillis:
**Do NOT merge until after 2.1.1 is released.**

This implements a simple event emitter interface that will eventually supersede the current logging callbacks. No tests are added here since the `emit` function is private in scope and not currently wired up internally. One goal here is to avoid a giant PR that both implement the event emitter and wires it up to all possible events.

A follow-up commit to this will use the API implemented here to emit data according to the [js-reporters standard](https://github.com/js-reporters/js-reporters#event-data).

### DIFF
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -28,19 +28,16 @@ export function diff( a, b ) {
 	return result;
 }
 
-// From jquery.js
+/**
+ * Determines whether an element exists in a given array or not.
+ *
+ * @method inArray
+ * @param {Any} elem
+ * @param {Array} array
+ * @return {Boolean}
+ */
 export function inArray( elem, array ) {
-	if ( array.indexOf ) {
-		return array.indexOf( elem );
-	}
-
-	for ( var i = 0, length = array.length; i < length; i++ ) {
-		if ( array[ i ] === elem ) {
-			return i;
-		}
-	}
-
-	return -1;
+	return array.indexOf( elem ) !== -1;
 }
 
 /**
--- a/src/dump.js
+++ b/src/dump.js
@@ -60,7 +60,7 @@ export default ( function() {
 			parse: function( obj, objType, stack ) {
 				stack = stack || [];
 				var res, parser, parserType,
-					inStack = inArray( obj, stack );
+					inStack = stack.indexOf( obj );
 
 				if ( inStack !== -1 ) {
 					return "recursion(" + ( inStack - stack.length ) + ")";
@@ -190,7 +190,7 @@ export default ( function() {
 					nonEnumerableProperties = [ "message", "name" ];
 					for ( i in nonEnumerableProperties ) {
 						key = nonEnumerableProperties[ i ];
-						if ( key in map && inArray( key, keys ) < 0 ) {
+						if ( key in map && !inArray( key, keys ) ) {
 							keys.push( key );
 						}
 					}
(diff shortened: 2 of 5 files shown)


src/dump.js
```diff
@@ -60,7 +60,7 @@ export default ( function() {
 			parse: function( obj, objType, stack ) {
 				stack = stack || [];
 				var res, parser, parserType,
-					inStack = inArray( obj, stack );
+					inStack = stack.indexOf( obj );
```
platinumazure: Should this say `stack.indexOf( obj ) !== -1`?

trentmwillis: Nope, we need the actual index value later. Renamed this variable to make it clearer as to what is going on.

trentmwillis:
> Is there any way we can use an existing event emitter? I know QUnit doesn't like to have hard dependencies, but could Rollup be configured to bundle in certain dependencies and then we wouldn't have to roll our own EventEmitter?

I thought about this, but we don't need a super robust eventing system and we would still likely need to wrap the library in some of our own logic to verify usage according to what we expect. Therefore, since the actual code is only like 30 lines, I figured we should just maintain it ourselves. Seem good?

src/dump.js
```diff
@@ -60,10 +60,10 @@ export default ( function() {
 			parse: function( obj, objType, stack ) {
 				stack = stack || [];
 				var res, parser, parserType,
-					inStack = inArray( obj, stack );
+					objIndex = stack.indexOf( obj );
 
-				if ( inStack !== -1 ) {
-					return "recursion(" + ( inStack - stack.length ) + ")";
+				if ( objIndex !== -1 ) {
+					return "recursion(" + ( objIndex - stack.length ) + ")";
```
platinumazure: This would probably work better as a template literal, since you're using them elsewhere in this PR.

trentmwillis: Good point! Will update.

trentmwillis:
Updated. Thanks for the review.

trentmwillis:
@platinumazure does this look good to you now?

@leobalter any update about releasing 2.1.1?

src/core/utilities.js
```diff
@@ -28,19 +28,16 @@ export function diff( a, b ) {
...
-	}
-
-	for ( var i = 0, length = array.length; i < length; i++ ) {
-		if ( array[ i ] === elem ) {
-			return i;
-		}
-	}
-
-	return -1;
+	return array.indexOf( elem ) !== -1;
```
leobalter: in a close future, hopefully, we'll be able to use `[].includes`.

src/events.js
```diff
@@ -0,0 +1,67 @@
...
+ * @return {Void}
+ */
+export function emit( eventName, data ) {
+	if ( objectType( eventName ) !== "string" ) {
+		throw new TypeError( "eventName must be a string when emitting an event" );
+	}
+
+	// Clone the callbacks in case one of them registers a new callback
+	let originalCallbacks = LISTENERS[ eventName ];
+	let callbacks = originalCallbacks ? [ ...originalCallbacks ] : [];
```
leobalter: It doesn't improve readability, but I was playing with the syntax: `[ ...(originalCallbacks || []) ]`. 

I still prefer your code. 👍

leobalter:
LGTM

PR #1088, opened 2016-12-17
Assert: ES6 Refactor

trentmwillis:
Potentially unneeded, but as we move forward and start introducing new features (such as Step assertions, Todo tests, new reporters, etc.), it'd be nice for the codebase to be modernized and cleaned up. I personally find this helps lower the barrier to entry for new contributors if the codebase is easy to read.

A couple things to discuss:
1. Do we need to do refactoring separately (like this PR) or should we let it happen over time? I tend to think letting it happen over time will tend towards it not happening at all.
2. Is there an ES6 style guide for jQuery? If not, we should define eslint rules for things like `let` vs `const` as they come up.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -26,8 +26,7 @@ class Assert {
 
 	// Put a hold on processing and return a function that will release it a maximum of once.
 	async( count ) {
-		var resume,
-			test = this.test,
+		let test = this.test,
 			popped = false,
 			acceptCallCount = count;
 
@@ -36,15 +35,15 @@ class Assert {
 		}
 
 		test.usedAsync = true;
-		resume = internalStop( test );
+		let resume = internalStop( test );
 
 		return function done() {
-
 			if ( popped ) {
 				test.pushFailure( "Too many calls to the `assert.async` callback",
 					sourceFromStacktrace( 2 ) );
 				return;
 			}
+
 			acceptCallCount -= 1;
 			if ( acceptCallCount > 0 ) {
 				return;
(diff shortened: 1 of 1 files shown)


platinumazure:
Note that another way we can do ES6 refactors, in some cases, is to enable an ESLint rule which enforces a certain style or syntax and then fix all the violations at once. Please let me know if you are looking for strategies around specific refactoring goals and I can let you know if ESLint can help. Thanks!

src/assert.js
```diff
@@ -98,122 +97,131 @@ Assert.prototype = {
...
 		return assert.test.pushResult( resultInfo );
-	},
+	}
+
+	ok( result, message ) {
+		if ( !message ) {
+			message = result ?
+				"okay" :
+				`failed, expected argument to be truthy, was: ${dump.parse( result )}`;
+		}
```
leobalter:
```js
ok( result, message = "okay" ) {
  if ( !result ) {
    message = `failed, expected argument to be truthy, was: ${dump.parse( result )}`;
  }
  ...
}
```

platinumazure: @leobalter Not sure that will work. A custom message (two arguments passed) will always be clobbered by that code when the assertion value is falsy. Think the if condition should check for undefined.

src/assert.js
```diff
@@ -238,59 +246,63 @@ Assert.prototype = {
...
-
+// eslint-disable-next-line dot-notation
+Assert.prototype.raises = Assert.prototype[ "throws" ];
+
+/**
+ * Converts an error into a simple string for comparisons.
+ *
+ * @param {Error} error
+ * @return {String}
+ */
```
leobalter: should we keep it this way?

leobalter:
There's nothing here breaking, fixing a bug or introducing a new feature, everything is just using some new language features. I like this PR and don't think it's needed, it will open a discussion for us to discuss a potential style guide.

> Do we need to do refactoring separately (like this PR) or should we let it happen over time? I tend to think letting it happen over time will tend towards it not happening at all.

I'll solve this problem in a couple minutes.

> Is there an ES6 style guide for jQuery? If not, we should define eslint rules for things like let vs const as they come up.

Not that I am aware of, but I believe it's a nice opportunity for us to see if we want to define our own. Defining our own specific eslint rules, even if it's the same - or almost - of the jQuery rules, is inevitable.

Issue #1090, opened 2017-01-18
QUnit 2.x cannot be run in a sandboxed iframe due to direct reference to window.sessionStorage


Gaurav0:
See https://github.com/ember-cli/ember-twiddle/issues/516

rwjblue:
Thanks for opening @Gaurav0, hopefully some of the core devs can chime in and confirm if this was intentional...

trentmwillis:
We did some refactoring to introduce `config.storage` recently, so likely either https://github.com/qunitjs/qunit/pull/1042 or https://github.com/qunitjs/qunit/pull/1055. Definitely a bug; @Gaurav0 if you have time, would you mind digging in and trying to figure out what's wrong?

PR #1091, opened 2017-01-19
Core: Ability to run in sandboxed iframe, (#1090)

Gaurav0:
Do not directly refer to window.sessionStorage (#1090)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -23,6 +23,14 @@ function process( code ) {
 
 grunt.initConfig( {
 	pkg: grunt.file.readJSON( "package.json" ),
+	connect: {
+		server: {
+			options: {
+				port: 8000,
+				base: "."
+			}
+		}
+	},
 	copy: {
 		options: { process: process },
 
@@ -96,6 +104,9 @@ grunt.initConfig( {
 			inject: [
 				path.resolve( "./build/coverage-bridge.js" ),
 				require.resolve( "grunt-contrib-qunit/phantomjs/bridge" )
+			],
+			urls: [
+				"http://localhost:8000/test/sandboxed-iframe.html"
 			]
 		},
 		qunit: [
@@ -202,7 +213,8 @@ grunt.event.on( "qunit.coverage", function( file, coverage ) {
 
 grunt.loadTasks( "build/tasks" );
 grunt.registerTask( "build", [ "rollup:src", "copy:src-js", "copy:src-css" ] );
-grunt.registerTask( "test", [ "eslint", "search", "test-on-node", "qunit" ] );
+grunt.registerTask( "qunit-test", [ "connect", "qunit" ] );
+grunt.registerTask( "test", [ "eslint", "search", "test-on-node", "qunit-test" ] );
 grunt.registerTask( "coverage", [
 	"build",
 	"instrument",
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "grunt": "1.0.1",
     "grunt-cli": "1.2.0",
     "grunt-concurrent": "2.3.1",
+    "grunt-contrib-connect": "1.0.2",
     "grunt-contrib-copy": "1.0.0",
     "grunt-contrib-qunit": "1.2.0",
     "grunt-contrib-watch": "1.0.0",
--- a/src/globals.js
+++ b/src/globals.js
@@ -7,4 +7,14 @@ export const clearTimeout = global.clearTimeout;
 
 export const document = window && window.document;
 export const navigator = window && window.navigator;
-export const sessionStorage = window && window.sessionStorage;
+
+export const sessionStorage = ( function() {
+	var x = "qunit-test-string";
+	try {
+		sessionStorage.setItem( x, x );
+		sessionStorage.removeItem( x );
+		return sessionStorage;
+	} catch ( e ) {
+		return undefined;
+	}
+}() );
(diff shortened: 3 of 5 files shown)


Gaurav0:
@platinumazure I have added the unit test as requested. Some changes to the test harness were required.

platinumazure:
@trentmwillis @leobalter Can one of you please review this? I inadvertently requested a change which resulted in an overhaul of the test harness, is that change okay?

test/sandbox-inside.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<title>QUnit Only Test Suite</title>
```
trentmwillis: Nitpick: Let's update the title here

src/globals.js
```diff
@@ -7,4 +7,19 @@ export const clearTimeout = global.clearTimeout;
 
 export const document = window && window.document;
 export const navigator = window && window.navigator;
-export const sessionStorage = window && window.sessionStorage;
+
+let defined = {
+	sessionStorage: ( function() {
```
trentmwillis: Let's just export the result directly:
```js
export const sessionStorage = ( function() { ... }() );
```
No need to have an intermediate value.

package.json
```diff
@@ -40,6 +40,7 @@
     "grunt": "1.0.1",
     "grunt-cli": "1.2.0",
     "grunt-concurrent": "2.3.1",
+    "grunt-contrib-connect": "^1.0.2",
```
leobalter: conventionally, we lock the package versions. Can you remove the `^` here, please?

src/globals.js
```diff
@@ -7,4 +7,19 @@ export const clearTimeout = global.clearTimeout;
...
+		var x = "qunit-test-string";
+		try {
+			sessionStorage.setItem( x, x );
+			sessionStorage.removeItem( x );
+			return sessionStorage;
```
leobalter: we could use something better than a try catch, something like the `in` operator as `return 'setItem' in sessionStorage && 'removeItem' in sessionStorage && sessionStorage`

This way we also don't need to catch any false alarms if sessionStorage doesn't work on storing or removing an item. All we want is to confirm the API is available.

test/sandbox.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
leobalter: sandbox and sandbox-inside doesn't play well with the other html files in this folder. It's repeating another test and the intent is implicit only. After this PR one would look and ask why this is here. The same can be asked when reading the Gruntfile.

naming is hard, be we should try again w/ this one. Maybe sessionstorage-through-iframe.html is long but good enough, I'm bad with names too.

Gruntfile.js
```diff
@@ -10,6 +10,7 @@ module.exports = function( grunt ) {
 
 // Load grunt tasks from NPM packages
 require( "load-grunt-tasks" )( grunt );
+grunt.loadNpmTasks( "grunt-contrib-connect" );
```
leobalter: I believe this can be discarded as the line above should handle this.

src/globals.js
Gaurav0: Unfortunately it blows up in the sandbox if we do that.

Gruntfile.js
Gaurav0: Thanks!

package.json
Gaurav0: Done.

src/globals.js
Gaurav0: Yes, makes sense.

test/sandbox-inside.html
Gaurav0: Done.

test/sandbox.html
Gaurav0: I'm open to suggestions, but don't want a name specific to sessionStorage issue, this test catches any error when loading QUnit through an iframe.

trentmwillis: Can we do `sanboxed-iframe` and `sandboxed-iframe-contents`? I think that is a good balance between overly specific and generic

Gaurav0: Yes!

trentmwillis:
Looks good to me! @leobalter / @platinumazure any other concerns?

platinumazure:
Oh, actually, one suggestion. Do we need to load a test file at all actually? Maybe the iframe content file could just use an inline `<script>` tag and just do a quick inline test with an `assert.ok(true)` or something?

src/globals.js
```diff
@@ -7,4 +7,14 @@ export const clearTimeout = global.clearTimeout;
...
-export const sessionStorage = window && window.sessionStorage;
+
+export const sessionStorage = ( function() {
+	var x = "qunit-test-string";
+	try {
+		sessionStorage.setItem( x, x );
+		sessionStorage.removeItem( x );
+		return sessionStorage;
+	} catch ( e ) {
+		return undefined;
```
leobalter: I don't like this try catch block, but that's something I can investigate later how to improve it.

trentmwillis: Agreed, but given the nature of the error I think it unavoidable.

src/globals.js
```diff
@@ -7,4 +7,14 @@ export const clearTimeout = global.clearTimeout;
 
 export const document = window && window.document;
 export const navigator = window && window.navigator;
-export const sessionStorage = window && window.sessionStorage;
+
+export const sessionStorage = ( function() {
```
stefanpenner: this `const sessionStorage` shadows the global one, forcing the `sessionStorage` in the function scope (line 14) to always be undefined.

stefanpenner: one approach to fixing: https://github.com/qunitjs/qunit/pull/1112

PR #1093, opened 2017-01-19
Fixes #1092

### LINKED ISSUE #1092: Error in IE: "Unable to get value of the property 'warn': object is null or undefined"
There is no `console` object in IE (at least 8-9) if console is not enabled.

Invis1ble:
This commit should fix error in IE 8-9.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -57,6 +57,8 @@ class Assert {
 	// Exports test.push() to the user API
 	// Alias of pushResult.
 	push( result, actual, expected, message, negative ) {
+		// Supports IE9, not throwing if console is not active.
+		// Ref #1092
 		if ( console ) {
 			console.warn( "assert.push is deprecated and will be removed in QUnit 3.0." +
 				" Please use assert.pushResult instead (http://api.qunitjs.com/pushResult/)." );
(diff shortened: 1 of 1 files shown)


trentmwillis:
Thanks for this PR. Would you mind actually making this a bit more robust? I'm thinking we should have a logging singleton which abstracts away these guards. Something like:

```js
const Logger = {
  log(...args) {
    if (global.console) {
      console.log(...args);
    }
  }
};
```

Then in our code we can simply do:

```js
Logger.log('foo');
// or
Logger.warn('bar');
```

src/assert.js
```diff
@@ -57,8 +57,10 @@ class Assert {
 	// Exports test.push() to the user API
 	// Alias of pushResult.
 	push( result, actual, expected, message, negative ) {
-		console.warn( "assert.push is deprecated and will be removed in QUnit 3.0." +
-			" Please use assert.pushResult instead (http://api.qunitjs.com/pushResult/)." );
+		if ( console ) {
```
leobalter: We should add a comment in the line above `// Supports IE9, not throwing if console is not active. Ref gh-1092`

trentmwillis:
> I like the idea from @trentmwillis to introduce a Logger api but that might come as a follow up work.

Sounds good to me. I'll do that after this lands.

src/assert.js
Invis1ble: I've added comment. Sorry for multiple commits.

trentmwillis:
Thanks for the contribution @Invis1ble!

Issue #1075, opened 2016-12-06
step API


leobalter:
@wycatz suggested this step API and I'm game for it.

It records steps inside a test. Given this example:

```js
QUnit.test('references are granular', assert => {
  let { SubSub } = MakeSub();

  let obj = SubSub.create({ loud: true });
  let root = rootFor(obj);

  let state = new State(root);

  set(obj, 'name', 'Daniel');

  step('Update name');

  state.validateTags({ name: false });
  state.validateValues({ name: 'Daniel' });
  state.update();

  set(obj, 'loud', false);

  step('Update loud');

  state.validateTags({ loud: false });
  state.validateValues({ loud: false });
  state.update();
});
```

Each step can be recorded as a `assert.ok(true)`, but printing the given description.

There's one con to consider for this API, which is the idea that assertions can be split in different tests specially inside nested modules.

Even though, it is an interesting approach to consider.

leobalter:
I did a quick prototype here: https://github.com/qunitjs/qunit/compare/master...leobalter:step?expand=1

trentmwillis:
I like this (and think there are several use cases within our own test suite for it), but seems like it is missing one thing: asserting that the correct steps happened.

Seems like maybe we need `assert.step( tag )` and then `assert.verifySteps([ tag1, tag2, ... ])`.

leobalter:
That's a great and useful addition to the step api. I like it.

platinumazure:
I'm not opposed to this landing, but I am not sure I understand the use case fully. I see the example above, but I'm not sure I fully understand what advantage this has over `assert.ok(true)` besides a slight readability improvement. Can anyone enlighten me, please?

leobalter:
Improves readability and it's also motivated by occurrence. `ok(true)` is being used here and there to flag given steps and their order. While this could be solved using the current API, `step` and `verifyStep` will provide an API with an clear goal to be used in these cases.

Afaik, this case was frequent in some async jquery tests and will certainly gonna be used by Ember as this is an original request from Yehuda Katz.

platinumazure:
@leobalter Thanks for the explanation, that works for me. That said, I'd love to see a longer-form case study for steps and step verification show up in the QUnit Cookbook maybe (unless that was covered in the documentation PRs?). If that seems worthwhile, I can create an issue.

leobalter:
it is! Using async code is one of the nice ways to use for examples. Async fns and promises will be great for that.

PR #1094, opened 2017-01-20
Assert: Add step and verifySteps functions

trentmwillis:
As discussed in #1075. Relevant commit messages:

> Introduces the assert.step function which logs a passing assertion
> for steps in a test. These steps are stored as string values to
> help verify order of operations.

> Introduces the assert.verifySteps function to complement assert.step.
> This acceptances an array of string values to compare to the steps
> that have executed for the current test.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -28,6 +28,11 @@ class Assert {
 		} );
 	}
 
+	// Verifies the steps in a test match a given array of string values
+	verifySteps( steps, message ) {
+		this.deepEqual( this.test.steps, steps, message );
+	}
+
 	// Specify the number of expected assertions to guarantee that failed test
 	// (no assertions are run at all) don't slip through.
 	expect( asserts ) {
--- a/src/assert.js
+++ b/src/assert.js
@@ -14,6 +14,20 @@ class Assert {
 
 	// Assert helpers
 
+	// Documents a "step", which is a string value, in a test as a passing assertion
+	step( message ) {
+		let result = !!message;
+
+		message = message || "You must provide a message to assert.step";
+
+		this.test.steps.push( message );
+
+		return this.pushResult( {
+			result,
+			message
+		} );
+	}
+
 	// Specify the number of expected assertions to guarantee that failed test
 	// (no assertions are run at all) don't slip through.
 	expect( asserts ) {
--- a/test/main/assert/step.js
+++ b/test/main/assert/step.js
@@ -34,3 +34,25 @@ QUnit.test( "pushes a passing assertion if a message is given", function( assert
 	assert.step( "One step" );
 	assert.step( "Two step" );
 } );
+
+QUnit.module( "assert.verifySteps" );
+
+QUnit.test( "verifies the order and value of steps", function( assert ) {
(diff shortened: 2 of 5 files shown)


leobalter:
LGTM, I was not so sure about counting steps for `expect` but reading the original issue again I can see they defer to some `ok(true)` anyway.

Can we have the docs for it as well?

trentmwillis:
> I was not so sure about counting steps for `expect`

My reasoning is this: I think it useful for situations where you may want to "count" the number of times something happens. For example:

```js
test('some async event', function(assert) {
  assert.expect(2);
  obj.someHook = function() {
    assert.step('Some hook called');
  };
  obj.shouldInvokeSomeHookTwice();
});
```

I'll open the docs PR soon and then we can merge them together.

trentmwillis:
@leobalter documentation PR has been opened at https://github.com/qunitjs/api.qunitjs.com/pull/142

src/assert.js
```diff
@@ -14,6 +14,25 @@ class Assert {
...
+
+		return this.pushResult( {
+			result,
+			message
+		} );
+	}
+
+	// Verifies the steps in a test match a given array of string values
+	verifySteps( steps, message ) {
+		this.deepEqual( steps, this.test.steps, message );
```
gibson042: The first two arguments should be swapped for our standard (actual, expected [, message ]) signature.

test/main/assert/step.js
```diff
@@ -0,0 +1,44 @@
...
+
+	var originalPushResult = assert.pushResult;
+	assert.pushResult = function pushResultStub( resultInfo ) {
+		assert.pushResult = originalPushResult;
+
+		assert.notOk( resultInfo.result );
+		assert.equal( resultInfo.message, "You must provide a message to assert.step" );
+	};
+
+	assert.step();
```
gibson042: I think it's more important to cover empty-string input, but we should probably include both in the test suite.

src/assert.js
trentmwillis: Good point. Will update.

test/main/assert/step.js
trentmwillis: Will do.

trentmwillis:
Updated to address @gibson042's comments

test/main/assert/step.js
```diff
@@ -0,0 +1,58 @@
...
+
+	var originalPushResult = assert.pushResult;
+	assert.pushResult = function pushResultStub( resultInfo ) {
+		assert.pushResult = originalPushResult;
+
+		assert.notOk( resultInfo.result );
+	};
+
+	assert.verifySteps( [ "One step", "Red step", "Two step", "Blue step" ] );
+} );
```
leobalter: how do we assert w/ verifyStep after `step()` and `step("")`? This should not block this PR as it's verify named steps, and I don't think we should cast undefined to be the empty string or anything, but it's good to know in case this come as a question in any future.

trentmwillis: I'll do a follow up and add a test case for that and we can determine what the expected behavior should be.

PR #1076, opened 2016-12-06
Build: Add yarn.lock

leobalter:
In case we want to use [yarn](https://yarnpkg.com/), we can start with this lock file.

PR #1095, opened 2017-01-20
Build: Add yarn

trentmwillis:
Now that we're off of Node 0.12, we should be able to use Yarn for all CI builds. Hopefully this will make those runs even faster.

Should supersede #1076.


leobalter:
That's great, thanks!

PR #1096, opened 2017-01-20
Core: Add assertion event

trentmwillis:
Implements the assertion event according to js-reporters standard. Also, lays the test scaffolding for the other events to be implemented soon.


platinumazure:
I'm a bit new to yarn. Is yarn.lock something we should be updating automatically somehow?

trentmwillis:
@platinumazure not really, we just need to make sure we do [dependency changes through `yarn`](https://yarnpkg.com/en/docs/yarn-lock#toc-managed-by-yarn). We added the `yarn.lock` file and then landed a new dependency in #1091; that dependency wasn't add via `yarn add` though so the lockfile didn't update.

For additional info: https://yarnpkg.com/en/docs/yarn-lock

platinumazure:
Okay, so in future we should use `yarn add` to add dependencies and then `yarn.lock` changes as dependencies are added. Makes sense, thanks for clarifying.

PR #1097, opened 2017-01-21
Core: Add logger singleton to handle messages

trentmwillis:
As brought up in https://github.com/qunitjs/qunit/pull/1093, this adds a `Logger` singleton to handle proxying to `console` methods to abstract potential logic branches.

Unsure how useful it will be in the long run, but for now it makes for simpler code (especially as we plan to deprecate things (e.g., `assert.push`, `QUnit.load`)


trentmwillis:
@platinumazure, can do as a follow up

PR #1098, opened 2017-01-24
Assert: Use actual values for verifySteps on failed steps

trentmwillis:
As brought up in https://github.com/qunitjs/qunit/pull/1094#discussion_r97558254. When calling `assert.verifySteps()` after a failed `assert.step()`, the bad values will be used in the comparison, even though `You must provide a message to assert.step` is the logged message.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -18,13 +18,11 @@ class Assert {
 	step( message ) {
 		let result = !!message;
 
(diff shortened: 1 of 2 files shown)


trentmwillis:
@leobalter will wait for you to confirm/deny if this is the behavior you expect.

leobalter:
Sorry for the delay! +1

PR #1099, opened 2017-01-25
Refactoring onerror logic into something exportable/reusable/testable

platinumazure:
Refs gh-1061.

I've extracted and exported a `QUnit.onError` function whose sole task is to do the following:

1. Check for a current configuration/test
1. If it exists, check for ignoreGlobalErrors, and if that's false, call that test's `pushFailure`
1. If no configuration/test exists, create a new test and call `pushFailure` inside that test

This is, in my view, the core logic for handling unhandled exceptions. Everything else in the file is working with `window.onerror`, ~~which I'm hoping to move to the HTML Reporter in a future commit.~~ which I've moved into the HTML reporter.

~~Wanted to get early feedback to make sure this direction seems sensible. I could also use some help in testing errors thrown outside of tests, since I don't have coverage for that yet.~~ This should be about 90% ready, just want to solidify the API for `QUnit.onError`.


src/core.js
```diff
@@ -164,7 +164,9 @@ extend( QUnit, {
 	stack: function( offset ) {
 		offset = ( offset || 0 ) + 2;
 		return sourceFromStacktrace( offset );
-	}
+	},
+
+    onError
```
platinumazure: Oops, need to tabify this line. I can't wait until we can get ESLint indent working...

Gruntfile.js
```diff
@@ -145,6 +146,7 @@ grunt.initConfig( {
 			"test/main/modules",
 			"test/main/deepEqual",
 			"test/main/stack",
+            "test/onerror/onerror",
```
platinumazure: Oops, need to tabify this line. I can't wait until we can get ESLint indent working...

Gruntfile.js
```diff
@@ -125,6 +125,7 @@ grunt.initConfig( {
 			"test/reporter-html/single-testid.html",
 			"test/reporter-urlparams.html",
 			"test/moduleId.html",
+            "test/onerror/onerror.html",
```
platinumazure: Oops, need to tabify this line. I can't wait until we can get ESLint indent working...

test/onerror/onerror.js
```diff
@@ -0,0 +1,28 @@
...
+
+		QUnit.config.current.pushFailure = function() {
+			assert.ok( false, "No error should be pushed" );
+		};
+
+		QUnit.config.current.ignoreGlobalErrors = true;
+
+		var result = QUnit.onError( "Error message", "filePath.js", 1 );
+
+		assert.strictEqual( result, true, "onError should allow other error handlers to run" );
```
platinumazure: Bad assert message, oops.

platinumazure:
Open questions at the moment:

1. Is it possible to test QUnit.onError when it's called outside a test context?
1. Thoughts on the name and parameter list for QUnit.onError (since this is a new API, we do not need to conform to window.onerror)
1. So far, this PR seems to be leaning in the direction that reporters should be responsible for linking up the unhandled error event with QUnit's handler. Does this seem reasonable?

src/core/onerror.js
```diff
@@ -1,43 +1,23 @@
-import { window } from "../globals";
```
gibson042: :clap: :clap: :clap:

src/core/onerror.js
```diff
@@ -1,43 +1,23 @@
...
+// Handle an unhandled exception. By convention, returns true if further
+// error handling should be suppressed and false otherwise.
+// In this case, we will only suppress further error handling if the
+// "ignoreGlobalErrors" configuration option is enabled.
+export default function onError( errorMessage, filePath, lineNumber ) {
+	if ( config.current ) {
+		if ( config.current.ignoreGlobalErrors ) {
+			return true;
 		}
+		pushFailure( errorMessage, filePath + ":" + lineNumber );
```
gibson042: In a similar vein, we might dump all excess arguments here too.

platinumazure:
@gibson042 Thanks for the review. I've pushed a commit to avoid dropping arguments in the areas you've called out.

Regarding testing for `QUnit.onError` outside of tests, I tried setting autostart to false and that didn't really work (I also found that I could just call `QUnit.onError` from inside a module but outside of a test for mostly the same effect). Unfortunately, I've run into some problems:

1. `QUnit.test` can't be duck-punched (internally, `onError` references that function through a closure).
1. I couldn't figure out how to duck-punch `QUnit.config.current.pushFailure` like I did in the other test file: I couldn't think of a way to access `QUnit.config.current` between when the "global failure" test is created (in `QUnit.onError`) and when that test's callback would run.
1. Duck-punching any of the `QUnit.assert` methods won't work because `pushFailure` does not actually call any assertion methods (it emits a direct report, then pushes the assertion object onto the test's internal array).

Of those problems, the last seems the most easily solved, if we're okay with having `pushFailure` call something like `assert.ok(false)` internally (or maybe `assert.pushResult`?). Then I could duck-punch the assertion method and use that for testing.

Regarding the question of responsibility-- I don't know too much about the JS Reporters design/architecture/goals, but I'm a bit worried about having them be responsible for error handling. It seems to me that the JS Reporters work primarily by listening to events emitted by the test runner (please interpret that loosely). Do JS Reporters have a way, at this point, of communicating *back* to the test runner? And if not, should they? I'm not 100% sure they should.

I do have plenty of reading to do about that, though. So I could be way off base.

platinumazure:
@gibson042 Just pushed some commits which test this logic outside of an existing test. Had to refactor Test `pushFailure` to call Assert `pushResult` in order to be able to intercept that call and make the test work.

Slightly increased the surface of `pushResult` to allow callers to pass in a precalculated source string (filename:lineno); I hope that's okay since it did result in DRYer code.

Overall I think `QUnit.onError` is pretty well tested now. I still want to maybe revisit the contract (I think a single descriptor similar to Assert pushResult might be more explicit and more extensible long-term), but otherwise this is at least 90% ready.

What do you think?

platinumazure:
@qunitjs/qunit-team Okay, so I've rebased this and it's like 95% ready to go (less API documentation changes, which I'll prepare this week). Only open question here is what the API contract should look like.

This API is primarily about handling errors, so I think it might be nice if we could except an Error object and infer useful information from it right off the bat (e.g., get a filename and line number using the stack trace, use the error message as part of the eventual assert message, etc.). But I'm not sure what other information might also be useful for logging.

If anyone has any thoughts on this, I'd love to hear what you think.

reporter/html.js
```diff
@@ -896,4 +897,27 @@ if ( notPhantom && document.readyState === "complete" ) {
...
+window.onerror = function( ...args ) {
+	var ret = false;
+	if ( originalWindowOnError ) {
+		ret = originalWindowOnError( ...args );
+	}
+
+	// Treat return value as window.onerror itself does,
+	// Only do our handling if not suppressed.
+	if ( ret !== true ) {
+		onError( ...args );
```
gibson042: Shouldn't the return value be assigned to `ret`?

reporter/html.js
```diff
@@ -896,4 +897,27 @@ if ( notPhantom && document.readyState === "complete" ) {
...
+// QUnit.onError function.
+var originalWindowOnError = window.onerror;
+
+// Cover uncaught exceptions
+// Returning true will suppress the default browser handler,
+// returning false will let it run.
+window.onerror = function( ...args ) {
+	var ret = false;
+	if ( originalWindowOnError ) {
+		ret = originalWindowOnError( ...args );
```
gibson042: This should be `ret = originalWindowOnError.apply( this, args )` for strict adherence to [the spec](https://html.spec.whatwg.org/multipage/webappapis.html#onerroreventhandler):
> …and with the [_callback this value_](https://heycam.github.io/webidl/#dfn-callback-this-value) set to _E_'s [`currentTarget`](https://dom.spec.whatwg.org/#dom-event-currenttarget).

It will probably never matter, but environments can be faked and being explicit doesn't hurt.

platinumazure:
@gibson042 Thanks for reviewing once more. I believe I've resolved your concerns (and I even added some tests around the window.onerror logic for HTML Reporter, let me know if you see anything weird there).

My only real concern with following the window.onerror interface is for test runners that don't exist in a browser and might only have an error object available (such as Node.js' `process.on("uncaughtException")`). I think it might be nice to have an API allowing just an error object to be passed in. However, that *may* be something we could do in the future if it's not a breaking change: at the moment, the code implicitly assumes the first argument is an error message string, so we could detect an Error and handle things differently in the future.

If you agree that it might be possible to add an error-centric API later as a non-breaking change, this may be ready for merge. Otherwise, maybe we should spend a little more time on the contract.

trentmwillis:
Will try to do a thorough review tonight, but at a glance seems good.

test/reporter-html/window-onerror-preexisting-handler.js
```diff
@@ -0,0 +1,128 @@
...
+			// Restore assert.pushResult so the next assertions work
+			assert.pushResult = originalPushResult;
+
+			assert.strictEqual( resultInfo.message, "An error message" );
+			assert.strictEqual( resultInfo.source, "filename.js:1" );
+		};
+
+		var result = window.onerror( "An error message", "filename.js", 1 );
+
+        assert.strictEqual( onerrorCallingContext, window, "Handler called with correct context" );
```
platinumazure: Indentation issue, will fix soon.

test/reporter-html/window-onerror-preexisting-handler.js
```diff
@@ -0,0 +1,128 @@
...
```
platinumazure: Indentation issue, will fix soon.

test/reporter-html/window-onerror-preexisting-handler.js
```diff
@@ -0,0 +1,128 @@
...
+
+			assert.ok( false, "Should not push an assertion" );
+		};
+
+		var result = window.onerror( "An error message", "filename.js", 1 );
+
+		// Restore assert.pushResult so the next assertions work
+		assert.pushResult = originalPushResult;
+
+        assert.strictEqual( onerrorCallingContext, window, "Handler called with correct context" );
```
platinumazure: Indentation issue, will fix soon.

test/reporter-html/window-onerror-preexisting-handler.js
```diff
@@ -0,0 +1,128 @@
...
```
platinumazure: Indentation issue, will fix soon.

test/reporter-html/window-onerror-preexisting-handler.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit HTML Reporter - No Markup</title>
+	<link rel="stylesheet" href="../../dist/qunit.css">
+	<script>
+		/* exported onerrorReturnValue, onerrorCallingContext */
```
platinumazure: Might be able to remove "onerrorReturnValue" from exported comment here, will check.

reporter/html.js
```diff
@@ -1,4 +1,5 @@
 import QUnit from "../src/core";
+import onError from "../src/core/onerror";
```
trentmwillis: This file was previously not importing anything from `core` directly (other than the `QUnit` object) to ensure that the HTML Reporter stays loosely-coupled.

So, I'd like to see us just use `QUnit.onError` instead. This way, once we move to a pluggable reporter setup, the HTML Reporter is just another potential option.

src/core/onerror.js
```diff
@@ -1,43 +1,23 @@
...
-			} else {
-				test( "global failure", extend( function() {
-					pushFailure( error, filePath + ":" + linerNr );
-				}, { validTest: true } ) );
-			}
-			return false;
+// Handle an unhandled exception. By convention, returns true if further
+// error handling should be suppressed and false otherwise.
+// In this case, we will only suppress further error handling if the
+// "ignoreGlobalErrors" configuration option is enabled.
```
trentmwillis: For my own education, can you explain what `ignoreGlobalErrors` is for? Seems like an internal thing that I (somehow) have not stumbled across before.

platinumazure: I wish I knew! This configuration option was present in the original `window.onerror` that is currently on master.

From the looks of it, it seems to just suppress QUnit's reporting of any unhandled errors (i.e., errors that make it to window.onerror).

I could see an argument that perhaps this logic should be kept in `window.onerror` and that any call to `QUnit.onError` should report the error no matter what.

reporter/html.js
platinumazure: Makes sense, I'll fix this.

platinumazure:
Regarding window.onerror-like vs single error arg-based approach, I'm honestly okay with either. I'm just hoping to see consensus.

@gibson042 What do you think of @trentmwillis' suggestion in his [latest review](https://github.com/qunitjs/qunit/pull/1099#pullrequestreview-22180625)?

platinumazure:
@trentmwillis HTML Reporter is now using `QUnit.onError`, which allowed me to simplify the associated tests because I could mock QUnit.onError directly. Please let me know what you think.

Remaining open questions:

* Should I change QUnit.onError contract to take in an error-like object (and assert that object is passed in window.onerror tests)?
  * Mostly just needs consensus
* Should I move ignoreGlobalErrors logic to window.onerror handler?
  * Underlying question is: Does QUnit core need that setting or is that really an HTML Reporter setting?
  * ...Can we remove it if it's undocumented? :smiling_imp: 

Any thoughts on these open questions would be greatly appreciated.

trentmwillis:
> Should I change QUnit.onError contract to take in an error-like object (and assert that object is passed in window.onerror tests)?

I would prefer this.

> Should I move ignoreGlobalErrors logic to window.onerror handler?

Nah, doesn't seem specific to `window.onerror`. I'm unsure about removing it, but that could (should?) be handled in a follow-up PR.

platinumazure:
@gibson042 You had indicated a preference for the window.onerror-like contract. Do you object to @trentmwillis' preference? In theory, the same argument about backward-compability being possible in a future contract enhancement would go both ways, if we wanted to add support for the other signature later in QUnit.onError. (That said, I think I like the idea of just working with an error object, myself.)

platinumazure:
@gibson042 Since I haven't heard from you, I'm going to go ahead and implement @trentmwillis' recommendations. If you have a strong objection, please speak up. Thanks!

platinumazure:
@leobalter Hoping to apply @trentmwillis' changes tonight. Any chance this can make it into the minor release? Or do you absolutely need to cut that now?

leobalter:
I believe it would be good to land this as well. Better than two minor releases that are too close to each other.


cc @trentmwillis

trentmwillis:
:+1: sounds good to me; will hold off on doing the release then

platinumazure:
Very sorry for letting this sit. I will definitely get this fixed up tonight. Thanks for your patience!

platinumazure:
@trentmwillis This should be ready for review again. Let me know if I missed something. Thanks!

trentmwillis:
@platinumazure seems good to me, can you squash the fixup commits and merge yourself?

platinumazure:
@trentmwillis Sure, can do. Strong preference between retaining all non-fixup commits, vs having just one overall commit? (I usually aim for the latter as long as the PR was cohesive, i.e., about one feature.)

trentmwillis:
I am impartial, as long as the commit messages are clear and descriptive, so whatever you feel is best!

platinumazure:
Damnit, forgot to include a close tag in the commit messages. Seriously should have just used the squash-and-merge interface here. Anyway, it's merged and pushed with baff17d.

gibson042:
Sorry about the delay; I'm trying to dig out today. Retroactive LGTM.

PR #1105, opened 2017-03-11
Core: Fix start on Node when autostart is not set to true

trentmwillis:
When I submitted https://github.com/qunitjs/qunit/pull/1103, it turns out the test-on-node task never actually ran. The reason is that we don't set `autostart` to `true`. This fix will handle that case as well.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -129,16 +129,18 @@ extend( QUnit, {
 			} else if ( config.autostart ) {
 				throw new Error( "Called start() outside of a test context when " +
 					"QUnit.config.autostart was true" );
-			} else if ( !defined.document && !config.pageLoaded ) {
-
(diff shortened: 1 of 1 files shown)


platinumazure:
Oh, and is the issue fixed by this PR the reason why `grunt` doesn't actually run the node or browser tests? If so, I'd love to get this merged soon!

trentmwillis:
> I assume the goal is to ensure `QUnit.autostart` is set whenever `!config.pageLoaded` (while invoking `QUnit.load` in non-browser contexts)?

Yes, this is because we can assume that someone calling `start` is expecting the tests to start ASAP.

> is the issue fixed by this PR the reason why grunt doesn't actually run the node or browser tests?

Yep. And the Travis job shows them properly running now. If this looks good to you, we can merge.

PR #1106, opened 2017-03-11
[WIP] Core: Introduce reporter control API

trentmwillis:
This introduces an API to control what reporters are used by QUnit at run time.

This should eventually be used to register the HTML Reporter by default for Browser environments and the TAP Reporter by default for non-Browser environments.

### DIFF
--- a/src/reporter.js
+++ b/src/reporter.js
@@ -1,0 +1,37 @@
+import { objectType } from "./core/utilities";
+import QUnit, { internalState } from "./core";
+
+const reporters = [];
+let defaultReporter;
+
+function validateReporterChange( functionName, reporter ) {
+	let reporterType = objectType( reporter );
+	if ( reporterType !== "function" ) {
+		throw new Error(
+			`Expected ${functionName} to be called with a constructor function, ` +
+			`but was called with a ${reporterType}.`
+		);
+	} else if ( internalState.runStarted ) {
+		throw new Error(
+			`Called ${functionName} after the test run has started.`
+		);
+	}
+}
+
+export function addReporter( reporter ) {
+	validateReporterChange( "addReporter", reporter );
+	reporters.push( reporter );
+}
+
+export function setDefaultReporter( reporter ) {
+	validateReporterChange( "setDefaultReporter", reporter );
+	defaultReporter = reporter;
+}
+
+export function initializeReporters() {
+	[ defaultReporter, ...reporters ]
+		.filter( Boolean )
+		.forEach( reporter => {
+			return reporter.init ? reporter.init( QUnit ) : new reporter( QUnit );
+		} );
+}
--- a/test/reporter.js
+++ b/test/reporter.js
@@ -1,0 +1,80 @@
+var defaultReporter1Called = false;
+function DefaultReporter1() {
+	defaultReporter1Called = true;
+}
+
+var defaultReporter2Called = false;
+function DefaultReporter2() {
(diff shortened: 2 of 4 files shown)


test/reporter.js
```diff
@@ -0,0 +1,80 @@
...
+	QUnit.module( "setDefaultReporter", function() {
+		QUnit.test( "only initializes last default reporter", function( assert ) {
+			assert.notOk( defaultReporter1Called, "first default reporter not initialized" );
+			assert.ok( defaultReporter2Called, "last default reporter initialized" );
+		} );
+
+		QUnit.test( "throws an error if the reporter is not a function", function( assert ) {
+			assert.throws(
+				() => QUnit.setDefaultReporter( "TAP" ),
+				// eslint-disable-next-line
```
platinumazure: If possible, could we specify which lint rule we want to disable here? (I'm assuming `max-len`. If multiple rules need to be disabled, you can use a comma (and optionally space) delimited list.)

More info [here](http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments).

test/reporter.js
```diff
@@ -0,0 +1,80 @@
...
+				initReporterCalledWith,
+				QUnit,
+				"Called reporter init method with QUnit"
+			);
+		} );
+
+		QUnit.test( "throws an error if the reporter is not a function", function( assert ) {
+			assert.throws(
+				() => QUnit.addReporter( "TAP" ),
+				// eslint-disable-next-line
```
platinumazure: If possible, could we specify which lint rule we want to disable here? (I'm assuming `max-len`. If multiple rules need to be disabled, you can use a comma (and optionally space) delimited list.)

More info [here](http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments).

platinumazure:
@trentmwillis FYI, I have filed jquery/eslint-config-jquery#12 which configures eslint-config-jquery to not warn for max-len if the line contains a URL or a regex literal (both of which are explicitly listed as exceptions to the max line length rule in the jQuery style guide). So you may be able to remove the config comments if that gets merged and released and we consume the new version of eslint-config-jquery. We could also change our local .eslintrc.json in the meantime, if desired.

src/reporter.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function addReporter( reporter ) {
+	validateReporterChange( "addReporter", reporter );
+	reporters.push( reporter );
+}
+
+export function setDefaultReporter( reporter ) {
+	validateReporterChange( "setDefaultReporter", reporter );
+	defaultReporter = reporter;
```
gibson042: What is the purpose of a default reporter?

src/reporter.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function initializeReporters() {
+	[ defaultReporter, ...reporters ]
+		.filter( Boolean )
+		.forEach( reporter => {
+			return reporter.init ? reporter.init( QUnit ) : new reporter( QUnit );
```
gibson042: Having a single interface would simplify documentation and testing.

trentmwillis:
> any code that can call `QUnit.addReporter( val )` can call e.g. `val.init( QUnit )` on its own.

Good point. The original inspiration was https://github.com/js-reporters/js-reporters/issues/3#issuecomment-245248376, but after further thinking I agree that this does not add value at this point in time. I'm going to close.

PR #1108, opened 2017-03-11
Core: QUnit.onError resumes the test runner

platinumazure:
Fixes gh-990.

At first I was thinking we needed a configuration option to control this behavior, but... who wouldn't want a fatal error to at least try to resume the test runner? (We could add a configuration option later, if necessary.)

If someone needs to account for multiple errors possibly being thrown asynchronously, I think that could be handled by implementing a custom `window.onerror` handler (or equivalent) which would handle collecting the individual errors and only call `QUnit.onError()` with an aggregate error object when it is for sure abandoning the test. (Of course, better practice is to implement error handling in the test itself, or write more reliable code...)

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -663,7 +663,7 @@ export function internalStop( test ) {
 }
 
 // Forcefully release all processing holds.
-function internalRecover( test ) {
+export function internalRecover( test ) {
 	test.semaphore = 0;
 	internalStart( test );
 }
--- a/test/onerror/inside-test.js
+++ b/test/onerror/inside-test.js
@@ -33,4 +33,26 @@ QUnit.module( "QUnit.onError", function() {
 
 		assert.strictEqual( result, true, "onError should not allow other error handlers to run" );
 	} );
+
+	QUnit.test( "Should resume test runner if test was stopped", function( assert ) {
+		assert.expect( 2 );
+
+		// This would hold up the test runner, but onError should resume it
+		assert.async();
+
+		// Avoid assertion error
+		QUnit.config.current.pushFailure = function() {
+			assert.ok( true, "pushFailure was called" );
+		};
+
+		setTimeout( function() {
+			assert.ok( true, "Waited for async successfully" );
+
+			QUnit.onError( {
+				message: "Error message",
+				fileName: "filePath.js",
+				lineNumber: 1
+			} );
+		}, 50 );
+	} );
 } );
--- a/src/core/onerror.js
+++ b/src/core/onerror.js
@@ -1,4 +1,4 @@
-import { pushFailure, test } from "../test";
+import { internalRecover, pushFailure, test } from "../test";
 
 import config from "./config";
 import { extend } from "./utilities";
@@ -12,7 +12,10 @@ export default function onError( error, ...args ) {
 		if ( config.current.ignoreGlobalErrors ) {
 			return true;
 		}
+
 		pushFailure( error.message, error.fileName + ":" + error.lineNumber, ...args );
+
+		internalRecover( config.current );
 	} else {
 		test( "global failure", extend( function() {
 			pushFailure( error.message, error.fileName + ":" + error.lineNumber, ...args );
--- a/test/main/async.js
+++ b/test/main/async.js
@@ -173,6 +173,19 @@ QUnit.test( "fails if callback is called more than callback call count", functio
 	done();
 } );
 
+QUnit.test( "Current test should not advance while waiting for async", function( assert ) {
+	var done = assert.async();
+
+	assert.expect( 2 );
+
+	assert.strictEqual( QUnit.config.current, assert.test, "Current test is this test" );
+
+	setTimeout( function() {
+		assert.strictEqual( QUnit.config.current, assert.test, "Current test is still this test" );
+		done();
+	}, 100 );
+} );
+
 QUnit.module( "assert.async in module hooks", {
 	before: asyncCallback,
 	beforeEach: asyncCallback,


test/onerror/inside-test.js
```diff
@@ -33,4 +33,20 @@ QUnit.module( "QUnit.onError", function() {
...
+	QUnit.test( "Should resume test runner if test was stopped", function( assert ) {
+		assert.expect( 0 );
+
+		// This would hold up the test runner, but onError should resume it
+		assert.async();
+
+		// Avoid assertion error
+		QUnit.config.current.pushFailure = function() {};
+
+		QUnit.onError( {
```
trentmwillis: Can we make this actually async and add an assertion? Something simple like:

```js
setTimeout( function() {
  assert.ok( true, "waited for async" );
  QUnit.onError( .. );
}, 50 );
```

Just to ensure that we actually stopped and then resumed.

src/test.js
```diff
@@ -728,7 +728,7 @@ export function internalStop( test ) {
 }
 
 // Forcefully release all processing holds.
-function internalRecover( test ) {
+export function internalRecover( test ) {
```
trentmwillis: Doesn't have to be in this PR, but we should probably separate out the processing start/stop logic from the test logic.

platinumazure: I'm not sure I understand what you mean here.

Would it be worth creating an issue to explain this in more detail (as well as to help track it so we eventually fix it)?

platinumazure:
@trentmwillis Thanks for the review. I've updated the unit test with your recommendations. I wasn't quite sure what to make of your other comment, so I haven't made any changes there. This may be ready for review again.

test/onerror/inside-test.js
```diff
@@ -33,4 +33,24 @@ QUnit.module( "QUnit.onError", function() {
...
 	} );
+
+	QUnit.test( "Should resume test runner if test was stopped", function( assert ) {
+		assert.expect( 0 );
+
+		// This would hold up the test runner, but onError should resume it
+		assert.async();
+
+		// Avoid assertion error
+		QUnit.config.current.pushFailure = function() {};
```
gibson042: There's something screwy going on with this module in the browser, which unfortunately has been hidden because we forgot to add `onerror/inside-test.js` and `onerror/outside-test.js` to test/index.html. There may be cross-test interaction—commenting out the `onError` in _this_ test doesn't hang the runner unless I also comment out the preceding "Should call pushFailure when inside a test" test. And I think this replacement `pushFailure` should make some assertions of its own.

platinumazure:
Ack, sorry. I'll investigate and write a separate PR today.

platinumazure:
I've been digging for most of the day and I'm seeing some examples of things being wrong (especially with the main changes in this PR), but I haven't been able to find a root cause. At this point, I am thinking that one of the onError tests does have side-effects outside the module, but also that some of the async tests might as well.

@gibson042 It would really help if you could clarify your comment for me. You point to a line in this PR as problematic but you also say that the test suite is broken since #1099 was merged (unless I'm misunderstanding something). I can't see any breakages post-#1099 without these changes. What am I missing? I'd rather fix anything that might be wrong with #1099 before trying to fix anything wrong with this PR, so that the test suite can be in a good state again (if it isn't right now). I know your time is precious and I appreciate in advance any clarification you can offer. Thanks.

gibson042:
@platinumazure  I don't think I know much more than you do. I can see that the test suite is malfunctioning because when I fetch this PR and comment out [the critical part of the new test](https://github.com/platinumazure/qunit/blob/d71249c308118c6d3db93ad1d92e2d9889683101/test/onerror/inside-test.js#L49), my browser still reports success for test/onerror/inside-test.html (i.e., loaded directly). BUT when I disable [an earlier test in the module](https://github.com/platinumazure/qunit/blob/d71249c308118c6d3db93ad1d92e2d9889683101/test/onerror/inside-test.js#L2), the run hangs as expected.

Upon closer inspection, I _think_ the issue stems from this PR's code. Initial omission of `onError` tests from the browser suite was a mistake, but I think it was a harmless one. And it may be that the only changes needed here are separation of erroring tests from verifying tests (like we do for callback data).

I wish I could offer more right now, but I'm happy to explore further if you run into roadblocks (though it will unfortunately take the form of more back-and-forth like this).

platinumazure:
Okay, excellent, I've rebased now that #1111 is merged and now Travis is failing, matching @gibson042's experience. I'll keep digging.

platinumazure:
Been digging more and I think I understand the basic issue better. In the async test, there's an implicit expectation that because the test used `assert.async`, by the time the callback is executed `QUnit.config.current === assert.test` should continue to hold. It may or may not hold depending on the timeout delay: Anything lower than 10ms tends to work, anything above 20ms tends not to work, with some inconsistent results between those values. This suggests a race condition of some kind.

I don't think it is ever correct that the current test should advance while waiting for async to resolve, so I will continue to investigate along those lines. Either there is a bug in core in this area, or my changes and tests are causing this. I may try creating a test where `QUnit.onError` isn't used, but rather an ordinary async test with resolution in a setTimeout callback. That test will also assert that `QUnit.config.current === assert.test`. If that is false, then I will assume there is an existing bug in core.

platinumazure:
Proved that there's no issue in core with regard to the test runner advancing prematurely while waiting for async (in a normal async test). So yeah, it's definitely something wrong with either `QUnit.onError` or the tests. Apologies to @gibson042 for taking so long to prove him right. :laughing:

I think I probably need to study other error scenarios and see if I can edit my tests to follow a similar approach. I'm fairly optimistic that the code changes in this PR (outside of tests) are basically fine-- they're pretty simple after all. But we'll see.

Something else I thought of: This approach does not call `saveGlobal()` before going to the next test, so later tests will get blamed for new globals.
1. Should I handle that in this PR, or is it okay if I do it separately? (See below for why I might want to do it separately)
1. I was thinking it might be a good idea to refactor `internalRecover()` to call `saveGlobal()`, since I can't come up with a scenario where we would want to resume the test runner (i.e., abandon the current test and advance to the next one) without also resetting the global variable tracker. And I don't see much point in having `saveGlobal()` be exported and then called from `src/core/onerror.js`.

Thanks for everyone's patience as I continue to try to work through this!

trentmwillis:
> Should I handle that in this PR, or is it okay if I do it separately?

I think it is fine to do it separately, as long as it doesn't leave `master` in a broken state in the interim. At the least, they should likely be separate commits.

> I was thinking it might be a good idea to refactor `internalRecover()` to call `saveGlobal()`

I'm unsure about this. I feel like an abandoned test would actually want to reset the global by removing any new globals that maybe didn't get cleaned up.

trentmwillis:
@platinumazure where did we land with this? Is it still worth pursuing?

platinumazure:
I want to, yes, but I haven't had time. Feel free to close.

trentmwillis:
No worries; we can keep it open. I just wanted to understand if it is still valid as a feature or not.

platinumazure:
I'm going to close this since I don't know when I'll make progress here. Hopefully if I take a fresh crack, I can have better results than I did here.

PR #1109, opened 2017-03-11
All: Enable ESLint prefer-const rule

platinumazure:
Flags `let` variables that don't ever get reassigned and could be turned into `const` variables. Does not flag `var` variables.

See [here](http://eslint.org/docs/rules/prefer-const) for more information.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,6 +16,7 @@
         "no-this-before-super": "error",
         "no-useless-constructor": "error",
         "no-useless-rename": "error",
+        "prefer-const": "error",
         "rest-spread-spacing": ["error", "never"],
         "wrap-iife": ["error", "any"]
     }
--- a/src/test.js
+++ b/src/test.js
@@ -403,7 +403,7 @@ Test.prototype = {
 	logAssertion( details ) {
 		runLoggingCallbacks( "log", details );
 
-		let assertion = {
+		const assertion = {
 			passed: details.result,
 			actual: details.actual,
 			expected: details.expected,
@@ -647,7 +647,7 @@ export function test( testName, callback ) {
 		return;
 	}
 
-	let newTest = new Test( {
+	const newTest = new Test( {
 		testName: testName,
 		callback: callback
 	} );
@@ -660,7 +660,7 @@ export function todo( testName, callback ) {
 		return;
 	}
 
-	let newTest = new Test( {
+	const newTest = new Test( {
 		testName,
 		callback,
 		todo: true
@@ -675,7 +675,7 @@ export function skip( testName ) {
 		return;
 	}
 
-	let test = new Test( {
+	const test = new Test( {
 		testName: testName,
 		skip: true
 	} );
@@ -692,7 +692,7 @@ export function only( testName, callback ) {
 	config.queue.length = 0;
 	focused = true;
 
-	let newTest = new Test( {
+	const newTest = new Test( {
 		testName: testName,
 		callback: callback
 	} );
@@ -785,12 +785,12 @@ function internalStart( test ) {
 }
 
 function numberOfTests( module ) {
-	let count = module.tests.length,
-		modules = [ ...module.childModules ];
+	let count = module.tests.length;
+	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
-		let nextModule =  modules.shift();
+		const nextModule =  modules.shift();
 		count += nextModule.tests.length;
 		modules.push( ...nextModule.childModules );
 	}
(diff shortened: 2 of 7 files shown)


jzaefferer:
As far as I know, the main argument against using `const` is that parameters can't be declared as `const`. After using this rule in two projects for 6 months+, I don't think that is enough of a concern.

In other words: +1 to using `const` where ever possible.

platinumazure:
@gibson042 @leobalter Any thoughts on implementing this?

.eslintrc.json
```diff
@@ -16,6 +16,7 @@
         "no-this-before-super": "error",
         "no-useless-constructor": "error",
         "no-useless-rename": "error",
+	"prefer-const": "error",
```
Krinkle: Unexpected tab. This file uses spaces for indentation.

Krinkle: Unexpected tab. This file uses spaces for indentation.

leobalter:
I don't like using const unless I have some real intentional desire to protect the value stored when it's initialized. `const PI` would be a good thing, while `const name` is meh. Protecting a name is one extra step to require from the runtime.

`let something` is great as we can not have it twice in the same scope, so that prevents what I really want to be prevented, reassigning a variable with the same name. `var` does not protect us from that.

If you want to use `const` for everything that is not intentionally reassigned, I'm +0 to it (not -1 neither +1).

> parameters can't be declared as const

We can define a lint rule to disallow reassignment to parameters names.

src/test.js
```diff
@@ -785,12 +785,12 @@ function internalStart( test ) {
...
+	let count = module.tests.length;
+	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
-		let nextModule =  modules.shift();
+		const nextModule =  modules.shift();
 		count += nextModule.tests.length;
 		modules.push( ...nextModule.childModules );
 	}
```
leobalter: I think I already tried before, but reduce would be nice here.

Krinkle:
@leobalter I see merit in what you're saying. I agree it seems like making all variables that aren't re-assigned defined with `const` will likely lead more to confusion and maintenance overhead then any tangible benefit. They'd just have to be turned back into `let` when changing the code even a little bit. Especially since the vast majority of cases here have no semantic or other implicit intent that suggests they aren't supposed to be re-defined or that it is somehow unexpected from a developer perspective. it's just the current state as it happens to be.

platinumazure:
In ESLint, we've had great success with `prefer-const` and very many of our declarations are `const`, but I think a lot of that also has to do with the code being a lot more modular. (QUnit's migration to ES6 modules was a *huge* step in the right direction on that front, but there are still some necessarily large modules such as `src/test.js`.) So I can see why `prefer-const` might not be as successful here as it is in ESLint.

I guess my question to the @qunitjs/qunit-team is this: Is it worth trying out the rule for a few weeks, and we could then turn it off if it turns out to be too much of a pain? Or should we just not worry about it? (I personally just use `const` as much as I can anyway, so it wouldn't really affect my workflow.)

trentmwillis:
@platinumazure I'm good with landing this. I think anything that encourages more modular code with our new setup is a step in the right direction. I personally use `prefer-const` in many of my personal projects, so I'm good with it. Though I understand the other arguments about different signals being communicated from other approaches.

All-in-all, I am :+1: and really just want consistency. That said, make sure to address the one white space issue.

leobalter:
I'm not opposing to these proposed changes. While don't think it's necessary to enforce the use of const, I believe it's good to enforce consistency. I'm +1 in the perspective this will be a defined rule through our project. We can always review our coding style when necessary, as we are doing right now.

platinumazure:
Awesome. I'll fix the whitespace issue tonight and then merge.

Thanks,
Kevin

src/test.js
platinumazure: @leobalter Okay if we make a separate issue and/or PR for this? Doesn't seem related to what I'm doing in this PR.

platinumazure:
@Krinkle Mind reviewing once more? I've rebased and fixed the whitespace issue.

src/test.js
leobalter: of course! It's indeed not related (I'm sorry for the noise), and I'm always in favor of follow up PRs rather than blocking anything for too long.

leobalter:
The changes are good to land and I'm not happy with this merge block. :)

+1 to remove this block setting and merge this PR.

platinumazure:
@leobalter I can always dismiss @Krinkle's review, it just seemed rude. :smile: Hopefully he won't mind since I did address his request.

Krinkle:
@platinumazure Yeah, no worries. Thanks :)

PR #1112, opened 2017-03-13
[BUGFIX] fix sessionStorage feature detection

stefanpenner:
Prior to this PR `const sessionStorage` shadowed the global `sessionStorage`.
This resulted in the feature sniffing function to always fail, as `sessionStorage` would always be `undefined`


trentmwillis:
Thanks for this. I'm putting together a failing test so we catch future regressions.

trentmwillis:
Opened another PR since I couldn't modify history on this PR: https://github.com/qunitjs/qunit/pull/1113. That PR has a test and proper commit message. Thanks for the initial work!

PR #1113, opened 2017-03-13
Core: Fix sessionStorage feature detection

trentmwillis:
Continuing https://github.com/qunitjs/qunit/pull/1112. GitHub wouldn't allow me to push to that PR since I modified the history.

> Prior to this PR `const sessionStorage` shadowed the global `sessionStorage`.
> This resulted in the feature sniffing function to always fail, as `sessionStorage` would always be `undefined`

cc @stefanpenner

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -113,6 +113,7 @@ grunt.initConfig( {
 			"test/index.html",
 			"test/autostart.html",
 			"test/startError.html",
+			"test/reorder.html",
 			"test/reorderError1.html",
 			"test/reorderError2.html",
 			"test/callbacks.html",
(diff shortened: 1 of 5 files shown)


platinumazure:
@trentmwillis `git push -f` doesn't work?

trentmwillis:
@platinumazure I guess for the "Allow edits from maintainers" feature it doesn't work. I tried several things, but GitHub kicked it back each time 🙁

platinumazure:
Yeah, maybe they only allow fast-forwards in that case, which is probably sensible.

@trentmwillis I'm not going to merge this without another set of eyes, but if you are confident in the fix, you can merge if you like.

stefanpenner:
👍

trentmwillis:
I'm confident in this fix, particularly because it has tests to verify the behavior. Going to merge...

Issue #1119, opened 2017-03-22
Recursive error in Test#pushFailure and Assert#pushResult


ryanto:
## Tell us about your runtime:

* **QUnit version**: 2.2.1
* **What environment are you running QUnit in? (e.g., browser, Node)**: Browser, Chrome 56
* **How are you running QUnit? (e.g., script, testem, Grunt)**: ember-cli, testem

## What are you trying to do?

Code that reproduces the problem:

Calling this code during an async test after catching an exception.

```js
QUnit.config.current.assert.ok(false);
```

If you have any relevant configuration information, please include that here:

## What did you expect to happen?

The test to fail...

## What actually happened?

Recursive error with call stack exceeded.

`Test#pushFailure` calls `Assert#pushResult`
https://github.com/qunitjs/qunit/blob/master/src/test.js#L388

`Assert#pushResult` calls `Test#pushFailure`
https://github.com/qunitjs/qunit/blob/master/src/assert.js#L107

trentmwillis:
Thanks for the report and investigation! This is a bug, I believe all that we need to do to fix is update [`Test#pushFailure`](https://github.com/qunitjs/qunit/blob/7e5d0556acaff7df2145c394840293f0c3ac8302/src/test.js#L388) to call `Test#pushResult` instead of `Assert#pushResult`.

ryanto:
Cool. I can take a shot at this! I'll put this on tomorrow's TODO list.

trentmwillis:
Fixed by #1120.

PR #1120, opened 2017-03-23
Core: Test#pushFailure now calls Test#pushResult

ryanto:
Fix endless recursive loop between Test#pushFailure and
Assert#pushResult.

See #1119 for details

Let me know if anything more is needed for this PR.


trentmwillis:
I went ahead and tested this locally. Seems to have fixed the reported issue. Thanks @ryanto!

PR #1121, opened 2017-03-24
Core: Introduce ProcessingQueue in internals

trentmwillis:
Currently, the logic for processing the queue of tests is scattered. This pulls the two core pieces into a new construct `ProcessingQueue` with an interface that more clearly communicates what `process` and `synchronize` used to do.

In other words:
- `process()` -> `ProcessingQueue.advance()`: Moves the queue forward
- `synchronize()` -> `ProcessingQueue.add()`: Adds a function to the queue

This is being done preemptively to support the "watch" feature in the forthcoming CLI. We'll need to refactor a portion of how the queue is processed in order to properly abort test runs when watched files change.

### DIFF
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -1,0 +1,155 @@
+import config from "./config";
+import {
+	defined,
+	generateHash,
+	now,
+	objectType
+} from "./utilities";
+import {
+	runLoggingCallbacks
+} from "./logging";
+
+import {
+	internalState,
+	globalSuite
+} from "../core";
+import {
+	emit
+} from "../events";
+import {
+	setTimeout
+} from "../globals";
+
+let priorityCount = 0;
+let unitSampler;
+
+/**
+ * Advances the ProcessingQueue to the next item if it is ready.
+ * @param {Boolean} last
+ */
+function advance( last ) {
+	function next() {
+		advance( last );
+	}
+
+	const start = now();
+	config.depth = ( config.depth || 0 ) + 1;
+
+	while ( config.queue.length && !config.blocking ) {
+		const elapsedTime = now() - start;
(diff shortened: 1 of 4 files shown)


src/core/processing-queue.js
```diff
@@ -0,0 +1,155 @@
+import config from "./config";
+import {
+  defined,
+  generateHash,
+  now,
+  objectType
+} from "./utilities";
```
platinumazure: Inconsistent indentation here.

trentmwillis: Good catch! We should probably turn on the eslint rule for indentation...and I should probably update my editor workspace to use tabs in QUnit.

trentmwillis:
@platinumazure updated, thanks for the review!

src/core/processing-queue.js
platinumazure: Re: ESLint rule for indentation: One bit of the jQuery style guide, which the ESLint indent rule doesn't yet support, stymies this.

Specifically: Things that look like a top-level whole-file IIFE or function wrapper, but aren't pure IIFEs, are supposed to have an indent level of zero for contained statements. For example, AMD `define` callbacks, or assignments to `module.exports` that are functions, or IIFEs that follow the flexible factory pattern (the factory selector/"intro" code needs one indent level, the actual code needs zero). All of those are use cases not yet supported by ESLint.

That said, we could choose to activate the rule anyway and just accept that for this project, all function bodies will be indented by one tab. I don't think that's the end of the world to be completely honest. And both when we activate the rule and if we consume an upgraded version of the rule which supports our special function indentation preferences, we can use `eslint --fix` to fix the existing problems fairly easily.

I'll open an issue re: indent rule.

PR #1123, opened 2017-03-24
CLI: Add file watching option

trentmwillis:
This PR introduces a `--watch` option to the CLI so that tests will re-run when a file is added, changed, or removed from the working directory. It also ensures that if one of the above occurs while tests are executing, then the current test finishes and does any cleanup (via `after`/`afterEach` hooks) before restarting.

This depends on https://github.com/qunitjs/qunit/pull/1121 and https://github.com/qunitjs/qunit/pull/1124. I recommend not reviewing until both of those PRs are merged, because this PR will be much simpler afterward.

Todo:
- [x] Figure out why tests are hanging in Travis

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -69,14 +69,31 @@ function run( args, options ) {
 };
 
 run.restart = function( args ) {
-	run.abort();
-	run.apply( null, args );
+	if ( QUnit.config.queue.length ) {
+		console.log( "Finishing current test and restarting..." );
+	} else {
+		console.log( "Restarting..." );
+	}
+
+	run.abort( () => run.apply( null, args ) );
 };
 
-run.abort = function() {
-	QUnit.config.queue.length = 0;
-	delete global.QUnit;
-	QUnit = null;
+run.abort = function( callback ) {
+	function clearQUnit() {
+		delete global.QUnit;
+		QUnit = null;
+		if ( callback ) {
+			callback();
+		}
+	}
+
+	if ( QUnit.config.queue.length ) {
+		const nextTestIndex = QUnit.config.queue.findIndex( fn => fn.name === "runTest" );
+		QUnit.config.queue.splice( nextTestIndex );
+		QUnit.on( "runEnd", clearQUnit );
+	} else {
+		clearQUnit();
+	}
 };
 
 function watcherEvent( event, args ) {
@@ -102,9 +119,11 @@ run.watch = function watch() {
 
 	function stop() {
 		console.log( "Stopping QUnit..." );
-		run.abort();
+
 		watcher.close();
-		process.exit();
+		run.abort( () => {
+			process.exit();
+		} );
 	}
 
 	process.on( "SIGTERM", stop );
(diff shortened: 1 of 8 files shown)


bin/run.js
```diff
@@ -50,6 +62,74 @@ module.exports = function run( files, options ) {
...
+function watcherEvent( event, args ) {
+	return ( file ) => {
+		console.log( `File ${event}: ${file}` );
+		run.restart( args );
+	};
+}
+
+run.watch = function watch() {
+	const chokidar = require( "chokidar" );
+	const args = Array.prototype.slice.call( arguments );
```
platinumazure: We could use a rest parameter instead of slicing the `arguments` object here.

bin/run.js
```diff
@@ -50,6 +62,74 @@ module.exports = function run( files, options ) {
...
+};
+
+run.restart = function( args ) {
+	if ( QUnit.config.queue.length ) {
+		console.log( "Finishing current test and restarting..." );
+	} else {
+		console.log( "Restarting..." );
+	}
+
+	run.abort( () => run.apply( null, args ) );
```
platinumazure: Could use the spread operator here.

bin/run.js
```diff
@@ -50,6 +62,74 @@ module.exports = function run( files, options ) {
...
+		QUnit.config.queue.splice( nextTestIndex );
+		QUnit.on( "runEnd", function() {
+			delete global.QUnit;
+			QUnit = null;
+			if ( callback ) {
+				callback();
+			}
+		} );
+	} else {
+		delete global.QUnit;
```
platinumazure: Any point in extracting this logic to a closure, so that this code isn't duplicated with the above case?

test/cli/watch.js
```diff
@@ -0,0 +1,248 @@
...
+}
+
+const fixturePath = path.join( __dirname, "fixtures", "watching" );
+
+// Kills the provided executing process, handling differences in platforms
+function kill( execution, signal ) {
+	const sig = signal || "SIGINT";
+
+	// Linux increments the pid by 1 due to creating a new shell instance
+	if ( process.platform === "linux" ) {
```
platinumazure: What about OSX or MacOS? Can this be detected any better? Can we do a kill-with-children sort of thing on the process so we don't even need to worry about whether a grandchild process was even spawned?

platinumazure:
LGTM, but left some nitpicks just in case you want to address them.

bin/run.js
trentmwillis: Spread/rest are not supported in Node 4 unfortunately 😞 https://kangax.github.io/compat-table/es6/#test-rest_parameters

bin/run.js
trentmwillis: Good suggestion, I meant to do that but forgot.

test/cli/watch.js
trentmwillis: We could add more logic and special case per-platform, but I'd rather keep this simple to start and evolve it as needed over time.

The kill-with-child thing isn't really applicable, because we don't want to kill the parent process (which is the test suite running).

trentmwillis:
Thanks for the review!

PR #1126, opened 2017-03-24
CLI: Add support for custom reporters

trentmwillis:
This introduces a `--reporter` option to the CLI as discussed in https://github.com/qunitjs/qunit/issues/1024#issuecomment-287141622.

It accepts a string argument, which it will try to find either as a part of the `js-reporters` package or npm package. If it does not find it in either of those places, then it will log all the available reporters and exit the process.

Custom reporters can be introduced via npm packages that have `js-reporter` as a keyword in their `package.json`.

### DIFF
--- a/bin/find-reporter.js
+++ b/bin/find-reporter.js
@@ -1,0 +1,82 @@
+"use strict";
+
+const JSReporters = require( "js-reporters" );
+const path = require( "path" );
+const findup = require( "findup-sync" );
+const utils = require( "./utils" );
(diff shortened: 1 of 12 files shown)


trentmwillis:
I've rebased this and it is now ready for review. Most of the PR is fixtures and such for the tests.

trentmwillis:
Updated again.

leobalter:
screenshot 2017-03-28 17 21 14

GitHub is trolling me.

PR #1128, opened 2017-03-25
CLI: Add description to help command

trentmwillis:
This provides some helpful, descriptive text when using `qunit --help`:

```bash
$ qunit --help

  Usage: qunit [options] [files]

  Runs tests using the QUnit framework.

  Files should be a space-separated list of file/directory paths and/or glob
  expressions. Defaults to 'test/**/*.js'.

  For more info on working with QUnit, checkout http://qunitjs.com.
```

Meant to have this in the original PR, but forgot.

### DIFF
--- a/bin/qunit
+++ b/bin/qunit
@@ -7,10 +7,18 @@ const run = require( "./run" );
 const utils = require( "./utils" );
 const pkg = require( "../package.json" );
 
+const description = `Runs tests using the QUnit framework.
+
+  Files should be a space-separated list of file/directory paths and/or glob
+  expressions. Defaults to 'test/**/*.js'.
+
+  For more info on working with QUnit, check out http://qunitjs.com.`;
+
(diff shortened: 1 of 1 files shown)


bin/qunit
```diff
@@ -7,10 +7,18 @@ const run = require( "./run" );
 const utils = require( "./utils" );
 const pkg = require( "../package.json" );
 
+const description = `Runs tests using the QUnit framework.
+
+  Files should be a space-separated list of file/directory paths and/or glob
+  expressions. Defaults to 'test/**/*.js'.
+
+  For more info on working with QUnit, checkout http://qunitjs.com.`;
```
platinumazure: Just one nitpick: s/checkout/check out

Otherwise LGTM.

trentmwillis: Good catch! Updated.

PR #1130, opened 2017-03-28
HTML Reporter: Display todo tests on hidepassed

### LINKED ISSUE #1125: Pending QUnit.todo tests should be listed with "hide passed" checked.
I started experimenting with the new (to me at least) `QUnit.todo` feature that exists in qunitjs@2.2.1, and I really like it! Thank you all for working so hard to implement it.

I noticed a few things that stuck out to me while using todo's, and I would like to submit a PR to fix but would like guidance on the best approach before beginning.

In the default HTML reporter tests ran as `QUnit.todo`'s get a visual indication to the left of the test, saying `[todo]`:

todo screenshot

The mental model that I have when using `QUnit.todo` is that I am quickly putting down tests (often TDD where the test is written before the implementation even exists), and then once I have the tests written I am focusing on making them pass.  In a large test suite, these `[todo]` entries are difficult to find when skimming the list. Unlike skipped tests (which are also technically not failures) which are still displayed when "Hide Passed" is checked, these todo tests do not show up when the "Hide Passed" checkbox is checked.

**tldr;** I would like to have `QUnit.todo` tests not be hidden by "Hide Passed", and instead display them (similar to skipped tests) with an alternate color or other visual treatment in the test list.

trentmwillis:
Closes #1125

### DIFF
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -236,7 +236,7 @@
 }
 
 #qunit-tests.hidepass li.running,
-#qunit-tests.hidepass li.pass {
+#qunit-tests.hidepass li.pass:not(.todo) {
 	visibility: hidden;
 	position: absolute;
 	width:   0;
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -887,6 +887,7 @@ QUnit.testDone( function( details ) {
 			const todoLabel = document.createElement( "em" );
 			todoLabel.className = "qunit-todo-label";
 			todoLabel.innerHTML = "todo";
+			testItem.className += " todo";
 			testItem.insertBefore( todoLabel, testTitle );
 		}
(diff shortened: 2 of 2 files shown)


src/qunit.css
```diff
@@ -236,7 +236,7 @@
 }
 
 #qunit-tests.hidepass li.running,
-#qunit-tests.hidepass li.pass {
+#qunit-tests.hidepass li.pass:not(.todo) {
```
leobalter: it works on IE9!

Krinkle:
This somewhat relates to https://github.com/qunitjs/qunit/issues/1208 where we choose not to hide skipped tests. No problem either way, I'm just noting it here for future reference.

PR #1136, opened 2017-03-31
Docs: Remove baseurl as github is inserting an automatic value

leobalter:


leobalter:
I'm sorry for jumping the trigger here. This fixes several references to load styles and links.

PR #1138, opened 2017-03-31
Core: Fix test instance memory leak

### LINKED ISSUE #1137: Memory leak caused by retaining test instances

* **QUnit version**: 2.3.0
* **What environment are you running QUnit in? (e.g., browser, Node)**: Browser
* **How are you running QUnit? (e.g., script, testem, Grunt)**: Ember-CLI

The `testEnvironment` should not be retained after a test run completes. If consumers set properties on the `testEnvironment` and it is retained, then the test suite will have memory leaks.

## What actually happened?

The `testEnvironment` is getting retained due to a reference from the new js-reporters implementation. Specifically, [this line is problematic](https://github.com/qunitjs/qunit/blob/14e2bb927c0833ff2d727da097bede4294998817/src/reports/test.js#L14).

screen shot 2017-03-31 at 1 31 05 pm

trentmwillis:
Fixes #1137.

### DIFF
--- a/src/reports/suite.js
+++ b/src/reports/suite.js
@@ -57,7 +57,7 @@ export default class SuiteReport {
 
 	getTestCounts( counts = { passed: 0, failed: 0, skipped: 0, todo: 0, total: 0 } ) {
 		counts = this.tests.reduce( ( counts, test ) => {
-			if ( test.isValid() ) {
+			if ( test.valid ) {
 				counts[ test.getStatus() ]++;
 				counts.total++;
 			}
--- a/src/reports/test.js
+++ b/src/reports/test.js
@@ -11,16 +11,12 @@ export default class TestReport {
 		this.skipped = !!options.skip;
 		this.todo = !!options.todo;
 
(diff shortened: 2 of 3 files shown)


trentmwillis:
Unfortunately, I don't believe there is any sort of automated check we could perform. In general, checking for a memory leak in JS requires having a reference to the potential leak, but if you have a reference then it won't be garbage collected, thus making it impossible to tell if it would have leaked or not.

platinumazure:
@trentmwillis Is that so even with `WeakSet` or `WeakMap` available?

trentmwillis:
@platinumazure yes, since there is no way to enumerate the items in a `WeakSet` or `WeakMap`, you have to hold onto a reference of the thing you'd like to check, in which case it won't get garbage collected.

Theoretically, if a `WeakMap` was implemented such that the keys and values were weakly held independent of each other, then you could use the V8 flag to allow you to force `gc()` and then check using a known `key` if the value is still defined. But last time I tried this, it did not work.

PR #1141, opened 2017-04-04
Core: Slim assertions after reporting them

trentmwillis:
After assertions have been logged via the `assertion` and `testEnd` events, we should remove the `actual` and `expected` values from them in order to avoid memory leaks.

For instance, if you do `assert.strictEqual(foo, bar);` then both `foo` and `bar` will be retained forever unless we release `actual` and `expected` from memory at some point. I think that information is useful during reporting for the individual test, but after that point it is non-essential.


src/reports/test.js
```diff
@@ -75,4 +75,14 @@ export default class TestReport {
 	getAssertions() {
 		return this.assertions.slice();
 	}
+
+	// Remove actual and expected values from assertions. This is to prevent
+	// leaking memory throughout a test suite.
+	slimAssertions() {
+		this.assertions = this.assertions.map( assertion => {
```
platinumazure: I think using `this.assertions.forEach` makes more sense, since we're not mutating either the referenced array of assertions nor any object references within the array.

I'd love to see a test as well, if at all possible (maybe check the assertion reports at end of module?).

src/test.js
```diff
@@ -258,6 +258,7 @@ Test.prototype = {
 		}
 
 		emit( "testEnd", this.testReport.end( true ) );
+		this.testReport.slimAssertions();
```
platinumazure: Seems a bit weird to me that the `testEnd` event gets the full assertions, but the `testDone` logging callback does not. Can you explain your reasoning here?

trentmwillis: They're unrelated. The assertion objects used by the new js-reporters events are not the same as the ones used for the old callbacks.

platinumazure: A comment to that effect (or at least a line break between the js-reporters-related and the old callback code) would make that much more clear. :smile:

src/reports/test.js
trentmwillis: I personally prefer re-assignment even if it is doing the same thing as it makes it explicit that mutation occurs during the loop.

As for a test, sounds like a good idea. I'll look into it.

trentmwillis:
@platinumazure updated with revamped/more tests

PR #1142, opened 2017-04-05
All: Enable ESLint indent rule (zero-tab indent for outer IIFE bodies)

platinumazure:
This enables the ESLint "indent" rule.

There are some limitations of this rule which would force us to deviate from the jQuery style guide, especially around full file closures (whether IIFEs or factory functions). This PR chooses to have pure IIFEs be indented 0 tabs, which is closer to the style guide but results in inconsistencies between pure IIFEs and factory functions.

PR #1143 is the one-tab IIFE version.


platinumazure:
Closing since we seem to favor #1143.

PR #1143, opened 2017-04-05
All: Enable ESLint indent rule (one-tab indent for outer IIFE bodies)

platinumazure:
This enables the ESLint "indent" rule.

There are some limitations of this rule which would force us to deviate from the jQuery style guide, especially around full file closures (whether IIFEs or factory functions). This PR chooses to have pure IIFEs be indented 1 tab, which is further from the style guide but results in consistent indentation between pure IIFEs and factory functions.

PR #1142 is the zero-tab IIFE version.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,6 +10,14 @@
         "arrow-spacing": ["error", { "before": true, "after": true }],
         "brace-style": ["error", "1tbs", { "allowSingleLine": true }],
         "constructor-super": "error",
+        "indent": ["error", "tab", {
+                "CallExpression": {
+                        "arguments": 1
+                },
+		"MemberExpression": 1,
+		"SwitchCase": 0,
+		"outerIIFEBody": 1
+        }],
         "no-const-assign": "error",
         "no-dupe-class-members": "error",
         "no-duplicate-imports": "error",
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,236 +9,245 @@ var reportDir = "build/report";
 module.exports = function( grunt ) {
 
 // Load grunt tasks from NPM packages
-require( "load-grunt-tasks" )( grunt );
+	require( "load-grunt-tasks" )( grunt );
 
-function process( code ) {
-	return code
+	function process( code ) {
+		return code
 
 		// Embed version
-		.replace( /@VERSION/g, grunt.config( "pkg" ).version )
+			.replace( /@VERSION/g, grunt.config( "pkg" ).version )
 
 		// Embed date (yyyy-mm-ddThh:mmZ)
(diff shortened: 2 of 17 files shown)


.eslintrc.json
```diff
@@ -10,6 +10,14 @@
         "arrow-spacing": ["error", { "before": true, "after": true }],
         "brace-style": ["error", "1tbs", { "allowSingleLine": true }],
         "constructor-super": "error",
+        "indent": ["error", "tab", {
```
trentmwillis: In what ways does this deviate from the "normal" indent rule? Or, why do we need the complicated object config?

platinumazure: CallExpression indenting isn't on by default. Setting it as I have done forces arguments to function calls on subsequent lines to be indented one tab.

```js
QUnit.on(
    "event", // <-- poor man's tab
    doSomething);
```

MemberExpression does a similar thing with property accesses and would also be off by default:

```js
return myString
    .replace("foo", "bar")
    .replace("baz", "quux");
```

SwitchCase enforces this style of case statements (it's set same as default, but since it was called out in the style guide I thought it was good to include in case the default changes upstream):

```js
switch (x) {
case "foo":
    return true;
case "bar":
    return false;
```

And outerIIFEBody is for full file closures... at least, those that are actually pure IIFEs. Setting it to 1 means basically all function declarations/expressions/IIFEs require 1 tab indent of statements in the body.

.eslintrc.json
```diff
@@ -10,6 +10,14 @@
...
+                "CallExpression": {
+                        "arguments": 1
+                },
+		"MemberExpression": 1,
+                "SwitchCase": 0,
+                "outerIIFEBody": 1
```
platinumazure: Blast, these two lines need to be tabified.

.eslintrc.json
trentmwillis: Great explanation! Thank you, seems good to me then!

.eslintrc.json
trentmwillis: 😆 the irony

trentmwillis:
@platinumazure after a rebase and fixing the whitespace issue called out above, I'm :+1: to land this unless anyone else from @qunitjs/qunit-team disagrees with this approach.

platinumazure:
@trentmwillis Issues fixed and branch is rebased, please review at your convenience.

One thing I'll note: ESLint 3.x's indent rule (i.e., what I've configured) doesn't auto-fix comments correctly so they might not be aligned. When ESLint 4.0 comes out, the indent rule is rewritten to be a lot smarter and will handle comments correctly. I mention this because, if you're okay with the comments being a little off for now, the 4.0 indent rule will auto-fix them in one go. Or, if getting the comments correctly aligned is super important, feel free to request changes when reviewing. Thanks!

platinumazure:
Thanks for merging!

PR #1146, opened 2017-04-06
Assert: Throw if async callback invoked after test finishes

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1019.

### DIFF
--- a/test/main/async.js
+++ b/test/main/async.js
@@ -173,6 +173,20 @@ QUnit.test( "fails if callback is called more than callback call count", functio
 	done();
 } );
 
+( function() {
+	var previousTestDone;
+
+	QUnit.test( "errors if callback is called after test - part 1", function( assert ) {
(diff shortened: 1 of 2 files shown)


test/main/async.js
```diff
@@ -173,6 +175,17 @@ QUnit.test( "fails if callback is called more than callback call count", functio
...
+	assert.expect( 0 );
+	globalDone = assert.async();
+	globalDone();
+} );
+
+QUnit.test( "errors if callback is called after test - part 2", function( assert ) {
+	assert.throws( globalDone, /assert.async callback called after test finished./ );
+	globalDone = null;
+} );
+
```
platinumazure: There's a small part of me that wonders if we should just have a nested module to allow for a smaller scope of closure, but this looks fine as is.

trentmwillis:
@platinumazure updated with your comment

PR #1147, opened 2017-04-06
Core: Ensure assertions occur while test is running

trentmwillis:
Based on discussion in https://github.com/qunitjs/qunit/issues/1145#issuecomment-292290443, I investigated if we properly erred when asserting outside the active test, and it turns out we do not. This fixes that and adds a test.

There's a potential that this could break some currently passing test suites, but I would consider those bugs as it poor testing practice to have assertions take place outside the actual test they're associated with.

### DIFF
--- a/test/main/assert.js
+++ b/test/main/assert.js
@@ -390,3 +390,21 @@ QUnit.test( "throws", function( assert ) {
 		"throws fail when regexp doesn't match the error message"
 	);
 } );
+
+( function() {
+	var previousTestAssert;
+
+	QUnit.module( "delayed assertions" );
+
+	QUnit.test( "assertions after test finishes throws an error - part 1", function( assert ) {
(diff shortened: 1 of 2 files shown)


test/main/assert.js
```diff
@@ -390,3 +390,21 @@ QUnit.test( "throws", function( assert ) {
...
 } );
+
+( function() {
+	var globalAssert;
+
+	QUnit.module( "delayed assertions" );
+
+	QUnit.test( "assertions after test finishes throws an error - part 1", function( assert ) {
+		assert.expect( 0 );
+		globalAssert = assert;
```
Krinkle: May wanna call this `previousAssert` instead.

trentmwillis:
@Krinkle good suggestion, I've updated with a better name.

PR #1148, opened 2017-04-06
Assert: Allow assertions after async

trentmwillis:
~~Implements the feature discussed in https://github.com/qunitjs/qunit/issues/1145.~~

Allow assertions after the `assert.async()` callback is invoked. As discussed in https://github.com/qunitjs/qunit/issues/1145.

### DIFF
--- a/docs/assert/async.md
+++ b/docs/assert/async.md
@@ -17,7 +17,7 @@ Instruct QUnit to wait for an asynchronous operation.
 
 ### Description
 
-The callback returned from `assert.async()` will throw an Error if it is invoked more than once (or more often than the accepted call count, if provided).
+`assert.async()` returns a callback function and pauses test processing until the callback function is invoked the specified number of times. The callback will throw an `Error` if it is invoked more often than the accepted call count.
 
 This replaces functionality previously provided by [`QUnit.stop()`](/QUnit/stop) and [`QUnit.start()`](/QUnit/start).
 
--- a/src/assert.js
+++ b/src/assert.js
@@ -52,7 +52,6 @@ class Assert {
 			acceptCallCount = 1;
 		}
 
-		test.usedAsync = true;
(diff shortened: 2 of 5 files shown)


src/assert.js
```diff
@@ -52,7 +52,11 @@ class Assert {
 			acceptCallCount = 1;
 		}
 
-		test.usedAsync = true;
+		// If we have already used async, we don't want to set this
+		if ( !test.usedAsync ) {
+			test.usedAsync = !allowAssertionsAfter;
```
platinumazure: Should the `usedAsync` property be renamed, if we are now using it to determine if an error should be thrown for asserting (or resolving async) too late?

docs/assert/async.md
```diff
@@ -7,17 +7,20 @@ categories:
...
 
-## `async( [ acceptCallCount = 1 ] )`
+## `async( [ acceptCallCount = 1, allowAssertionsAfter = false ] )`
 
 Instruct QUnit to wait for an asynchronous operation.
 
 | name | description |
 |------|-------------|
 | `acceptCallCount` (Number) | Number of expected callbacks before the test is done. Defaults to `1`. |
+| `allowAssertionsAfter` (Boolean) | Whether to allow assertions after the returned callback is invoked. |
```
platinumazure: Just throwing this out there for consideration: If we expect more options later, should we just have an options object (which would support just the one key of `allowAssertionsAfter` for now)?

trentmwillis:
@platinumazure updated to always allow assertions after `done()`. Based on [the comment](https://github.com/qunitjs/qunit/issues/1145#issuecomment-292315734) from @Krinkle, which I agree with.

leobalter:
In a future we should rename the conventionally `done()` to `go()` because that's what we are doing and I like it.

PR #1153, opened 2017-04-09
Tests: Emit `suiteEnd` and trigger `moduleDone` for parent modules

### LINKED ISSUE #1152: `moduleDone/suiteEnd` not fired for parent module if its tests are already executed

* **QUnit version**: 2.3.0
* **What environment are you running QUnit in? (e.g., browser, Node)**: Node 6.10.0
* **How are you running QUnit? (e.g., script, testem, Grunt)**: Grunt using `grunt-contrib-qunit`

## What are you trying to do?

I want to track the end of every module/test. Basically, I have tests like the following:

```js
QUnit.module('A', function() {
  QUnit.test('a', function(assert) {
    assert.expect(1);
    assert.ok(true);
  });

  QUnit.module('B', function() {
    QUnit.test('b', function(assert) {
      assert.expect(1);
      assert.ok(true);
    });

    QUnit.module('C', function() {
      QUnit.test('c', function(assert) {
        assert.expect(1);
        assert.ok(true);
      });
    });
  });
});
```

And I log them as follow:

```js
grunt.event.on("qunit.moduleStart", function(name) {
  console.log("moduleStart:", name);
});

grunt.event.on("qunit.testStart", function(name) {
  console.log("testStart:", name);
});

grunt.event.on("qunit.testDone", function(name) {
  console.log("testDone:", name);
});

grunt.event.on("qunit.moduleDone", function(name) {
  console.log("moduleDone:", name);
});
```

I expected to get the following output:
```
moduleStart: A
testStart: a
testDone: a
moduleStart: A > B
testStart: b
testDone: b
moduleStart: A > B > C
testStart: c
testDone: c
moduleDone: A > B > C
moduleDone: A > B
moduleDone: A
```

## What actually happened?

But instead I got:

```
moduleStart: A
testStart: a
testDone: a
moduleStart: A > B
testStart: b
testDone: b
moduleStart: A > B > C
testStart: c
testDone: c
moduleDone: A > B > C
```

I tracked the issue down to these lines [`src/test.js#L282-L292`](https://github.com/qunitjs/qunit/blob/63bcd15061fc706c9a058ad71c32ada8550ff39e/src/test.js#L282-L292):
```js
if ( module.testsRun === numberOfTests( module ) ) {
	emit( "suiteEnd", module.suiteReport.end( true ) );
	runLoggingCallbacks( "moduleDone", {
		name: module.name,
		tests: module.tests,
		failed: module.stats.bad,
		passed: module.stats.all - module.stats.bad,
		total: module.stats.all,
		runtime: now() - module.stats.started
	} );
}
```

The above code should also check parent modules, otherwise `moduleDone` and `suiteEnd` will be triggered only for the module `C`.

I can provide a PR in case this is not the intended behaviour.

Thanks!

Arkni:
Fixes #1152

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -280,6 +280,20 @@ Test.prototype = {
 		} );
 
 		if ( module.testsRun === numberOfTests( module ) ) {
+			logSuiteEnd( module );
+
+			// Check if the parent modules, iteratively, are done. If that the case,
+			// we emit the `suiteEnd` event and trigger `moduleDone` callback.
+			let parent = module.parentModule;
+			while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+				logSuiteEnd( parent );
+				parent = parent.parentModule;
+			}
+		}
+
+		config.current = undefined;
+
+		function logSuiteEnd( module ) {
 			emit( "suiteEnd", module.suiteReport.end( true ) );
 			runLoggingCallbacks( "moduleDone", {
 				name: module.name,
@@ -290,8 +304,6 @@ Test.prototype = {
 				runtime: now() - module.stats.started
 			} );
 		}
-
-		config.current = undefined;
 	},
 
 	preserveTestEnvironment: function() {
--- a/test/callbacks.js
+++ b/test/callbacks.js
@@ -92,9 +92,26 @@ QUnit.done( function() {
 			"log1",
 			"log2",
 			"module1 > afterEach",
+			"testDone1",
+			"testDone2",
+			"moduleStart1",
+			"moduleStart2",
+			"testStart1",
+			"testStart2",
+			"module4 > before",
+			"module1 > beforeEach",
+			"module4 > beforeEach",
+			"module4 > test1",
+			"log1",
+			"log2",
+			"module4 > afterEach",
+			"module1 > afterEach",
+			"module4 > after",
 			"module1 > after",
 			"testDone1",
 			"testDone2",
+			"moduleDone1",
+			"moduleDone2",
 			"moduleDone1",
 			"moduleDone2",
 			"done1",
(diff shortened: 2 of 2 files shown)


src/test.js
```diff
@@ -280,6 +280,20 @@ Test.prototype = {
...
+			let parent = module.parentModule;
+			while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+				declareModuleEnd( parent );
+				parent = parent.parentModule;
+			}
+		}
+
+		config.current = undefined;
+
+		function declareModuleEnd( module ) {
```
Arkni: I'm open for a new name for this function.

trentmwillis: Can we rename this `logSuiteEnd`? It will be similar to `logAssertion` which exists elsewhere, and prefers the js-reporters name (which is what we're moving towards).

src/test.js
```diff
@@ -280,6 +280,20 @@ Test.prototype = {
 		} );
 
 		if ( module.testsRun === numberOfTests( module ) ) {
+			declareModuleEnd( module );
+
+			// Check if the parent modules, recursively, are done. If that the case,
```
trentmwillis: Nitpick: This isn't recursive, it's iterative.

Gruntfile.js
```diff
@@ -140,6 +140,7 @@ module.exports = function( grunt ) {
 				"test/reporter-html/window-onerror-preexisting-handler.html",
 				"test/reporter-urlparams.html",
 				"test/moduleId.html",
+				"test/module-done.html",
```
trentmwillis: I'd prefer to avoid adding an additional test file, can you just update the existing tests, maybe [here](https://github.com/qunitjs/qunit/blob/63bcd15061fc706c9a058ad71c32ada8550ff39e/test/callbacks.js#L144)?

src/test.js
Arkni: Sorry, my bad. I'll update it!

src/test.js
Arkni: Will update!
Thanks!

Gruntfile.js
Arkni: I'm working on it right now!

Arkni:
Updated per above suggestions!

PR #1154, opened 2017-04-10
Tests: Reject async test promise when config.notrycatch is set.

### LINKED ISSUE #1149: Rejected promise not throwing an uncaught exception with "no try-catch"

* **QUnit version**: QUnit 2.0.0; Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36
* **What environment are you running QUnit in? (e.g., browser, Node)**: Chrome 57.0.2987
* **How are you running QUnit? (e.g., script, testem, Grunt)**: HTML test runner

I have a test which returns a promise which fails. I generally have no try-catch on so that I can debug errors in my IDE, but this doesn't seem to work in conjunction with promises. Is this by design or a bug?

    QUnit.test('promise fail', function (assert) {
        return new Promise((resolve, reject) => {
            resolve()
        }).then(() => {
            throw new Error("hurk")
        })
    })

HTML Reporter shows: 
Promise rejected during promise fail: hurk@ 1 ms
Source: 	
Error: hurk
    at Promise.then (http://localhost:63342/cs/out/test/PICurve.qunit.js:18:15)

NaridaL:
Fixes #1149 

Soo.. as far as I can tell the following is semantically equivalent to the changed version in the PR. But the resume at // unblock is undefined?
EDIT: NVM

### DIFF
--- a/docs/config/QUnit.config.md
+++ b/docs/config/QUnit.config.md
@@ -63,6 +63,8 @@ By default, QUnit will run tests within a `try-catch` block to prevent uncaught 
 
 Enabling this flag will run tests without the `try-catch` to allow exceptions to remain uncaught for easier debugging in certain environments.
 
+If set, no rejection handler will be attached to promises returned by tests. This has a similar effect and allows debuggers to pause at the right location if an error occurs.
+
 ### `QUnit.config.noglobals` (boolean) | default: `false`
 
 Enabling this flag will cause QUnit to check if any new properties have been added to the global context after each test. New global properties being found will result in test failures to help ensure your tests are not leaking state.
--- a/src/test.js
+++ b/src/test.js
@@ -437,23 +437,27 @@ Test.prototype = {
 			then = promise.then;
 			if ( objectType( then ) === "function" ) {
 				resume = internalStop( test );
-				then.call(
-					promise,
-					function() { resume(); },
-					function( error ) {
-						message = "Promise rejected " +
-							( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
-							" \"" + test.testName + "\": " +
-							( ( error && error.message ) || error );
-						test.pushFailure( message, extractStacktrace( error, 0 ) );
+				if ( config.notrycatch ) {
+					then.call( promise, function() { resume(); } );
+				} else {
+					then.call(
+						promise,
+						function() { resume(); },
+						function( error ) {
+							message = "Promise rejected " +
+								( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
+								" \"" + test.testName + "\": " +
+								( ( error && error.message ) || error );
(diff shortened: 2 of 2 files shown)


NaridaL:
@platinumazure I'm not sure how. This patch doesn't cause an exception to be thrown. Promises generally fail silently unless a handler is attached with catch(), which is what Test.resolvePromise() does. However it doesn't actually return the modified promise, so there's no way to externally check that the error goes unhandled. 

The debugger on the other hand, notices that the promise goes unhandled and can pause execution. That's not really possible to do as a simple test. Also, the original notrycatch doesn't have a test either.

leobalter:
can't we fetch the difference if we have an isolated test setting the value for `config.notrycatch`?

NaridaL:
@leobalter Fetch the difference of what?

trentmwillis:
@NaridaL could we have an isolated test file that sets up a global error handler? Basically, with `notrycatch=false` we would expect the error handler to never be called, but with `notrycatch=true` we would expect the error handler to be called.

NaridaL:
@trentmwillis I don't think so.
```
try {
  new Promise((resolve, reject) => resolve("foo")).then(result => { throw new Error() })
} catch (e) {
  // Execution will never reach this point. Promises fail silently, the only way to find out 
  // it fails is by attaching a rejection handler with .catch() to the promise. As we don't 
  // have a reference to the promise object, we can't do this afterwards.
}
```
The debugger can catch rejections which go unhandled and pause execution, but it isn't possible in pure JS.

trentmwillis:
You can listen for the `unhandledRejection` event: https://nodejs.org/api/process.html#process_event_unhandledrejection

Due to [spotty browser support](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection#Browser_compatibility), we can just test it in Node which would be enough verification for the logic you've added.

NaridaL:
I stand corrected. How would one go about testing a test?

leobalter:
[monkey patch](https://en.wikipedia.org/wiki/Monkey_patch) a test, observe the results and assert they are as expected. If it adds a coverage in the new code, that's +1 win.

You can find examples of times we've had to do something similar checking our files from our test folder, e.g.: 
 https://github.com/qunitjs/qunit/blob/baff17dbc12168c139dc339ba1e0d07e4a7e8215/test/onerror/inside-test.js#L5

NaridaL:
So my first approach was to create a new test with Object.create( Object.getPrototypeOf( QUnit.config.current ) and use a proxy to make QUnit.config transparently readonly, but QUnit.config is only a reference and I can't change the original object.

Then I tried QUnit.config.current.resolvePromise(testPromise), but resolvePromise uses internalStop to halt testing execution, and only resumes when the error handler executes, which doesn't exist with notrycatch = true.

PR #1155, opened 2017-04-11
HTML Reporter: numerical diff of differing numbers instead of str-diff.

NaridaL:
The HTML reporter does a str diff on differing numbers, which is uniquely useless. This shows the numerical difference. This checks if the values actually have typeof number, which I find preferable, but I guess checking if both values are parsable as numbers is also possible.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -785,19 +785,16 @@ export function escapeText( s ) {
 					showDiff = true;
 					diff = ( details.actual - details.expected );
 					diff = ( diff > 0 ? "+" : "" ) + diff;
-				} else {
+				} else if ( "boolean" !== typeof details.actual &&
+							"boolean" !== typeof details.expected ) {
+					diff = QUnit.diff( expected, actual );
 
-					// Don't show diff if actual or expected are booleans
-					if ( !( /^(true|false)$/.test( actual ) ) &&
-						!( /^(true|false)$/.test( expected ) ) ) {
-						diff = QUnit.diff( expected, actual );
-						showDiff = stripHtml( diff ).length !==
-							stripHtml( expected ).length +
-							stripHtml( actual ).length;
-					}
+					// don't show diff if there is zero overlap
+					showDiff = stripHtml( diff ).length !==
+						stripHtml( expected ).length +
+						stripHtml( actual ).length;
 				}
 
-				// Don't show diff if expected and actual are totally different
 				if ( showDiff ) {
 					message += "<tr class='test-diff'><th>Diff: </th><td><pre>" +
 					diff + "</pre></td></tr>";
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -782,9 +782,11 @@ export function escapeText( s ) {
 					escapeText( actual ) + "</pre></td></tr>";
 
 				if ( typeof details.actual === "number" && typeof details.expected === "number" ) {
-					showDiff = true;
-					diff = ( details.actual - details.expected );
-					diff = ( diff > 0 ? "+" : "" ) + diff;
+					if ( !isNaN( details.actual ) && !isNaN( details.expected ) ) {
+						showDiff = true;
+						diff = ( details.actual - details.expected );
+						diff = ( diff > 0 ? "+" : "" ) + diff;
+					}
 				} else if ( typeof details.actual !== "boolean" &&
 							typeof details.expected !== "boolean" ) {
 					diff = QUnit.diff( expected, actual );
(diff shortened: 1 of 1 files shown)


reporter/html.js
```diff
@@ -779,18 +779,25 @@ export function escapeText( s ) {
...
-					diff = QUnit.diff( expected, actual );
-					showDiff = stripHtml( diff ).length !==
-					stripHtml( expected ).length +
-					stripHtml( actual ).length;
+					escapeText( actual ) + "</pre></td></tr>";
+
+				if ( "number" === typeof details.actual && "number" === typeof details.expected ) {
+					showDiff = true;
+					diff = ( details.actual - details.expected );
+					diff = ( diff > 0 ? "+" : "" ) + diff;
```
gibson042: Just for code style, this should probably be an `if` block.

reporter/html.js
```diff
@@ -779,18 +779,25 @@ export function escapeText( s ) {
...
+				} else {
+
+					// Don't show diff if actual or expected are booleans
+					if ( !( /^(true|false)$/.test( actual ) ) &&
```
gibson042: And this can be an `else if`.

NaridaL:
While I'm at it, the regex test for true/false seems kinda iffy. Wouldn't it make more sense to test if the non-stringified values were actual booleans? (A str-diff for "true" and "rue" technically makes sense, even if it doesn't really matter for strings this short.)

reporter/html.js
```diff
@@ -779,18 +779,22 @@ export function escapeText( s ) {
...
 
 				message += "<tr class='test-actual'><th>Result: </th><td><pre>" +
-				escapeText( actual ) + "</pre></td></tr>";
-
-			// Don't show diff if actual or expected are booleans
-				if ( !( /^(true|false)$/.test( actual ) ) &&
-					!( /^(true|false)$/.test( expected ) ) ) {
+					escapeText( actual ) + "</pre></td></tr>";
+
+				if ( "number" === typeof details.actual && "number" === typeof details.expected ) {
```
leobalter: why is the typeof in the right side? This feels inverted.

Btw, we also have a `core/utilities` module checking for objectType. typeof is pretty much ok for number types but this order is confusing.

Other question: should we also confirm that neither details.actual and details.expected are not NaN?

Otherwise, a couple lines below, `diff = ( details.actual - details.expected );`, diff will be `NaN` as well.

reporter/html.js
leobalter: diff would become `"NaN"` here.

reporter/html.js
NaridaL: diff being NaN makes sense the way I see it.

leobalter: but if `details.actual` is `42` and `details.expected` is `NaN`, `diff` will become NaN only and won't really tell what is going on, right?

reporter/html.js
```diff
@@ -782,9 +782,11 @@ export function escapeText( s ) {
 					escapeText( actual ) + "</pre></td></tr>";
 
 				if ( typeof details.actual === "number" && typeof details.expected === "number" ) {
-					showDiff = true;
-					diff = ( details.actual - details.expected );
-					diff = ( diff > 0 ? "+" : "" ) + diff;
+					if ( !isNaN( details.actual ) && !isNaN( details.expected ) ) {
```
leobalter: I know we are protecting this line with the type check above, but the global `isNaN` is still [broken and unreliable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN). It makes me want to protect the code from further weirdness.

`Number.isNaN` is doing the right thing. Unfortunately, it is not available in IE9, which we support.

If can use a polyfill...

```js
Number.isNaN = Number.isNaN || function(value) {     
    return value !== value;
}
```

or just:

```js
// Confirm the values are not NaN
if ( details.actual === details.actual && details.expected === details.expected ) {
```

I can confirm JS won't have any other value that returns false in a strict comparison, it would break the web, actually. So that is reliable.

NaridaL: After the type check isNaN is fine. Feel free to fix it; I'm leaving it.

gibson042: @leobalter I actually like this PR's original output for NaN cases (i.e., before introducing the isNaN checks):
> **Expected:** `NaN`
> **Result:** `42`
> **Diff**: `NaN`

leobalter: you're right, @gibson042.

@NaridaL I'm sorry if I'm being annoying, but if you want, we could revert these changes to the original output. I appreciate your time supporting QUnit.

NaridaL: As expected/actual are displayed anyway, I don't think it matters what the diff says when a value is NaN. As it doesn't really add any information we might as well not show it.

leobalter:
@trentmwillis I'm +1 to land this as it is rather than adding more bikeshed.

leobalter:
@trentmwillis I'm +1 to land this as it is rather than adding more bikeshed.

PR #1156, opened 2017-04-12
Core: Fix bug calling hooks with skipped tests

bendemboski:
When the first/last test in a module was skipped, it would cause the module's
before()/after() hook to not be called. This ensures those hooks will always be
called.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -44,8 +44,10 @@ function createModule( name, testEnvironment ) {
 		name: moduleName,
 		parentModule: parentModule,
 		tests: [],
+		unskippedTests: [],
 		moduleId: generateHash( moduleName ),
 		testsRun: 0,
+		unskippedTestsRun: 0,
 		childModules: [],
 		suiteReport: new SuiteReport( name, parentSuite )
 	};
--- a/src/core.js
+++ b/src/core.js
@@ -44,7 +44,6 @@ function createModule( name, testEnvironment ) {
 		name: moduleName,
 		parentModule: parentModule,
 		tests: [],
-		unskippedTests: [],
 		moduleId: generateHash( moduleName ),
 		testsRun: 0,
 		unskippedTestsRun: 0,
--- a/src/test.js
+++ b/src/test.js
@@ -761,17 +761,13 @@ function numberOfUnskippedTests( module ) {
 }
 
 function notifyTestsRan( module, skipped ) {
-	const thisModule = module;
-
 	module.testsRun++;
+	if ( !skipped ) {
+		module.unskippedTestsRun++;
+	}
 	while ( ( module = module.parentModule ) ) {
 		module.testsRun++;
-	}
-
-	if ( !skipped ) {
-		module = thisModule;
-		module.unskippedTestsRun++;
-		while ( ( module = module.parentModule ) ) {
+		if ( !skipped ) {
 			module.unskippedTestsRun++;
 		}
 	}
--- a/src/test.js
+++ b/src/test.js
@@ -64,6 +64,10 @@ export default function Test( settings ) {
 		this.async = false;
 		this.expected = 0;
 	} else {
+		this.module.unskippedTests.push( {
+			name: this.testName,
+			testId: this.testId
+		} );
 		this.assert = new Assert( this );
 	}
 }
@@ -158,7 +162,7 @@ Test.prototype = {
 			test = this;
 		return function runHook() {
 			if ( hookName === "before" ) {
-				if ( hookOwner.testsRun !== 0 ) {
+				if ( hookOwner.unskippedTestsRun !== 0 ) {
 					return;
 				}
 
@@ -166,7 +170,7 @@ Test.prototype = {
 			}
 
 			if ( hookName === "after" &&
-				hookOwner.testsRun !== numberOfTests( hookOwner ) - 1 &&
+				hookOwner.unskippedTestsRun !== numberOfUnskippedTests( hookOwner ) - 1 &&
 				config.queue.length > 2 ) {
 				return;
 			}
@@ -245,7 +249,7 @@ Test.prototype = {
 			}
 		}
 
-		notifyTestsRan( module );
+		notifyTestsRan( module, skipped );
 
 		// Store result when possible
 		if ( storage ) {
@@ -737,23 +741,41 @@ function internalStart( test ) {
 	}
 }
 
-function numberOfTests( module ) {
-	let count = module.tests.length;
+function collectTests( module, key ) {
+	let tests = [].concat( module[ key ] );
 	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
 		const nextModule =  modules.shift();
-		count += nextModule.tests.length;
+		tests = tests.concat( nextModule[ key ] );
 		modules.push( ...nextModule.childModules );
 	}
 
-	return count;
+	return tests;
 }
 
-function notifyTestsRan( module ) {
+function numberOfTests( module ) {
+	return collectTests( module, "tests" ).length;
+}
+
+function numberOfUnskippedTests( module ) {
+	return collectTests( module, "unskippedTests" ).length;
+}
+
+function notifyTestsRan( module, skipped ) {
+	const thisModule = module;
+
 	module.testsRun++;
 	while ( ( module = module.parentModule ) ) {
 		module.testsRun++;
+	}
+
+	if ( !skipped ) {
+		module = thisModule;
+		module.unskippedTestsRun++;
+		while ( ( module = module.parentModule ) ) {
+			module.unskippedTestsRun++;
+		}
 	}
 }
--- a/src/test.js
+++ b/src/test.js
@@ -742,13 +742,13 @@ function internalStart( test ) {
 }
 
 function collectTests( module, key ) {
-	let tests = [].concat( module[ key ] );
+	const tests = [].concat( module[ key ] );
 	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
 		const nextModule =  modules.shift();
-		tests = tests.concat( nextModule[ key ] );
+		tests.push.apply( tests, nextModule[ key ] );
 		modules.push( ...nextModule.childModules );
 	}
 
--- a/src/test.js
+++ b/src/test.js
@@ -54,7 +54,8 @@ export default function Test( settings ) {
 
 	this.module.tests.push( {
 		name: this.testName,
-		testId: this.testId
+		testId: this.testId,
+		skip: !!settings.skip
 	} );
 
 	if ( settings.skip ) {
@@ -64,10 +65,6 @@ export default function Test( settings ) {
 		this.async = false;
 		this.expected = 0;
 	} else {
-		this.module.unskippedTests.push( {
-			name: this.testName,
-			testId: this.testId
-		} );
 		this.assert = new Assert( this );
 	}
 }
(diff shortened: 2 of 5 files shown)


rwjblue:
Nice, sleuth work!

test/main/modules.js
```diff
@@ -49,6 +49,33 @@ QUnit.test( "does not run before subsequent tests", function( assert ) {
...
+} );
+
+QUnit.skip( "skipped, but should still run before hook", function( assert ) {
+	assert.expect( 1 );
+	assert.ok( true );
+} );
+
+QUnit.test( "has run by the time second (unskipped) test runs", function( assert ) {
+	assert.expect( 1 );
+	assert.equal( this.beforeCount, 1, "beforeCount should be one" );
```
leobalter: if we remove the `this.assert = new Assert( this );` for skipped tests line will this fail?

test/main/modules.js
```diff
@@ -63,6 +90,31 @@ QUnit.test( "runs after final test", function( assert ) {
...
+QUnit.skip( "skipped, but should still run after hook", function( assert ) {
+	assert.expect( 1 );
+} );
+
+QUnit.module( "after (skip) verifier" );
+
+QUnit.test( "after hook ran after last (skipped) test", function( assert ) {
+	assert.expect( 1 );
+	assert.ok( ranAfterHook, "after hook ran" );
+} );
```
leobalter: can we nest the `after (skip)` and `after (skip) verifier` modules in a single module? mostly a

```js
QUnit.module( "after (skip)", function() {
	var ranAfterHook = 0;
	QUnit.module( "main", function(hooks) {
		hooks.after = function( assert ) {
			ranAfterHook += 1;
		};

		QUnit.test( "does not run after initial tests", ... );
		QUnit.skip( "skipped, but should still run after hook", ... );
	} );

	QUnit.module( "verifier", function() {
		QUnit.test( "after hook ran after last (skipped) test", ... );
	} );
} );
```

test/main/modules.js
```diff
@@ -63,6 +90,31 @@ QUnit.test( "runs after final test", function( assert ) {
...
+	}
+} );
+
+QUnit.test( "does not run after initial tests", function( assert ) {
+	assert.expect( 0 );
+	ranAfterHook = false;
+} );
+
+QUnit.skip( "skipped, but should still run after hook", function( assert ) {
+	assert.expect( 1 );
```
leobalter: this should not be necessary. Some throwing expression here, like `throw 'foo'` here would help better

test/main/modules.js
```diff
@@ -49,6 +49,33 @@ QUnit.test( "does not run before subsequent tests", function( assert ) {
...
+			this.beforeCount = 0;
+		}
+
+		this.beforeCount++;
+	}
+} );
+
+QUnit.skip( "skipped, but should still run before hook", function( assert ) {
+	assert.expect( 1 );
+	assert.ok( true );
```
leobalter: this should not be necessary. Some throwing expression here, like throw 'foo' here would help better, or even `assert.ok( false );`

leobalter:
> if all tests in a module were skipped, then before/after will never run

+1

bendemboski:
@trentmwillis @leobalter how does this look? These changes didn't really have much in common with the previous version, so I squashed the commits.

src/test.js
```diff
@@ -737,23 +741,41 @@ function internalStart( test ) {
...
-	let count = module.tests.length;
+function collectTests( module, key ) {
+	let tests = [].concat( module[ key ] );
 	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
 		const nextModule =  modules.shift();
-		count += nextModule.tests.length;
+		tests = tests.concat( nextModule[ key ] );
```
leobalter: `tests.push.apply( tests, nextModule[ keys ] );`

it avoids creating a new array and discarding the previous one.

test/main/modules.js
```diff
@@ -63,6 +87,46 @@ QUnit.test( "runs after final test", function( assert ) {
...
+		hooks.after = function() {
+			ranAfterHook += 1;
+		};
+
+		QUnit.skip( "only test in module is skipped" );
+	} );
+
+	QUnit.module( "verifier", function() {
+		QUnit.test( "hooks did not run", function( assert ) {
+			assert.expect( 2 );
```
leobalter: this is not necessary here.

test/main/modules.js
```diff
@@ -63,6 +87,46 @@ QUnit.test( "runs after final test", function( assert ) {
...
+
+QUnit.test( "does not run after initial tests", function( assert ) {
+	assert.expect( 0 );
+} );
+
+QUnit.test( "runs after final unskipped test", function( assert ) {
+	assert.expect( 1 );
+} );
+
+QUnit.skip( "last test in module is skipped" );
```
leobalter: this is nice!

bendemboski:
@leobalter I addresses your feedback in new commit so you can easily see the diff, but I'd like to squash it with the previous before merging, so let me know if the latest commit is 👍  by you, and if so I'll squash them and wait for @trentmwillis to provide feedback or merge.

src/core.js
```diff
@@ -44,8 +44,10 @@ function createModule( name, testEnvironment ) {
 		name: moduleName,
 		parentModule: parentModule,
 		tests: [],
+		unskippedTests: [],
 		moduleId: generateHash( moduleName ),
 		testsRun: 0,
+		unskippedTestsRun: 0,
```
trentmwillis: `testsRun` isn't used for anything other than determining whether to invoke module hooks, so we can just use that instead of introducing a new variable. Basically, we should count a test that actually ran should be an unskipped test. Make sense?

src/test.js
```diff
@@ -64,6 +64,10 @@ export default function Test( settings ) {
 		this.async = false;
 		this.expected = 0;
 	} else {
+		this.module.unskippedTests.push( {
```
trentmwillis: We should create a single object up on line 55 and push it into both the `tests` and `unskippedTests` array, since we want that info to be in sync.

Also, I'd appreciate a flag `skipped` to be added to the object, for more info.

src/test.js
```diff
@@ -737,23 +741,41 @@ function internalStart( test ) {
...
-	let count = module.tests.length;
+function collectTests( module, key ) {
+	const tests = [].concat( module[ key ] );
 	const modules = [ ...module.childModules ];
 
 	// Do a breadth-first traversal of the child modules
 	while ( modules.length ) {
 		const nextModule =  modules.shift();
-		count += nextModule.tests.length;
+		tests.push.apply( tests, nextModule[ key ] );
```
trentmwillis: I'd prefer to keep this as a counter instead of an array as we don't need the flattened array for anything other than length.

src/core.js
bendemboski: I created a new property because `testsRun` is also referenced [here](https://github.com/qunitjs/qunit/blob/master/src/test.js#L77) and [here](https://github.com/qunitjs/qunit/blob/master/src/test.js#L282). I'm not sure what the first one is doing, but if we change the meaning of `testsRun` then per the second one, if a module ends with skipped tests, wouldn't we log the end of the suite before logging those skipped tests?

src/test.js
bendemboski: I actually considered keeping just one array of tests (`module.tests`) and adding a `skip` flag to its entries. That way we wouldn't need anything extra because we can always find out if the current test is the first/last unskipped test by looking through `module.tests` (and child modules) and comparing `testId` to `this.testId` and looking at `skip` flags.

I didn't go in that direction because `module.tests` is passed into various callbacks as the context, and I didn't want to make any public-facing changes. If you prefer that route (then we don't need any new counts or arrays or anything, just a `skip` flag on the `module.tests` entries), I can certainly do that.

If not, can you elaborate on exactly where you want the `skipped` flag? From your comment I'm not entirely sure.

src/core.js
trentmwillis: Ah good point, carry on.

src/test.js
trentmwillis: I think adding the flag is fine, it's an additive change so it should be considered backwards compatible.

bendemboski: Blerg. Turns out we need to keep an unskipped count. I was hoping to just look at the list of tests and when we run the first/last unskipped test, run the before/after hook, but when looking through the `module`/`childModule`/`tests` tree, we can't tell the difference between this:

```
QUnit.module("outer", function() {
  test("outer test", function() { ... });
  QUnit.module("inner", function() {
    test("inner test", function() { ... });
  });
});
```

and this:

```
QUnit.module("outer", function() {
  QUnit.module("inner", function() {
    test("inner test", function() { ... });
  });
  test("outer test", function() { ... });
});
```

So we don't know whether to run `outer`'s `before` hook when we run `outer` or `inner`. Without some knowledge of how a module's tests are ordered with respect to its child modules, this approach will not work.

So I guess I'll fall back on still adding the `skip` flag to `module.tests` (so we can dynamically count the number of unskipped tests in a module and don't have to keep a separate list/count of them), but also keeping an `unskippedTestsRun` on the module.

bendemboski:
@trentmwillis I've addressed your feedback, but left the commits un-squashed so it's easier to see the changes. Let me know if you'd like me to squash this PR down to a single commit (or if you just want to do it as part of the merge).

src/test.js
```diff
@@ -737,23 +738,41 @@ function internalStart( test ) {
...
-function notifyTestsRan( module ) {
+function notifyTestsRan( module, skipped ) {
+	const thisModule = module;
+
 	module.testsRun++;
 	while ( ( module = module.parentModule ) ) {
 		module.testsRun++;
 	}
+
+	if ( !skipped ) {
```
trentmwillis: We should perform this check inside the `while` loop above so that we're only iterating once.

test/main/modules.js
```diff
@@ -63,6 +87,45 @@ QUnit.test( "runs after final test", function( assert ) {
...
+
+QUnit.module( "before/after with all tests skipped (wrapper)", function() {
```
trentmwillis: We can make this test much simpler:

```js
QUnit.module( "before/after with all tests skipped", {
  before: function( assert ) {
    assert.ok( false, "should not occur" );
  },
  after: function( assert ) {
    assert.ok( false, "should not occur" );
  }
} );

QUnit.skip( "verifier" );
```

test/main/modules.js
```diff
@@ -49,6 +49,30 @@ QUnit.test( "does not run before subsequent tests", function( assert ) {
...
+
+		this.beforeCount++;
+	}
+} );
+
+QUnit.skip( "first test in module is skipped" );
+
+QUnit.test( "runs before first unskipped test", function( assert ) {
+	assert.expect( 2 );
+	assert.equal( this.beforeCount, 1, "beforeCount should be one" );
```
trentmwillis: We can ditch the `beforeCount` stuff. Since we `assert.expect()` we know whether or not the hook ran.

src/test.js
bendemboski: I cringe at checking a loop invariant inside the loop, although this time I guess I went the direction of incurring a performance cost in the common case rather that the uncommon one :man_facepalming: 

How about this:

```
let notify;
if ( skipped ) {
  notify = function(module) {
    module.testsRun++;
  };
} else {
  notify = function(module) {
    module.testsRun++;
    module.unskippedTestsRun++;
  };
}

notify(module);
while ( ( module = module.parentModule ) ) {
  notify(module);
}
```

Not a big deal if you don't like it and prefer just checking the invariant inside the loop -- let me know.

test/main/modules.js
bendemboski: I was following the convention in a few other tests in this file like [this](https://github.com/qunitjs/qunit/blob/master/test/main/modules.js#L42) one. I'm happy either way, but want to double-check that you prefer terseness over consistency here -- let me know which way you prefer.

test/main/modules.js
bendemboski: 👍 -- let me know your decision on the other two comments, and I'll roll all three into one commit.

src/test.js
trentmwillis: I'd prefer just checking the invariant. It's a fast check and less code to reason about.

test/main/modules.js
trentmwillis: I'd prefer the terse version. It's likely that the other ones should be changed as well, but that doesn't need to be done here.

bendemboski:
Okay @trentmwillis comments addressed. Let me know if you see anything else or want me to squash commits, otherwise merge away.

trentmwillis:
Thanks again @bendemboski!

PR #1159, opened 2017-04-13
HTML Reporter: Add 'Run module' link to test title

dalefukami:


platinumazure:
Hi @dalefukami, thanks for the contribution!

I know the HTML Reporter's test coverage is very incomplete, but do you think it might be possible to add a unit test for this change? It can be as simple as you like-- we just want to make sure that that link shows up in a test.

dalefukami:
Sure thing!

dalefukami:
@platinumazure Does this test look reasonable to you?

test/reporter-html/reporter-html.js
```diff
@@ -44,6 +49,20 @@ QUnit.test( "running test name displayed", function( assert ) {
 	);
 } );
 
+QUnit.test( "previous test displays filtering run options", function( assert ) {
+	assert.expect( 2 );
+
+	var previous = this.getPreviousTest( assert ),
+		filteringLinks = previous.getElementsByTagName( "a" );
+
+	assert.ok( /^Rerun$/.test( filteringLinks[ 0 ].innerHTML ),
```
platinumazure: Given that you're basically doing a full string compare, why not just use `assert.strictEqual( filteringLinks[ 0 ].innerHTML, "'Rerun' link displayed" );`?

test/reporter-html/reporter-html.js
```diff
@@ -44,6 +49,20 @@ QUnit.test( "running test name displayed", function( assert ) {
...
+		"'Rerun' link displayed"
+	);
+	assert.ok( /^Run module$/.test( filteringLinks[ 1 ].innerHTML ),
```
platinumazure: Given that you're basically doing a full string compare, why not just use `assert.strictEqual( filteringLinks[ 1 ].innerHTML, "'Run module' link displayed" );`?

trentmwillis:
Overall, I like the idea of adding a quick way to filter to a specific module, but I'm not sure this is the best approach. I would actually like to bake in better support for modules (including nested modules) by breaking up the test list into collapsible sections based on module. Then, since we'd have sections based on modules, it'd be easy to have a "Run module" option at the start of the section. Thoughts? I know it is a lot more work, but I think it would make the overall presentation much nicer.

gibson042:
I agree with @trentmwillis; this feels a bit too heavy. I would support incremental improvements, though... maybe exposing `?moduleId=…` links in per-module "qunit-tests" \<li> children and/or the "qunit-filteredTest" div.

test/reporter-html/reporter-html.js
leobalter: I would say tags, but we have `^` and `$` tags in the regexp.

dalefukami:
@trentmwillis sorry for leaving this hanging. I had thrown it together quickly during a moment of spare time and then got distracted. I'm not likely to be able to pick this up again any time soon.

PR #1160, opened 2017-04-13
Core: extractStacktrace removes config.projectPath from lines.

NaridaL:
I added a config option config.projectPath: string, which is removed from stack traces in extractStacktrace in order to clean up the output. e.g. config.projectPath = "http://localhost:12345/".

Is there any interest in pulling this?

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -51,7 +51,11 @@ const config = {
 	callbacks: {},
 
 	// The storage module to use for reordering tests
-	storage: localSessionStorage
+	storage: localSessionStorage,
+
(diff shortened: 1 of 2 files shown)


trentmwillis:
I'd rather not add an additional config option for this minor presentation thing. In the reporter, however, we could take into account the `location.origin` value and strip that from the stack trace, would that suffice?

leobalter:
Like @trentmwillis said, we can already solve this through the window api, it's not worth the cost of adding one more option to our config.

This is also being used to clean the presented stack traces, I believe it's out of our scope if the stacktrace api contains the whole path or not.

PR #1164, opened 2017-04-19
CLI: Prefer local version of QUnit

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1158.

### DIFF
--- a/bin/require-qunit.js
+++ b/bin/require-qunit.js
@@ -13,13 +13,13 @@ module.exports = function requireQUnit() {
 		try {
 
 			// Second, we use the globally installed QUnit
-			delete require.cache[ require.resolve( "../qunit/qunit" ) ];
+			delete require.cache[ resolve.sync( "../qunit/qunit" ) ];
 			return require( "../qunit/qunit" );
 		} catch ( e ) {
 			if ( e.code === "MODULE_NOT_FOUND" ) {
 
 				// Finally, we use the local development version of QUnit
-				delete require.cache[ require.resolve( "../dist/qunit" ) ];
+				delete require.cache[ resolve.sync( "../dist/qunit" ) ];
 				return require( "../dist/qunit" );
 			}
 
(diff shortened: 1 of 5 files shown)


bin/require-qunit.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Depending on the exact usage, QUnit could be in one of several places, this
+// function handles finding it.
+module.exports = function requireQUnit() {
+	try {
+
+		// Something is borked here for tests
+
+		// First we attempt to find QUnit relative to the current working directory.
+		const localQUnitPath = resolve( "qunitjs", { basedir: process.cwd() } );
```
Arkni: I think you meant to use `resolve.sync(...)` instead of the async  `resolve(...)`

trentmwillis: Yep! Good catch.

trentmwillis:
PR has been updated and I have included unit tests for the complex `require` logic.

test/cli/require-qunit-test.js
```diff
@@ -0,0 +1,59 @@
+const proxyquire = require( "proxyquire" );
+const resolveStub = {
+	sync( path ) {
+		return path;
+	}
+};
+
+QUnit.module( "requireQUnit", function() {
+	QUnit.test( "finds QUnit in the current working directory", function( assert ) {
+		const localQUnit = {
```
platinumazure: Is this worth extracting to a module-scope variable since it has the same value? (In theory, any variations in individual tests that might come up later could then be handled with `Object.assign()` or similar.)

trentmwillis: We can do that if you feel it is really needed. I opted for the current format so that the variables had more detailed names within each test.

platinumazure: No, I don't feel it's particularly needed. I just wanted to see if there was a particular rationale or if it happened organically/unintentionally. Go ahead and merge if you like (or wait for others to review).

Issue #1132, opened 2017-03-30
Async tests can abruptly stop in standalone Node.js mode


getify:
I'm not sure if this behavior is by-design, or if there's just no way for QUnit to fix it, but this situation just bit me and it was very confusing.

Here's my minimal test case:

```js
"use strict";

var QUnit = require("qunit");

QUnit.done(function done(results) {
	console.log("done:",results);
});

QUnit.test("temp",function(assert){
	var done = assert.async();
	assert.expect(1);

	setTimeout(function(){
		console.log("timer!");
	},1000);
});

QUnit.test( "temp2", function(assert){
	assert.expect(1);
	assert.ok(true,"temp2");
});

QUnit.start();
```

And when I run this:

```
$] node tmp.js
timer!
$]
```

In other words, the first test case fires the timer, but since it never runs the `done()`, it just silently dies. Since node doesn't have any other waiting event handlers, node just exits. The test suite doesn't fire my completion handler to notify me that a test failed to meet its assertion, and it doesn't try to run the second test at all. It just stops.

And even worse, the test suite exits with a zero exit code, meaning my other CLI tools consider this a passing test suite run. :(

This seems like broken behavior to me. Shouldn't there be some sort of timeout in the background for async tests, or something like that?

Or maybe qunit could register a `process.on("exit", ..)` handler, that if the test suite hasn't completed by the time the process tries to finish, the handler forces a non-zero exit code to signal failure to the CLI, and emits some error about an abnormally terminated test suite?

getify:
Should mention: I'm using v2.3.0 and node 7.7.4.

platinumazure:
It definitely feels like a bug to me. On the browser, the tests would just hang indefinitely and the run would not be considered a passing run.

I like the idea of using `process.on("exit")` to detect abnormal termination, since we should be able to track if all registered tests have run or not.

Regarding timeouts, I wouldn't want to implement one by default but I would be open to an option for the same. I believe we have per-test timeouts supported in the browser, but I don't know if the CLI supports it yet.

@trentmwillis Any thoughts on this?

getify:
> Regarding timeouts, I wouldn't want to implement one by default but I would be open to an option for the same.

Similar to `assert.expect(..)`, it would be useful if `assert.timeout(..)` could be called to provide a timeout delay, after which the test would be considered a failure.

[...]

PR #1165, opened 2017-04-20
Assert: Introduce timeout to set per-test timeout durations

trentmwillis:
As discussed in https://github.com/qunitjs/qunit/issues/1132.

This makes it easy to set per-test timeouts, instead of the globally configured `QUnit.config.testTimeout`.

### DIFF
--- a/docs/assert/timeout.md
+++ b/docs/assert/timeout.md
@@ -1,0 +1,46 @@
+---
+layout: default
+title: timeout
+description: Sets the length of time to wait for async operations before failing the test.
+categories:
+  - assert
+  - async
+---
+
+## `timeout( duration )`
+
+Sets the length of time to wait for async operations before failing the test.
+
+| name | description |
+|------|-------------|
+| `duration` (Number) | The length of time, in milliseconds, to wait for async operations. |
+
+### Description
+
+`assert.timeout()` sets the length of time, in milliseconds, to wait for async operations in the current test. This is equivalent to setting `config.testTimeout` on a per-test basis. The timeout length only applies when performing async operations.
+
+If `0` is passed, then the test will be assumed to be completely synchronous. If a non-numeric value is passed as an argument, the function will throw an error.
+
+### Examples
+
+```js
+QUnit.test( "Waiting for focus event", function( assert ) {
+  assert.timeout( 1000 ); // Timeout of 1 second
+  var done = assert.async();
+  var input = $( "#test-input" ).focus();
+  setTimeout(function() {
+    assert.equal( document.activeElement, input[0], "Input was focused" );
+    done();
+  });
+});
+```
+
+```js
+QUnit.test( "Waiting for async function", function( assert ) {
+  assert.timeout( 500 ); // Timeout of .5 seconds
+  var promise = asyncFunction();
+  return promise.then( function( result ) {
+    assert.ok( result );
+  } );
+});
+```
--- a/src/assert.js
+++ b/src/assert.js
@@ -14,6 +14,14 @@ class Assert {
 
 	// Assert helpers
 
+	timeout( duration ) {
+		if ( typeof duration !== "number" ) {
+			throw new Error( "You must pass a number as the duration to assert.timeout" );
+		}
+
+		this.test.timeout = duration;
+	}
+
 	// Documents a "step", which is a string value, in a test as a passing assertion
 	step( message ) {
 		const result = !!message;
--- a/src/test.js
+++ b/src/test.js
@@ -35,6 +35,7 @@ export default function Test( settings ) {
 	this.module = config.currentModule;
 	this.stack = sourceFromStacktrace( 3 );
 	this.steps = [];
+	this.timeout = undefined;
 
 	// If a module is skipped, all its tests and the tests of the child suites
 	// should be treated as skipped even if they are defined as `only` or `todo`.
@@ -164,6 +165,15 @@ Test.prototype = {
 		function runTest( test ) {
 			promise = test.callback.call( test.testEnvironment, test.assert );
 			test.resolvePromise( promise );
+
+			// If the test has a "lock" on it, but the timeout is 0, then we push a
+			// failure as the test should be synchronous.
+			if ( test.timeout === 0 && test.semaphore !== 0 ) {
+				pushFailure(
+					"Test did not finish synchronously even though assert.timeout( 0 ) was used.",
+					sourceFromStacktrace( 2 )
+				);
+			}
 		}
 	},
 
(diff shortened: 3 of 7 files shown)


docs/assert/timeout.md
```diff
@@ -0,0 +1,36 @@
...
+
+Sets the length of time to wait for async operations before failing the test.
+
+| name | description |
+|------|-------------|
+| `duration` (Number) | The length of time, in milliseconds, to wait for async operations. |
+
+### Description
+
+`assert.timeout()`sets the length of time, in milliseconds, to wait for async operations in the current test. This is equivalent to setting `config.testTimeout` on a per-test basis. The timeout length only applies when performing async operations.
```
Krinkle: Missing space between "timeout()" and "sets".

src/assert.js
```diff
@@ -14,6 +14,14 @@ class Assert {
 
 	// Assert helpers
 
+	timeout( duration ) {
+		if ( objectType( duration ) !== "number" ) {
```
Krinkle: I don't think we need objectType here. `typeof` should suffices, right?

src/test.js
```diff
@@ -36,6 +36,7 @@ export default function Test( settings ) {
 	this.module = config.currentModule;
 	this.stack = sourceFromStacktrace( 3 );
 	this.steps = [];
+	this.timeout = 0;
```
Krinkle: `config.testTimeout` is undefined by default. Having this one be `0` feels inconsistent. On the other hand, I don't like undefined as default for `config.testTimeout`. `false` is probably a better fit here. `0` could also work although my first-impression expectation based on other software would be that 0 either means infinite waiting or that no waiting is allowed (e.g. any async will be considered an error). Whereas in our case it means no waiting. Although I suppose infinite waiting and no waiting are very similar in the case of qunit - except that issue #1132 exposes a difference, albeit through a bug we should fix separately.

If we go with `0` we should document that for this and for `config.testTimeout` and which ever value we pick, we should file an issue to also change `config.testTimeout`'s default in the next major release.

platinumazure:
@trentmwillis I'm going to wait to see what shakes out of Timo's review before reviewing this myself. Feel free to ping me when you've committed any changes.

src/assert.js
trentmwillis: For 99% of cases it's probably fine, but if someone were to pass a `new Number(1000)`, we'd miss it, even though it is a `number`.

Krinkle: I don't think we should support that. I'd rather have it throw an exception. These are not wild values, they are crafted for the specific purpose of setting a timeout in QUnit.

docs/assert/timeout.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Examples
+
+```js
+QUnit.test( "async test", function( assert ) {
+  assert.timeout( 1000 ); // Timeout of 1 second
+  var done = assert.async();
+  var input = $( "#test-input" ).focus();
+  setTimeout(function() {
+    assert.equal( document.activeElement, input[0], "Input was focused" );
```
gibson042: Could we have a non-DOM example?

src/test.js
gibson042: I agree that `0` is a poor default (it should describe the lowest meaningful timeout value), but I don't mind `undefined` for "no test-specific timeout".

src/test.js
```diff
@@ -656,20 +657,23 @@ export function only( testName, callback ) {
...
 export function internalStop( test ) {
-	var released = false;
-
 	test.semaphore += 1;
 	config.blocking = true;
 
 	// Set a recovery timeout, if so configured.
-	if ( config.testTimeout && defined.setTimeout ) {
+	const timeoutDuration = test.timeout || config.testTimeout;
+	if ( timeoutDuration && defined.setTimeout ) {
```
gibson042: Per the above, this should be more explicit. There's also a question of whether the global value is a fallback or a _maximum_. I'm personally inclined towards the latter, e.g.
```js
const specificTimeout = typeof test.timeout === "number" ? test.timeout : Infinity;
const globalTimeout = typeof config.testTimeout === "number" ? config.testTimeout : Infinity;
const resolvedTimeout = Math.min( explicitTimeout, globalTimeout );
if ( resolvedTimeout < Infinity && defined.setTimeout ) {
	…
}
```

src/test.js
trentmwillis: I didn't put much thought into this when I was implementing it originally. `undefined` seems reasonable and is consistent with the current `testTimeout`, so I will go that route.

src/test.js
trentmwillis: I prefer the fallback approach. It is conceivable to want the majority of your tests to use a short timeout and then override that for a specific, slow test. If we go with the global maximum approach, we force the default for all tests to be the longer value.

trentmwillis:
@Krinkle @gibson042 I've updated and believe I have addressed all comments. Another review would be appreciated 😄

trentmwillis:
@qunitjs/qunit-team can I get someone to review this? I'd like to land this before releasing `2.4`. Thanks!

test/main/assert/timeout.js
```diff
@@ -0,0 +1,61 @@
...
+	// Return a "thenable" to serve as a mock Promise
+	return {
+		then: function() {}
+	};
+} );
+
+QUnit.test( "does not push a failure if test is synchronous", function( assert ) {
+	assert.timeout( 1 );
+
+	var wait = Date.now() + 50;
```
Krinkle: Let's use `10` here as with the other one. Will make the test run a bit faster, too.

trentmwillis: Rebased + updated.

docs/assert/timeout.md
```diff
@@ -0,0 +1,46 @@
...
+
+| name | description |
+|------|-------------|
+| `duration` (Number) | The length of time, in milliseconds, to wait for async operations. |
+
+### Description
+
+`assert.timeout()` sets the length of time, in milliseconds, to wait for async operations in the current test. This is equivalent to setting `config.testTimeout` on a per-test basis. The timeout length only applies when performing async operations.
+
+If `0` is passed, then the test will be assumed to be completely synchronous. If a non-numeric value is passed as an argument, the function will throw an error.
```
trentmwillis: To summarize (to ensure I'm understanding everything):
* `setTimeout(, 0)` doesn't guarantee synchronicity (agreed)
* However, we _do want_ to guarantee synchronicity when using a timeout of `0`
* Therefore, we should special case `assert.timeout(0)`
* And, add tests handling edge cases that likely fail with the current implementation
Does that seem correct?

Krinkle: Yep, that's right.

trentmwillis:
I have updated and addressed @Krinkle's concerns.

@platinumazure I've disabled the `max-len` ESLint rule for the `test` directory since it is not uncommon for test descriptions to be longer than it and I feel it adds negative value to split them on multiple lines.

test/main/assert/timeout.js
```diff
@@ -0,0 +1,101 @@
...
+		} );
+
+		QUnit.test( "fails a test using assert.async and a setTimeout of 0", function( assert ) {
+			assert.timeout( 0 );
+			assert.expect( 1 );
+
+			stubPushFailure( assert );
+
+			var done = assert.async();
+			setTimeout( () => done(), 0 );
```
Krinkle: (Causes CI failure) Too new syntax for the web browser we use for our Travis tests, and presumably for some of our other browsers as well.

trentmwillis: Bah, a side-effect of too much Babel usage...

trentmwillis:
CI passing again.

src/test.js
```diff
@@ -164,6 +165,15 @@ Test.prototype = {
 		function runTest( test ) {
 			promise = test.callback.call( test.testEnvironment, test.assert );
 			test.resolvePromise( promise );
+
+			// If the test has a "lock" on it, but the timeout is 0, then we push a
+			// failure as the test should be synchronous.
+			if ( test.timeout === 0 && test.semaphore !== 0 ) {
```
Krinkle: I keep staring at `test.semaphore !== 0` here, and below at `expect( 1 )` in the failing test cases. I can't quite put my finger on it... Are they needed? What are they for in this case?

platinumazure:
@trentmwillis re: max-len Sounds good to me.

docs/assert/timeout.md
Krinkle: Agree with @gibson042 here. Note: Since the above comment was added, a non-DOM example was also added below, which is great.

I don't particularly mind having an example involving DOM in someway, but as unimportant as it might seem, the specific example we have above seems like an anti-pattern I'd rather not promote in the documentation.

I'd be okay with having just the below example, although in that case I'd rather have the main example showcase `assert.async()` which is easier to understand when learning QUnit. Having two examples, where one shows compat with Promise still would be nice too, though.

PR #1167, opened 2017-04-22
Build: Fix grunt watch

gibson042:
https://gitter.im/qunitjs/qunit?at=58f961fecfec9192725cf2f6
> `grunt watch` will run the tests in node, but I'm pretty sure it won't reload them in any open browser windows (and in fact the webserver aspect appears to be partially broken, I'm seeing a new port claimed every time a watched file changes)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -173,7 +173,7 @@ module.exports = function( grunt ) {
 				"test/node/storage-2"
 			]
 		},
-		watch: {
+		"watch-repeatable": {
 			options: {
 				atBegin: true,
 				spawn: false,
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -173,21 +174,26 @@ module.exports = function( grunt ) {
...
 				"*.js",
 				"build/*.js",
 				"{src,test,reporter}/**/*.js",
 				"src/qunit.css",
+				"test/*.{html,js}",
 				"test/**/*.html"
 			],
-			tasks: "default"
+
+			// TODO insert a new task to prompt reloads between "build" and "test-in-watch"
```
trentmwillis: Just to clarify, once this is done you should be able to `grunt watch` then view the tests at `localhost:8000` where they will livereload, correct?

trentmwillis:
Going to go ahead and merge this assuming my comment above is accurate (as far as I can tell it is).

PR #1168, opened 2017-04-22
Core: Implement only/skip and todo for modules

Arkni:
~~This is a work in progress to implement `QUnit.module.{only, skip, todo}`.
I still need to add some tests and make sure everything is working as expected.~~

Suggestions are more than welcome.

/Cc @trentmwillis

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -17,6 +17,7 @@ import SuiteReport from "./reports/suite";
 import { on, emit } from "./events";
 import onError from "./core/onerror";
 
+let focused = false;
 const QUnit = {};
 export const globalSuite = new SuiteReport();
 
@@ -35,7 +36,7 @@ QUnit.isLocal = !( defined.document && window.location.protocol !== "file:" );
 // Expose the current QUnit version
 QUnit.version = "@VERSION";
 
-function createModule( name, testEnvironment ) {
+function createModule( { name, testEnvironment, todo, skip } ) {
 	const parentModule = moduleStack.length ? moduleStack.slice( -1 )[ 0 ] : null;
 	const moduleName = parentModule !== null ? [ parentModule.name, name ].join( " > " ) : name;
 	const parentSuite = parentModule ? parentModule.suiteReport : globalSuite;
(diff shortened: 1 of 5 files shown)


Arkni:
About tests, should I add a file for each of the methods or there is/are file(s) that I can extend with new unit tests ?

Arkni:
I've rebased this and it is now ready for review.

src/core.js
```diff
@@ -63,53 +71,106 @@ function createModule( name, testEnvironment ) {
...
+		config.currentModule = module;
+		executeNow.call( module.testEnvironment, moduleFns );
+		moduleStack.pop();
+		module = module.parentModule || currentModule;
+	}
+
+	config.currentModule = module;
+}
+
+function module( name, testEnvironment, executeNow ) {
```
trentmwillis: Can you add a `TODO` comment here noting that we should extract this to a new file? (I think it'll require some refactoring, so I don't want it to be in this PR).

src/core.js
```diff
@@ -48,7 +49,14 @@ function createModule( name, testEnvironment ) {
 		testsRun: 0,
 		unskippedTestsRun: 0,
 		childModules: [],
-		suiteReport: new SuiteReport( name, parentSuite )
+		suiteReport: new SuiteReport( name, parentSuite ),
+
+		// Pass along `skip` and `todo` properties from patent module, in case there
```
trentmwillis: Typo: patent -> parent

src/core.js
```diff
@@ -63,53 +71,106 @@ function createModule( name, testEnvironment ) {
...
-			delete module.testEnvironment.before;
-			delete module.testEnvironment.beforeEach;
-			delete module.testEnvironment.afterEach;
-			delete module.testEnvironment.after;
+module.only = function() {
+	if ( focused ) {
+		return;
+	}
+
+	moduleStack.length = 0;
```
trentmwillis: I don't think we should do this. If you focus a module that is nested within another, then the `moduleStack` would still need to be used.

Arkni:
No worries!

I will try to split this PR & address the above comments as soon as possible.

Thanks!

Arkni:
Hi @trentmwillis 

I addressed your requested changes in #1179. The linked PR only contain the implementation of `QUnit.module.only()`. The other two functions will be implemented in subsequent PRs.

Closing this in favor of #1179

PR #1169, opened 2017-04-24
Docs: Document 'todo' in testDone details object

Arkni:


Arkni:
Forgot to add `todo` to the callback definition (just before the second table)
Addressed now!

PR #1171, opened 2017-04-26
Core: Support running in WorkerGlobalScope

### LINKED ISSUE #1022: Support running QUnit in a Web Worker
I'm working on hacking a version of QUnit 2.0 to work in a WebWorker/ServiceWorker, but it would be nice to have official support for it.

martndemus:
This adds support to QUnit for running a test suite inside Web/Service
Worker context.

I have written a test for, but haven't yet added them to any of the test
suites. This is because PhantomJS has a broken Web Worker
implementation. The test runs fine in Chrome/Safari/Firefox and likely
any other browser. Guidance on how to proceed with this would be
appreciated.

Closes #1022

### DIFF
--- a/test/webWorker.html
+++ b/test/webWorker.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Web Worker Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+	<script src="../dist/qunit.js"></script>
+	<script src="webWorker.js"></script>
+</head>
+<body>
+	<div id="qunit"></div>
+</body>
+</html>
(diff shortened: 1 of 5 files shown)


martndemus:
@trentmwillis Basically PhantomJS has a Web Worker implementation since 1.9.8 that doesn't actually work and has never been fixed in any later versions.

martndemus:
See: https://github.com/ariya/phantomjs/issues/13653

trentmwillis:
@qunitjs/qunit-team would appreciate it if one of y'all could review

test/webWorker.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Web Worker Test Suite</title>
+	<link rel="stylesheet" href="../dist/qunit.css">
+	<script src="../dist/qunit.js"></script>
```
Krinkle: I don't think we need to load `qunit.js` here, right?

trentmwillis:
Merged. Thanks, @martndemus!

martndemus:
Yay! Thanks!

PR #1178, opened 2017-05-23
Build: Improve watch and livereload tasks

trentmwillis:
Continues https://github.com/qunitjs/qunit/pull/1167.

After this, you can run `grunt watch` and then visit the tests in a browser at `localhost:8000/test`. Whenever you make a file change, the browser will reload. However, `watch` only restarts after the currently executing task finishes, so if you're on the `qunit` task, it may be a bit slow.

### DIFF
--- a/build/tasks/livereload.js
+++ b/build/tasks/livereload.js
@@ -5,12 +5,13 @@ var server;
 module.exports = function( grunt ) {
 	grunt.registerTask( "livereload", function() {
 		if ( !server ) {
+			var port = this.options().port;
 			server = livereload();
-			server.listen( 35729, function() {
-				console.log( "Livereload server listening on port 35729" );
+			server.listen( port, function() {
+				console.log( "Livereload server listening on port " + port );
 			} );
 		}
 
-		server.changed( { body: { files: [ "qunit.js" ] } } );
+		server.changed( { body: { files: [ "file-path-does-not-matter.js" ] } } );
 	} );
 };
--- a/build/tasks/livereload.js
+++ b/build/tasks/livereload.js
@@ -1,0 +1,16 @@
+/* eslint-env node */
+var livereload = require( "tiny-lr" );
+var server;
+
+module.exports = function( grunt ) {
+	grunt.registerTask( "livereload", function() {
+		if ( !server ) {
+			server = livereload();
+			server.listen( 35729, function() {
+				console.log( "Livereload server listening on port 35729" );
+			} );
+		}
+
+		server.changed( { body: { files: [ "qunit.js" ] } } );
+	} );
+};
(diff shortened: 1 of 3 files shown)


build/tasks/livereload.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-env node */
+var livereload = require( "tiny-lr" );
+var server;
+
+module.exports = function( grunt ) {
+	grunt.registerTask( "livereload", function() {
+		if ( !server ) {
+			server = livereload();
+			server.listen( 35729, function() {
```
gibson042: I left this out initially for lack of a solution that didn't demand a fixed port. Do you have any thoughts on that?

trentmwillis: Only thought is to set it up as an argument that you can pass when running `grunt`. Looks like it should be possible without too much difficulty via https://gruntjs.com/api/grunt.option.

I'll look into it soon and update this PR.

trentmwillis:
@gibson042 updated this to allow customizing the port via `--livereload-port=<port>` when running the grunt command.

build/tasks/livereload.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env node */
+var livereload = require( "tiny-lr" );
+var server;
+
+module.exports = function( grunt ) {
+	grunt.registerTask( "livereload", function() {
+		if ( !server ) {
+			var port = grunt.option( "livereload-port" ) || 35729;
```
gibson042: Wouldn't this make more sense in `Gruntfile.js` configuration?

build/tasks/livereload.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env node */
+var livereload = require( "tiny-lr" );
+var server;
+
+module.exports = function( grunt ) {
+	grunt.registerTask( "livereload", function() {
+		if ( !server ) {
```
gibson042: :+1:

build/tasks/test-on-node.js
```diff
@@ -80,7 +80,10 @@ module.exports = function( grunt ) {
 
 		QUnit.config.autostart = false;
 
-		require( "../../" + file );
+		var filePath = path.resolve( __dirname, "../../" + file + ".js" );
```
gibson042: Nice catch! But I think we want `var filePath = require.resolve( "../../" + file )`, and maybe even a generalized `requireFresh`.

trentmwillis:
@gibson042 I've updated again and addressed all your comments.

trentmwillis:
Just rebased this again. @qunitjs/qunit-team would appreciate a review when one of you has time :)

PR #1179, opened 2017-05-24
Implement QUnit.module.only()

Arkni:
/Cc @trentmwillis

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -17,6 +17,7 @@ import SuiteReport from "./reports/suite";
 import { on, emit } from "./events";
 import onError from "./core/onerror";
 
+let focused = false;
 const QUnit = {};
 export const globalSuite = new SuiteReport();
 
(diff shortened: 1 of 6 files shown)


src/core.js
```diff
@@ -63,53 +64,76 @@ function createModule( name, testEnvironment ) {
...
+
+module.only = function() {
+	if ( focused ) {
+		return;
+	}
+
+	config.modules.length = 0;
+	config.queue.length = 0;
+
+	module.apply( null, arguments );
```
trentmwillis: You should be able to do `module( ...arguments )` instead.

src/core.js
```diff
@@ -63,53 +64,76 @@ function createModule( name, testEnvironment ) {
...
+
+	focused = true;
+};
+
+extend( QUnit, {
+	on,
+
+	module: module,
```
trentmwillis: Should just be `module,` as we can use the ES6 object shorthand.

test/module.js
```diff
@@ -0,0 +1,134 @@
+var tests = {};
```
trentmwillis: We already have a `test/main/module.js`, so I'd prefer this be name `module-only.js` or something like that to avoid them being confused.

test/module.js
```diff
@@ -0,0 +1,134 @@
...
+	failed: 0, skipped: 0, todo: 0
+};
+
+var assertionsStats = {
+	passed: 0, failed: 0
+};
+
+var done = false;
+
+QUnit.testDone( function( details ) {
```
trentmwillis: I don't think we need to collect all this information. We should only collect the test that ran and then any modifiers. So `testId`, `skipped`, or `todo`.

test/module.js
```diff
@@ -0,0 +1,134 @@
...
+
+QUnit.module( "module B", function() {
+	QUnit.module( "This module should not run", function() {
+		QUnit.test( "normal test", function( assert ) {
+			assert.ok( false, "this test should not run" );
+		} );
+	} );
+
+	QUnit.module.only( "Only this module should run", function() {
+		QUnit.todo( "a todo test", function( assert ) {
```
trentmwillis: I think we only need one of each kind of test in here. The extras don't provide any additional value as far as I can tell.

test/module.js
Arkni: I didn't see that file. At first, I thought that given the fact that the new methods are all related to `QUnit.module`, I will put them all in one file and the only name I had at a time was `module`.

I will update it to reflect the implemented method.
Thanks!

src/core.js
Arkni: Good point!

src/core.js
Arkni: will do

Arkni:
@trentmwillis 

Addressed your above comments!

Arkni:
Thanks for reviewing & merging this :)

I'll soon open the other 2 PRs alongside another one for documentation.

PR #1184, opened 2017-06-01
Core: Updates for Node.js version 8 compatibility and testing

binoculars:
Add node 8 to travis and appveyor, change engines list to >=4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,6 @@
   },
   "main": "qunit/qunit.js",
   "engines": {
-    "node": "4.* || 6.* || 7.*"
+    "node": ">=4"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -97,6 +97,6 @@
   },
   "main": "qunit/qunit.js",
   "engines": {
-    "node": "4.* || 6.* || 7.*"
```
stefanpenner: I suspect we don't want to support unsupported versions.

`"node": "4.* || 6.* || 7.* || >= 8"`

note, 5 isn't in that list and shortly (end of this month, 7 will also be dropped)

binoculars: That fields isn't for what you explicitly support, it's what the code is compatible with.

stefanpenner: I suppose its up to interpretation. Either approach is ultimately fine to me (and unblocks me)

PR #1187, opened 2017-06-03
CLI: Clear require cache of watched files between runs

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1182

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -9,6 +9,7 @@ const utils = require( "./utils" );
 const IGNORED_GLOBS = [
 	"**/node_modules/**"
 ];
+const RESTART_DEBOUNCE_LENGTH = 200;
 
 let QUnit;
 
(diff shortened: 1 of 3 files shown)


bin/run.js
```diff
@@ -98,6 +101,12 @@ run.watch = function watch() {
 		ignoreInitial: true
 	} );
 
+	this.watchedFiles = walkSync( process.cwd(), {
```
rwjblue: This should either be done in `run`, the files array should be manually maintained, or we should just loop over `require.cache` when we re-run and remove any entries that start with our path...

I _believe_ that the current implementation here would never clear the require cache of a file added after `qunit --watch` was ran.  

Specifically, this usage scenario:

* run `qunit --watch tests/**/*-test.js`
* wait for first run to complete
* require a non-existent file from a test
* create that file
* change the file in some way
* note that even though the tests are reran, the changes to that file are not picked up

trentmwillis: Good point. I'll add a test for that scenario and update the implementation.

trentmwillis:
Updated with a test for the case described above and added a slight debounce so that multiple file changes can be batched into a single restart.

PR #1188, opened 2017-06-04
Core: Add support for multiple hooks

### LINKED ISSUE #977: Allow multiple before/beforeEach/after/afterEach

trentmwillis:
Fixes #977.

When using the nested module pattern, you can now define a multiple of each hook (`before`/`beforeEach`/`afterEach`/`after`). `before`/`beforeEach` hooks are run FIFO order and `after`/`afterEach` are run in LIFO order. This is consistent with the behavior already exhibited by those hooks when nested in modules.

cc @rwjblue (who I believe is interested in this feature for ember-qunit).

### DIFF
--- a/docs/QUnit/module.md
+++ b/docs/QUnit/module.md
@@ -29,11 +29,11 @@ Group related tests under a single label.
 
 #### Nested module: `nested( hooks )`
 
-A callback with grouped tests and nested modules to run under the current module label.
+A callback which groups tests and nested modules to run under the current module label.
 
 | name | description |
 |-----------|-------------|
-| `hooks` (object) | Runs before the first test. |
+| `hooks` (object) | An object with functions to define `before`/`beforeEach`/`afterEach`/`after` hooks. |
 
 ### Description
 
@@ -199,10 +199,12 @@ QUnit.test( "makes a car", function( assert ) {
 
 ---
 
-Hooks stack on nested modules
+`before`/`beforeEach` hooks queue on nested modules. `after`/`afterEach` hooks stack on nested modules.
 
 ```js
 QUnit.module( "grouped tests argument hooks", function( hooks ) {
+
+  // You can invoke the hooks methods more than once.
   hooks.beforeEach( function( assert ) {
     assert.ok( true, "beforeEach called" );
   } );
--- a/src/core.js
+++ b/src/core.js
@@ -64,29 +64,29 @@ function createModule( name, testEnvironment ) {
 	return module;
 }
 
-function processModule( name, testEnvironment, executeNow ) {
-	let module = createModule( name, testEnvironment );
+function processModule( name, options, executeNow ) {
+	let module = createModule( name, options );
 
 	// Move any hooks to a 'hooks' object
-	if ( module.testEnvironment ) {
-		module.hooks = {
-			before: module.testEnvironment.before,
-			beforeEach: module.testEnvironment.beforeEach,
-			afterEach: module.testEnvironment.afterEach,
-			after: module.testEnvironment.after
-		};
+	const testEnvironment = module.testEnvironment;
+	const hooks = module.hooks = {};
 
-		delete module.testEnvironment.before;
-		delete module.testEnvironment.beforeEach;
-		delete module.testEnvironment.afterEach;
-		delete module.testEnvironment.after;
+	setHookFromEnvironment( hooks, testEnvironment, "before" );
+	setHookFromEnvironment( hooks, testEnvironment, "beforeEach" );
+	setHookFromEnvironment( hooks, testEnvironment, "afterEach" );
+	setHookFromEnvironment( hooks, testEnvironment, "after" );
+
+	function setHookFromEnvironment( hooks, environment, name ) {
+		const potentialHook = environment[ name ];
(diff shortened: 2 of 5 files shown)


src/core.js
```diff
@@ -67,36 +67,36 @@ extend( QUnit, {
...
+		const hooks = module.hooks = {};
+
+		setHookFromEnvironment( hooks, testEnvironment, "before" );
+		setHookFromEnvironment( hooks, testEnvironment, "beforeEach" );
+		setHookFromEnvironment( hooks, testEnvironment, "afterEach" );
+		setHookFromEnvironment( hooks, testEnvironment, "after" );
+
+		function setHookFromEnvironment( hooks, environment, name ) {
+			const potentialHook = environment[ name ];
+			hooks[ name ] = typeof potentialHook === "function" ? [ potentialHook ] : [];
```
rwjblue: I'm unsure if it matters, but this creates 4 new arrays per module regardless of if they are needed.  Instead of falling back to `[]` we could set to `null`, and in `setHookFunction` below you could check for null and create the array when needed.

src/test.js
```diff
@@ -154,9 +154,9 @@ Test.prototype = {
 		checkPollution();
 	},
 
-	queueHook: function( hook, hookName, hookOwner ) {
-		var promise,
-			test = this;
+	queueHook( hook, hookName, hookOwner ) {
+		const test = this;
```
rwjblue: Can we use arrow functions below instead? I am unsure if we transpile though...

src/core.js
trentmwillis: I'm unsure which is better. I went this route so that we don't have to have as much branching in the logic and I doubt it will have much impact on test performance.

src/test.js
trentmwillis: Good suggestion. I can do that, but will need to save it to a variable so that we get a name in the stack trace for better debuggability.

trentmwillis:
Updated to address comments.

> Can I request doc changes in this PR since we have API docs in this repository now?

Yep, I added some comments but nothing too lengthy. The entire `QUnit.module` docs likely need to be reworked as they're kind of confusing as it stands.

> Code/tests LGTM, assuming we have coverage of things like nested module hooks or multiple tests (for beforeEach​/afterEach callbacks) elsewhere.

Yeah, the other tests in the file should handle the other features. We can add additional tests later if bugs are found.

docs/QUnit/module.md
```diff
@@ -199,10 +199,12 @@ QUnit.test( "makes a car", function( assert ) {
 
 ---
 
-Hooks stack on nested modules
+`before`/`beforeEach` hooks queue on nested modules. `after`/`afterEach` hooks stack on nested modules.
```
platinumazure: This makes sense for those familiar with the data structures, but it might not be accessible to software engineers who don't have a strong background in data structures. Maybe this could be reworded to say outer and earlier-declared `before`/`beforeEach` functions run before​ inner and later-declared ones, and the reverse for `after`/`afterEach`? (Though I'll admit what I've proposed isn't great, either.)

trentmwillis: Further up in the document it says:

> Additionally, any hook callbacks on a parent module will wrap the hooks on a nested module. In other words, before and beforeEach callbacks will form a queue while the afterEach and after callbacks will form a stack.

So I'm just trying to keep consistent messaging. If you feel strongly, I can reword it now, but as mentioned before I do think we need to rework the majority of this doc.

platinumazure: No, I just missed that. I agree this needs to be reworked later but also agree that consistent messaging now is good.

trentmwillis:
@qunitjs/qunit-team I'm going to leave this open for another 24 hours in case anyone else wants to review. I know there were some differing opinions on the order of execution for the hooks, so I don't want to jump the gun on merging.

That said, I'm fairly confident in this ordering since it is consistent with the current behavior for nested modules.

src/core.js
```diff
@@ -64,29 +64,29 @@ function createModule( name, testEnvironment ) {
...
-		delete module.testEnvironment.after;
+	const testEnvironment = module.testEnvironment;
+	const hooks = module.hooks = {};
+
+	setHookFromEnvironment( hooks, testEnvironment, "before" );
+	setHookFromEnvironment( hooks, testEnvironment, "beforeEach" );
+	setHookFromEnvironment( hooks, testEnvironment, "afterEach" );
+	setHookFromEnvironment( hooks, testEnvironment, "after" );
+
+	function setHookFromEnvironment( hooks, environment, name ) {
```
Krinkle: Not sure the `hooks` and `environment` parameters` add value in this case. Is this is a V8-optimisation perhaps?

trentmwillis: Unsure what you mean as both parameters are used in the function. Unless you're referring to the function itself, which is really just here to avoid repeating the same code for each hook.

Krinkle: Yeah, but I was suggesting to have `setHookFromEnvironment` take only one parameter, `name`. Since the rest is just here in scope?

PR #1190, opened 2017-06-06
CLI: Better handle various failure scenarios

trentmwillis:
Addressing some of the feedback from https://github.com/qunitjs/qunit/issues/1132.

After this PR, the only remaining item from that issue would be to change the default for `testTimeout` which we won't be able to do until `3.0`.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -12,6 +12,10 @@ const IGNORED_GLOBS = [
 let QUnit;
 
 function run( args, options ) {
+
+	// Default to non-zero exit code to avoid false positives
+	process.exitCode = 1;
+
 	const files = utils.getFilesFromArgs( args );
 
 	QUnit = requireQUnit();
--- a/bin/run.js
+++ b/bin/run.js
@@ -12,9 +12,24 @@ const IGNORED_GLOBS = [
 let QUnit;
 
 function run( args, options ) {
+	let running = true;
 
 	// Default to non-zero exit code to avoid false positives
 	process.exitCode = 1;
+
+	process.on( "exit", function() {
+		if ( running ) {
+			console.error( "Error: Process exited before tests finished running" );
+
+			const currentTest = QUnit.config.current;
+			if ( currentTest && currentTest.semaphore ) {
+				const name = currentTest.testName;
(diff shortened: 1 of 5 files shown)


test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -80,5 +80,11 @@ ok 5 A-Test > derp
 	"qunit --reporter does-not-exist": `No reporter found matching "does-not-exist".
 Available reporters from JS Reporters are: console, tap
 Available custom reporters from dependencies are: npm-reporter
+`,
+
+	/* eslint-disable max-len */
+	"qunit hanging-test": `Error: Process exited before tests finished running
+Last test to run has an async hold. Ensure all assert.async() callbacks are invoked and Promises resolve. You should also set a standard timeout via QUnit.config.testTimeout.
 `
+	/* eslint-enable */
```
platinumazure: This should probably say `/* eslint-enable max-len */`, otherwise any other rules disabled with `/* eslint-disable */` before this line would be re-enabled with this line.

trentmwillis: Good point!

trentmwillis:
@platinumazure updated with your comment addressed.

bin/run.js
```diff
@@ -12,6 +12,23 @@ const IGNORED_GLOBS = [
...
+
+	// Default to non-zero exit code to avoid false positives
+	process.exitCode = 1;
+
+	process.on( "exit", function() {
+		if ( running ) {
+			console.error( "Error: Process exited before tests finished running" );
+
+			if ( QUnit.config.current.semaphore ) {
+				console.error( "Last test to run has an async hold. " +
```
rwjblue: Could we include the test name or other identifiable information? In some circumstances with TAP you aren't aware of the test that is running until it completes, making tracking down this test a bit harder....

trentmwillis: Good suggestion, I'll add the test name.

trentmwillis:
Updated to report the name of the last run test as well. Since the diff from the last review was super tiny (just adding the test name to the error message), I'm going to go ahead and merge.

PR #1193, opened 2017-06-07
Implement QUnit.module.skip()

Arkni:


src/core.js
```diff
@@ -36,7 +36,7 @@ QUnit.isLocal = !( defined.document && window.location.protocol !== "file:" );
 // Expose the current QUnit version
 QUnit.version = "@VERSION";
 
-function createModule( name, testEnvironment ) {
+function createModule( name, testEnvironment, skip ) {
```
trentmwillis: The third argument here should likely be an options hash instead of a single value. That way we don't have to add an additional argument when we do `todo` (and who knows if we'll add something else in the future!)

src/core.js
```diff
@@ -64,8 +70,8 @@ function createModule( name, testEnvironment ) {
 	return module;
 }
 
-function processModule( name, options, executeNow ) {
-	let module = createModule( name, options );
+function processModule( name, options, executeNow, skip = false ) {
```
trentmwillis: Similar to above, the 4th arg should likely be a hash. Maybe `modifiers`?

src/test.js
```diff
@@ -30,13 +30,24 @@ export default function Test( settings ) {
...
 	this.expected = null;
-	extend( this, settings );
 	this.assertions = [];
 	this.semaphore = 0;
 	this.module = config.currentModule;
 	this.stack = sourceFromStacktrace( 3 );
 	this.steps = [];
 
+	// If a module is skipped, all its tests and the tests of the child suites
+	// should be treated as skipped even if they are defined as `only` or `todo`.
```
trentmwillis: I'm unsure if we want to do this for `only` tests. I can see a scenario where a module is skipped, but then they want to test an individual test without re-enabling the whole suite.

I tend to think there needs to be some order of precedence for the modifiers. I think `only > skip > todo > test`. Does that make sense?

src/core.js
Arkni: Initially, I tend to define all arguments as object and use destructuring. Something like this:
```js
function createModule( { name, testEnvironment, skip, todo } ) {
  // ...
}
```

This was my approach when I implemented all 3 methods in the first PR. But I also like your bellow suggestion to use `modifiers` as the third param

src/core.js
Arkni: See above comment https://github.com/qunitjs/qunit/pull/1193#discussion_r122173480

src/test.js
Arkni: >I'm unsure if we want to do this for only tests. I can see a scenario where a module is skipped, but then they want to work on an individual test without re-enabling the whole suite.

Sorry if I misunderstood something, but does that mean the `only` test should be the only one to run?
In that case, no need to skip the whole module, `QUnit` will skip all tests and only run the `only` test.



> I tend to think there needs to be some order of precedence for the modifiers. I think only > skip > todo > test. Does that make sense?

I see it like  this: `skip > only > todo > test`. Saying so, means that even if a `only` test is found inside a skipped module it should be skipped as that what the user expect to happen.

Also, Judging by the order, If a skipped test is inside a `todo` module, it should be left intact, right ?

src/core.js
```diff
@@ -49,7 +49,13 @@ function createModule( name, testEnvironment ) {
...
 		unskippedTestsRun: 0,
 		childModules: [],
-		suiteReport: new SuiteReport( name, parentSuite )
+		suiteReport: new SuiteReport( name, parentSuite ),
+
+		// Pass along `skip` property from parent module, in case there is one,
+		// to childs. And use own otherwise.
+		// This property will be used to mark own tests and tests of child suites
+		// as `skipped`.
+		skip: parentModule !== null && parentModule.skip || skip
```
Arkni: Just crossed my mind right now. What should be the order here ?
Should the modifier of the current module take precedence over the one from the parent module ?

One thing I forgot to test is what will happen if we defined an `only` module inside a `skipped/todo` one (I know this is rarely going to happen), or a `skipped` test inside a `todo` and vise versa ?

Arkni: Right now, what happens is as follow:
* The tests of a normal module that was defined inside a `skipped/todo` will be treated as either `todo` to `skipped` based on the modifier of the parent module.
* If a module was defined as `only`, it will be the only module that will run. But, in case it is inside a `todo` or `skipped` module, its tests will be marked as either `todo` or `skipped` based on the modifier of the parent module.
* If a module was defined as `skipped` and is inside a `todo` module, its tests will be left intact.
* If a module was defined as `todo` and was inside a `skipped` module, its tests will be treated as `skipped` not as `todo`s.

The last two points were based on the order I first defined in my first PR:
```js
if ( this.module.skip ) {
  // ...
} else if ( this.module.todo ) {
  // ....
}
```

src/core.js
Arkni: Changed to `modifiers` per suggestion!

src/core.js
```diff
@@ -49,7 +49,13 @@ function createModule( name, testEnvironment ) {
...
 		unskippedTestsRun: 0,
 		childModules: [],
-		suiteReport: new SuiteReport( name, parentSuite )
+		suiteReport: new SuiteReport( name, parentSuite ),
+
+		// Pass along `skip` property from parent module, in case there is one,
+		// to childs. And use own otherwise.
+		// This property will be used to mark own tests and tests of child suites
+		// as `skipped`.
+		skip: parentModule !== null && parentModule.skip || modifiers.skip
```
Arkni: Just crossed my mind right now. What should be the order here ?
Should the modifier of the current module take precedence over the one from the parent module ?

One thing I forgot to test is what should happen if we defined an `only` module inside a `skipped/todo` one (I know this is rarely going to happen), or a `skipped` test inside a `todo` and vise versa ?

Right now, what I intend to implement is as follow:
* The tests of a normal module that was defined inside a `skipped/todo` will be treated as either `todo` to `skipped` based on the modifier of the parent module.
* If a module was defined as `only`, it will be the only module that will run. But, in case it is inside a `todo` or `skipped` module, its tests will be marked as either `todo` or `skipped` based on the modifier of the parent module.
* If a module was defined as `skipped` and is inside a `todo` module, its tests will be left intact.
* If a module was defined as `todo` and was inside a `skipped` module, its tests will be treated as `skipped` not as `todo`s.

src/test.js
trentmwillis: I tested out the hypothetical scenario with Mocha and I like the way they handle it. Basically, if you have a `skip` module and an `only` test, then the single test is run but it is also skipped, which makes sense. I think we should follow that pattern.

But, I agree that `skip` should take precedence over `todo`.

src/core.js
trentmwillis: I don't think the order here matters. If the module is skipped, then it'll be skipped. If the parent is skipped, it will also be skipped. There is no way to "unskip".

I agree with the rest of your implementation plan :+1:.

src/test.js
```diff
@@ -647,8 +658,11 @@ export function only( testName, callback ) {
 		return;
 	}
 
-	config.queue.length = 0;
-	focused = true;
+	// Only empty the queue in case this test is not part of a skipped module
+	if ( !config.currentModule.skip ) {
```
trentmwillis: Per the above, let's remove this. We should always "focus" the test, even if it will wind up being skipped.

Arkni: Removed it!

I also removed the `only` test because it will be the only one to run (treated as `skip`) which will ignore my real test that check which tests are skipped & which are not. I hope that's acceptable.

trentmwillis: Yeah, we should likely add a test like that as a separate module, but that can happen in another PR.

trentmwillis:
Thanks for working on this!

Arkni:
Happy to help :)

PR #1194, opened 2017-06-10
Docs: Document QUnit.module.{only,skip,todo}

Arkni:


Arkni:
Should I split the docs of every method to its own file or just leave it the way it is now in this PR ?

docs/QUnit/module.md
```diff
@@ -54,6 +54,28 @@ The module's callback is invoked with the test environment as its `this` context
 
 `QUnit.module()`'s hooks can automatically handle the asynchronous resolution of a Promise on your behalf if you return a `then`able Promise as the result of your callback function.
 
+## Exclusive tests
+
+When you are debugging your code, you will often need to _only_ run a subset of tests. You can append `only` to `QUnit.module` to specify which module you want to run.
```
trentmwillis: We should likely note that this is similar to other filtering methods QUnit provides.

docs/QUnit/module.md
```diff
@@ -268,3 +290,103 @@ QUnit.module( "Database connection", {
   }
 } );
 ```
+
+---
+
+Only run a subset of tests:
+
+```js
+QUnit.module( "Tank", function() {
```
trentmwillis: While I appreciate keeping with the theme of "[firin' mah lazer](https://youtu.be/YCj8Yzx15BA)", we should try to avoid "war" imagery. While likely not a big deal for some folks, guns and tanks can be sensitive subjects for others and we want to ensure the documentation we provide is welcoming to all.

Maybe we can have a module for "Android"? That does things like "program" and "actHuman"?

docs/QUnit/module.md
Arkni: Which methods do you mean exactly?
Are your referring to `QUnit.only()` or those config that lets you run a specific module based on its name or its ID: `QUnit.config.module` & `QUnit.config.moduleId` ?

The former is already mentioned bellow that paragraph:
```
It works the same way `QUnit.only()` does for tests.
```

docs/QUnit/module.md
Arkni: Sorry about that.

I'll update it per suggestion.

Thanks for pointing that out!

Arkni: Updated!

docs/QUnit/module.md
trentmwillis: I meant the `config.filer`, `config.module`, etc., but, on second thought, I think this is unnecessary and it is mentioned in the `QUnit.only()` docs, which I think should suffice.

PR #1195, opened 2017-06-10
Implement QUnit.module.todo()

Arkni:


Arkni:
Rebased and fixed the merge conflict.
This is now ready for review!

trentmwillis:
Awesome! Will try to review this evening

src/core.js
```diff
@@ -51,11 +51,12 @@ function createModule( name, testEnvironment, modifiers ) {
...
-		// Pass along `skip` property from parent module, in case there is one,
-		// to childs. And use own otherwise.
+		// Pass along `skip` and `todo` properties from parent module, in case
+		// there is one, to childs. And use own otherwise.
 		// This property will be used to mark own tests and tests of child suites
-		// as `skipped`.
-		skip: parentModule !== null && parentModule.skip || modifiers.skip
+		// as either `skipped` or `todo`.
+		skip: parentModule !== null && parentModule.skip || modifiers.skip,
+		todo: parentModule !== null && parentModule.todo || modifiers.todo
```
trentmwillis: I'm pretty sure that with this setup you could wind up with a module that has both `skip` and `todo`, which seems like it could potentially cause problems. Can we update the logic to ensure only one of them is set to `true`?

Arkni: Yes, you are right, but will have no side effects AFAIK. Because the only place I use those properties are in file `src/test.js#L46-L52`. And in case they are both `true`, only `skip` will then be used.

Anyway, I updated it.

Thanks!

PR #1198, opened 2017-06-14
Tests: Add tests using native async functions.

rwjblue:
Adds tests ensuring that using async functions for callback arguments is under test.

Closes https://github.com/qunitjs/qunit/issues/1197.

Tested locally in Phantom (which can not parse `async function() {}`) and Chrome (which can). Tests pass in both environment, and in Chrome I verified that the new tests are running.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,8 @@
     "npm-reporter": "file:./test/cli/fixtures/npm-reporter",
     "proxyquire": "1.7.11",
     "requirejs": "2.3.2",
-    "rollup-plugin-babel": "2.6.1"
+    "rollup-plugin-babel": "2.6.1",
+    "semver": "5.3.0"
   },
   "scripts": {
     "browserstack": "grunt build && sh build/run-browserstack.sh",
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,16 +2,19 @@
 
 var path = require( "path" );
 var fs = require( "fs-extra" );
+var semver = require( "semver" );
 
 var instrumentedDir = "build/instrumented";
 var reportDir = "build/report";
+
+var HAS_ASYNC_FUNCTIONS = semver.satisfies( process.version, ">= 7.10" );
 
 module.exports = function( grunt ) {
 
 // Load grunt tasks from NPM packages
 	require( "load-grunt-tasks" )( grunt );
 
-	function process( code ) {
+	function preprocess( code ) {
 		return code
 
 		// Embed version
@@ -33,7 +36,7 @@ module.exports = function( grunt ) {
 			}
 		},
 		copy: {
-			options: { process: process },
+			options: { process: preprocess },
 
 			"src-js": {
 				src: "dist/qunit.js",
@@ -174,8 +177,9 @@ module.exports = function( grunt ) {
 				"test/main/dump",
 				"test/node/storage-1",
 				"test/node/storage-2",
-				"test/module-only"
-			]
+				"test/module-only",
+				HAS_ASYNC_FUNCTIONS ? "test/es2017/async-functions" : null
+			].filter( Boolean )
 		},
 		"watch-repeatable": {
 			options: {
--- a/test/es2017/.eslintrc.json
+++ b/test/es2017/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+		"parserOptions": {
+				"ecmaVersion": 2017
+		}
+}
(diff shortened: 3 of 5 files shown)


.eslintignore
```diff
@@ -0,0 +1 @@
+test/async-function-tests.js
```
rwjblue: This was needed because our version of `eslint` cannot parse `async function() {}` properly.

test/index.html
```diff
@@ -4,6 +4,7 @@
 	<meta charset="UTF-8">
 	<title>QUnit Main Test Suite</title>
 	<link rel="stylesheet" href="../dist/qunit.css">
+	<script src="async-function-tests.js"></script>
```
rwjblue: This must be loaded before `../dist/qunit.js`. If it is not, the parse failure on non-async supporting platforms creates a `global failure` test (due to `window.onerror` listener).

.eslintignore
platinumazure: Please create an issue for this and ping me-- ESLint definitely supports parsing of async function and I'm happy to take care of it when I get some free time.

rwjblue: Thank you for bringing this up! I was having a mental block on this. Turns out that I just forgot that I have to setup `ecmaVersion`.  In order to ensure that the rest of the files do not allow all of ES2017, I created a subdir (`test/es2017`) for this one file.

test/index.html
```diff
@@ -4,6 +4,7 @@
 	<meta charset="UTF-8">
 	<title>QUnit Main Test Suite</title>
 	<link rel="stylesheet" href="../dist/qunit.css">
+	<script src="es2017/async-function-tests.js"></script>
```
rwjblue: This must be loaded before `../dist/qunit.js`. If it is not, the parse failure on non-async supporting platforms creates a `global failure` test (due to `window.onerror` listener).

.eslintignore
platinumazure: @rwjblue Thanks. We can also use ESLint's `no-restricted-syntax` to disallow async functions if we don't want to use them anywhere else in the code. (I think async functions and await expressions are the only new syntax in ES2017 unless my memory is failing me.)

rwjblue:
> I'm wary of adding tests which aren't part of our CI (they're easy to neglect and thus don't provide much value).

Hmm, not sure what you mean.  CI is testing these, but the browserstack test run seems to be an "allowed failure". We should probably fix that so we can ensure our browser support guarantees are being met...

> So, I'd like to propose reworking these to run with the Node-based tests and check them in Node 7/8.

Yes, that seems totally fine to me, will update!

rwjblue:
Also, as an aside, it should be fairly straight forward to get Chrome and Firefox tests running in a headless mode (e.g. without BrowserStack) in Travis.

trentmwillis:
Ah, sorry, I'm dumb and didn't realize you had added the tests to the main module suite.

I'm gonna go ahead and land this then. We have an open issue to fix BrowserStack (just haven't had time/desire to get to it): https://github.com/qunitjs/qunit/issues/1139

rwjblue:
@trentmwillis - I started looking at adding these tests to run in node, but it wasn't immediately obvious where they should go.  It seems like all of the files in `test/cli/**/*.js` are related to CLI functionality itself, and doesn't seem to be testing general functionality.  Where would you like me to put them?

Arkni:
@rwjblue 
Just put them anywhere under `test/` and add the files to this list: [`Gruntfile.js#L158-L178`](https://github.com/qunitjs/qunit/blob/bbae0a95ee972cd5ca41b871a23f199a74a36b92/Gruntfile.js#L158-L178)

rwjblue:
@Arkni - Awesome, thank you!

rwjblue:
@trentmwillis - OK, updated to be a bit simpler and just run in Node when we are running under a supported version (v7.10+).

You can see that the new tests are running under v7 [here](https://travis-ci.org/qunitjs/qunit/jobs/243450583#L330), and did not run under v6 [here](https://travis-ci.org/qunitjs/qunit/jobs/243450581#L331).

package.json
```diff
@@ -69,7 +69,8 @@
     "npm-reporter": "file:./test/cli/fixtures/npm-reporter",
     "proxyquire": "1.7.11",
     "requirejs": "2.3.2",
-    "rollup-plugin-babel": "2.6.1"
+    "rollup-plugin-babel": "2.6.1",
+    "semver": "^5.3.0"
```
trentmwillis: We pin our dep versions, so would you mind removing the `^`?

rwjblue:
> Could you also update the yarn.lock?

I added this dep with `yarn add --dev semver`, the `yarn.lock` was not updated because `semver` was already included.

package.json
rwjblue: Updated!

rwjblue:
Updated `semver` dev dep to be pinned to exact version, and confirmed `yarn.lock` was updated (after locking to specific version).  Let me know if there are any other changes needed!

Issue #1197, opened 2017-06-14
Using native async function as hooks does not work


sechel:
Hi, Im using Qunit 2.3.3 and try to use native `async`/`await` syntax in modern browsers such as Chrome 59 or Firefox 55 to perform some asynchronous work in a `beforeEach` hook. However once I declare the `beforeEach` hook to be an `async `function it simply is not invoked anymore, at least not before any of the tests is. Here is a simple example that illustrates this:
```javascript
let testValue;
QUnit.module('async test module', {
  before: async () => {
    testValue = 42
  }
})
QUnit.test( "check value", function( assert ) {
  assert.ok( testValue == 42, "testValue = " + testValue);
});
```
See this [Fiddle](https://jsfiddle.net/19o4poww/1/).
If you replace this by the Promise returning variant as suggested by the documentation all works well:
```typescript
let testValue;
QUnit.module('async test module', {
  before: () => {
     return new Promise((resolve, reject) => {
      testValue = 42;
      resolve();
    });
  }
});
QUnit.test( "check value", function( assert ) {
  assert.ok( testValue == 42, "testValue = " + testValue);
});
```
I wonder what the difference between both examples is from the perspective of the caller? Is there some qualitative difference between an `async` function that ultimately returns a `Promise` and a function that directly returns a `Promise` without being async itself? 
Thanks for sharing insights.

rwjblue:
This was a fun one to poke at (thank you for making the JSFiddle!). The [case statement](https://github.com/qunitjs/qunit/blob/bbae0a95ee972cd5ca41b871a23f199a74a36b92/src/core/utilities.js#L89-L105) in `objectType` utility needs to be taught about `AsyncFunction`.

```js
let fn = async () => {};
console.log(Object.prototype.toString.call(fn));
// => "[object AsyncFunction]"
```

I assume that we want to make `AsyncFunction` return `function` (though I'm not sure why we don't return `typeof obj` in a `default:` clause there), but others should confirm. Also, I am unsure if we have the ability to target specific tests at specific platforms (e.g. a test that uses untranspiled `async function() {}` would crash Phantom/IE/etc).  If we can, then a test can be added to [this file](https://github.com/qunitjs/qunit/blob/master/test/callbacks.js) that uses an async function and adds a logged callback.

trentmwillis:
We should update that function to properly handle `AsyncFunction`. I think defaulting to `typeof obj` makes sense as well.

[...]

PR #1199, opened 2017-06-15
Core: Fallback to `typeof obj` in `QUnit.objectType`.

rwjblue:
Prior to these changes, `QUnit.objectType(async function() { });` would return `undefined` (because `Object.prototype.toString.call(async function() {})` equals `[object AsyncFunction])`).

This updates `QUnit.objectType` to fallback to returning `typeof obj` if no match was found, and adds explicit tests for the various expected inputs and outputs.

I discovered during the investigation of #1197 which is failing in qunitjs@2.3.3 due to `QUnit.objectType` not being aware of async functions (but master was fixed by a refactor done in #1188).

### DIFF
--- a/test/main/utilities.js
+++ b/test/main/utilities.js
@@ -1,0 +1,60 @@
+QUnit.module( "QUnit.objectType" );
+
+function Foo( ) { }
+
+function validateObjectType( item, expected ) {
+	QUnit.test( "should properly detect " + expected, function( assert ) {
+		assert.expect( 1 );
+
+		var actual = QUnit.objectType( item );
+		assert.equal( actual, expected );
+	} );
+}
+
+function maybeValidateObjectType( string, expected ) {
+	try {
+		var value = new Function( "return " + string + ";" )( );
+		validateObjectType( value, expected );
+	} catch ( e ) {
+
+		// do nothing if parsing failed
+		if ( e.name === "SyntaxError" ) {
+			return;
+		}
+
+		// do not hide other errors
+		throw e;
+	}
+}
+
+validateObjectType( 1, "number" );
+validateObjectType( "", "string" );
+validateObjectType( "foo", "string" );
+validateObjectType( null, "null" );
+validateObjectType( true, "boolean" );
+validateObjectType( false, "boolean" );
+validateObjectType( [ ], "array" );
+validateObjectType( new Date( ), "date" );
+validateObjectType( /foo/, "regexp" );
+validateObjectType( NaN, "nan" );
+validateObjectType( undefined, "undefined" );
+validateObjectType( { }, "object" );
+validateObjectType( new Foo( ), "object" );
+
+
+if ( typeof Symbol === "function" ) {
+	validateObjectType( Symbol( "HI!" ), "symbol" );
+}
+
+if ( typeof Map === "function" ) {
+	validateObjectType( new Map( ), "map" );
+}
+
+if ( typeof Set === "function" ) {
+	validateObjectType( new Set( ), "set" );
+}
+
+maybeValidateObjectType( "async function() { }", "function" );
+maybeValidateObjectType( "async () => { }", "function" );
+maybeValidateObjectType( "() => { }", "function" );
+maybeValidateObjectType( "function* { }", "function" );
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -165,6 +165,7 @@ module.exports = function( grunt ) {
 				"test/main/modules",
 				"test/main/deepEqual",
 				"test/main/stack",
+				"test/main/utilities",
 				"test/events",
 				"test/events-in-test",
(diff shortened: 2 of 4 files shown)


test/main/utilities.js
```diff
@@ -0,0 +1,43 @@
...
+
+	try {
+		var arrow = new Function( "return () => {};" )( );
+		objects.push( arrow );
+	} catch ( e ) { }
+
+	try {
+		var async = new Function( "return async function( ) { };" )( );
+		objects.push( async );
+	} catch ( e ) { }
```
leobalter: why this try / catch + wrapping the function forms in the Function ctor?  ES6 compat?

This might escalate in a weird way, as we might also have method definitions, generators, classes, async generators...

test/main/utilities.js
```diff
@@ -0,0 +1,43 @@
...
+			assert.equal( actual, expected );
+		}
+	} );
+}
+
+validateObjectType( [ 1, Number( 1 ) ], "number" );
+validateObjectType( [ "", "foo", String( "bar" ) ], "string" );
+validateObjectType( [ null ], "null" );
+validateObjectType( [ true, false, Boolean( true ) ], "boolean" );
+validateObjectType( [ [], new Array( ) ], "array" );
```
leobalter: the literal and ctor forms of Array are well tested on JS (Test262), same for primitive Number, String and Boolean types.

test/main/utilities.js
```diff
@@ -0,0 +1,43 @@
...
+validateObjectType( [ new Date( ) ], "date" );
+validateObjectType( [ /foo/, new RegExp( "foo" ) ], "regexp" );
+validateObjectType( [ new Date( ) ], "date" );
+validateObjectType( [ NaN ], "nan" );
+validateObjectType( buildFunctions(), "function" );
+validateObjectType( typeof Symbol === "function" ? [ Symbol( "HI!" ) ] : [ ], "symbol" );
+validateObjectType( typeof Map === "function" ? [ new Map( ) ] : [ ], "map" );
+validateObjectType( typeof Set === "function" ? [ new Set( ) ] : [ ], "set" );
```
leobalter: where is the ordinary object? `{}`

test/main/utilities.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+validateObjectType( [ 1, Number( 1 ) ], "number" );
+validateObjectType( [ "", "foo", String( "bar" ) ], "string" );
+validateObjectType( [ null ], "null" );
+validateObjectType( [ true, false, Boolean( true ) ], "boolean" );
+validateObjectType( [ [], new Array( ) ], "array" );
+validateObjectType( [ new Date( ) ], "date" );
+validateObjectType( [ /foo/, new RegExp( "foo" ) ], "regexp" );
+validateObjectType( [ new Date( ) ], "date" );
```
leobalter: duplicate

test/main/utilities.js
leobalter: and undefined

test/main/utilities.js
rwjblue: @leobalter - The same tests are ran on IE, Phantom, possibly Node 6, etc where `async function() { }` would be a parse error. The intent of the `try` / `catch` is to ensure that when `async function() { }` is not valid syntax, the tests for the plain function still runs.

test/main/utilities.js
rwjblue: Awesome! Do you think I should just remove the non-primitive versions here then?

test/main/utilities.js
rwjblue: Good catch! Added.

test/main/utilities.js
rwjblue: Whoops! Thats what I get for copy and pasting 😝

test/main/utilities.js
leobalter: I got it! Thanks for the clarification! 

btw eval allows a shorter syntax, but does the same:
`new Function( "return async function( ) { };" )( )` => `eval( "async function( ) {};" )`

test/main/utilities.js
leobalter: I'm +1 to keep only the primitive and literal values here

test/main/utilities.js
rwjblue: @leobalter - HA! Yes, I started by trying to use `eval`, but I do not think that it allows returning the eval'ed value. This is important for us, because we have to ensure that having the syntax does not throw during evaluation of the whole file (and therefore miss all these tests).

Have I missed something?  Is there a better way?

test/main/utilities.js
rwjblue: Sounds good, updated.

rwjblue:
Updated for feedback, and also updated `Gruntfile.js` so that the tests are also ran in node-land.

PR #1212, opened 2017-07-16
release testEnvironment after a test run

stefanpenner:


stefanpenner:
Actually, I'm not sure this does what I intended. Will investigate further.

PR #1214, opened 2017-07-25
Test: Provide descriptive feedback when missing test callback

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1213

### DIFF
--- a/test/main/test.js
+++ b/test/main/test.js
@@ -119,3 +119,17 @@ QUnit.skip( "test blocks are skipped", function( assert ) {
 } );
 
 QUnit.skip( "no function" );
+
+QUnit.module( "Missing Callbacks" );
+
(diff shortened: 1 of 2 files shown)


test/main/test.js
```diff
@@ -119,3 +119,17 @@ QUnit.skip( "test blocks are skipped", function( assert ) {
 } );
 
 QUnit.skip( "no function" );
+
+QUnit.module( "Missing Callbacks" );
+
+QUnit.test( "QUnit.test without a callback logs a descriptive error", function( assert ) {
+	assert.throws( () => {
```
Arkni: Should be a normal function as `PhantomJS` doesn't support arrow functions. The same goes for the other test.

trentmwillis: Forever trolling myself :( Thanks for looking into the failure! Will fix soon.

PR #1215, opened 2017-07-26
All: Upgraded to eslint@4.x via grunt-eslint@20.x and related plugins

platinumazure:
With eslint@4.x, the indent rule gets a lot smarter and flags a few things it didn't before, including comments. Most of the fallout from this PR is due to that.

Also, I've upgraded:
* eslint-plugin-html to latest (no apparent breaking changes there)
* eslint-plugin-qunit to latest (only breaking change was needing to `eslint-disable-line` a `QUnit.start()` call in a test since the associated rule was added to the config we're using-- that call is for testing the error message so it would have been wrong to remove it).

Once this gets merged in, I'll also add a change (in a separate PR) to forbid ArrowFunctionExpression nodes in our test files, to avoid PhantomJS issues around those.


platinumazure:
Yes! Sorry, I just forgot to do that. I'll push another commit and you can
squash on merge. Thanks for your patience!

test/main/modules.js
```diff
@@ -328,7 +328,7 @@ QUnit.module( "contained suite `this`", function( hooks ) {
 		} );
 	} );
 
-	QUnit.test( "tests can't see environments from nested modules", function( assert )  {
+	QUnit.test( "tests can't see environments from nested modules", function( assert )	{
```
Krinkle: This doesn't look right.

platinumazure:
This should be ready for another round of review.

@trentmwillis Pushed an updated yarn.lock file,

@Krinkle Fixed the line you called out.

PR #1220, opened 2017-08-16
HTML Reporter: Fix regression in error reporting

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1166

Still need to determine the best way to add a test for this.


trentmwillis:
@qunitjs/qunit-team could I get a review on this? Planning to do another patch release soon and this would be good to get in.

Krinkle:
LGTM. Will merge in a day or so if no concerns are raised.

PR #1225, opened 2017-10-12
Build: Add my new name to .mailmap, update it in AUTHORS.txt

mgol:
This also unifies my past contributions under one persona - so far I had entries
with 2 different ways to write "ę" which treated me as 2 distinct people.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -90,3 +90,4 @@ Matthew DuVall <mduvall89@gmail.com>
 Dave K. Smith <dave.k.smith@gmail.com>
 David Vollbracht <david.vollbracht@gmail.com>
 Jochen Ulrich <jochenulrich@t-online.de>
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -80,7 +80,7 @@ Leonardo Balter <leonardo.balter@gmail.com>
 Jeff Cooper <jeff@kickstorming.com>
 Corey Frang <gnarf37@gmail.com>
 Nathan Dauber <nathan@radialogica.com>
-Michał Gołębiowski <m.goleb@gmail.com>
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
 XhmikosR <xhmikosr@yahoo.com>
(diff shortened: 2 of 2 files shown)


mgol:
It's used by `git shortlog` (see e.g. what `git shortlog -nse` prints) & `git blame`. By default it doesn't affect `git log` or `git show`, mainly for historical reasons (see http://git.661346.n2.nabble.com/git-log-to-use-mailmap-by-default-td7570862.html). It can be made so in two ways:

1. By adding a `--use-mailmap` option: `git log --use-mailmap` or `git show COMMIT_HASH --use-mailmap`
2. By configuring Git globally to respect the `.mailmap` for logging: `git config --global log.mailmap true`. When set, this option is also respected for `git show`.

Even without my name change, having a canonical name in `.mailmap` also helps to unify all my contributions under a single person; there are two ways to write "ę" - one as a single symbol and another one as "e" + Unicode "ogonek" (paste this in the console: `["ę".length, "ę".length] // [1, 2]`) and in many repos I find myself written in both ways in some commits, not sure why.

mgol:
`.mailmap` is used to map many different commit authors into a single persona. Having different entries for different commits may be caused e.g. by a name change or using different mails for different commits.

`.mailmap` is used by `git shortlog` (see e.g. what `git shortlog -nse` prints) & `git blame`. By default it doesn't affect `git log` or `git show`, mainly for historical reasons (see http://git.661346.n2.nabble.com/git-log-to-use-mailmap-by-default-td7570862.html). It can be made so in two ways:

1. By adding a `--use-mailmap` option: `git log --use-mailmap` or `git show COMMIT_HASH --use-mailmap`
2. By configuring Git globally to respect the `.mailmap` for logging: `git config --global log.mailmap true`. When set, this option is also respected for `git show`.

Even without my name change, having a canonical name in `.mailmap` also helps to unify all my contributions under a single person; there are two ways to write "ę" - one as a single symbol and another one as "e" + Unicode "ogonek" (paste this in the console: `["ę".length, "ę".length] // [1, 2]`) and in many repos I find myself written in both ways in some commits, not sure why.

trentmwillis:
Thanks for the detailed response! Merging...

PR #1230, opened 2017-10-20
Test: Pause when promise is returned from testEnd callback

### LINKED ISSUE #1228: Provide a public API for manually advancing test execution

* **QUnit version**: *2.4.0*
* **What environment are you running QUnit in? (e.g., browser, Node)**: *both*
* **How are you running QUnit? (e.g., script, testem, Grunt)**: *testem*

## What are you trying to do?

I'm trying to run the tests and pause after each test. Then I need to manually resume the tests.

I've tried using `QUnit.config.blocking = true;` after a test is finished. It works as expected. The tests are paused. But there's no way to tell QUnit to resume the execution of tests. Setting `QUnit.config.blocking` to false doesn't help. Looking at [this code](https://github.com/qunitjs/qunit/blob/1d5df85fad33697371537df366539bbd005baf56/src/core/processing-queue.js#L29-L53) it seems like QUnit stops polling for the queue when `blocking` is set to true.

## What are you proposing?

Expose the `advance` method from `src/core/processing-queue.js` as a public API. Maybe add some checks to make sure `advance` is used correctly (e.g. it can't be called when a test is already running).

mdebbar:
- [ ] Implement the desired behavior
- [ ] Add tests
- [ ] Update the docs

Closes #1228

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -38,6 +38,36 @@ export function emit( eventName, data ) {
 }
 
 /**
+ * Same as `emit` above. If any callbacks return promises, `emitWithPromise` will
+ * return a promise that waits on all of them.
+ *
+ * @private
+ * @method emitWithPromise
+ * @param {String} eventName
+ * @param {Object} data
+ * @return {Void}
+ */
+export function emitWithPromise( eventName, data ) {
+	if ( objectType( eventName ) !== "string" ) {
+		throw new TypeError( "eventName must be a string when emitting an event" );
+	}
+
+	// Clone the callbacks in case one of them registers a new callback
+	const originalCallbacks = LISTENERS[ eventName ];
+	const callbacks = originalCallbacks ? [ ...originalCallbacks ] : [];
+
+	let promises = undefined;
+	for ( let i = 0; i < callbacks.length; i++ ) {
+		const maybePromise = callbacks[ i ]( data );
+		if ( maybePromise && typeof maybePromise.then === "function" ) {
+			promises = promises || [];
+			promises.push( maybePromise );
+		}
+	}
+	return promises && Promise.all( promises );
+}
+
+/**
  * Registers a callback as a listener to the specified event.
  *
  * @public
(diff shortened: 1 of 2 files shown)


src/events.js
```diff
@@ -38,6 +38,36 @@ export function emit( eventName, data ) {
...
+export function emitWithPromise( eventName, data ) {
+	if ( objectType( eventName ) !== "string" ) {
+		throw new TypeError( "eventName must be a string when emitting an event" );
+	}
+
+	// Clone the callbacks in case one of them registers a new callback
+	const originalCallbacks = LISTENERS[ eventName ];
+	const callbacks = originalCallbacks ? [ ...originalCallbacks ] : [];
+
+	let promises = undefined;
```
trentmwillis: I think we can simplify this by doing:

```js
const promises = callbacks.map(callback => Promise.resolve(callbacks[i](data)));
return Promise.all(promises);
```

I think it'll be cleaner if we always treat them as Promises, especially when we account for my other comment.

src/test.js
```diff
@@ -297,7 +297,10 @@ Test.prototype = {
 
 		// After emitting the js-reporters event we cleanup the assertion data to
 		// avoid leaking it. It is not used by the legacy testDone callbacks.
-		emit( "testEnd", this.testReport.end( true ) );
+		const promise = emitWithPromise( "testEnd", this.testReport.end( true ) );
+		if ( promise ) {
+			this.resolvePromise( promise );
```
trentmwillis: In the event we have a Promise, we'll need to block execution of the rest of this block. Otherwise, we're likely to run into ordering issues. For example, the `testEnd` hook may still be running but `QUnit.config.current` may be `undefined` or `suiteEnd` may execute.

rwjblue:
@mdebbar - Had any time to continue pushing this forward?

mdebbar:
@rwjblue got sidetracked by other things. Hopefully will get back to this next week.
@trentmwillis thanks for the feedback! Will iterate on this next week and update the PR.

rwjblue:
@mdebbar - Friendly ping 👋

trentmwillis:
@mdebbar @rwjblue is this still feature still desired?

trentmwillis:
:+1: okay, I'm gonna go ahead and close this and y'all can open a new pr/issue with the new idea.

PR #1232, opened 2017-10-23
All: Rename package to QUnit

trentmwillis:
First step of fixing https://github.com/qunitjs/node-qunit/issues/136.

After merging, I will manually do another 2.4.1 release under the `qunit` package name, so that users can migrate directly from `qunitjs@2.4.1` to `qunit@2.4.1`. Then, we will deprecate `qunitjs@2.4.1` with a message stating that users should switch to the `qunit` package.

All future releases will then be under the `qunit` name.

As for the current `qunit` package, we will be switching that to be released as `node-qunit`. I've reached out to the owner of the current `node-qunit` package (which is deprecated) and he is fine with us taking over the name.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -42,7 +42,7 @@ Follow the prompts along the way and you should be all good. This process will c
 You can verify all of the above were correctly published by using the following:
 
 - Bower: `bower info qunit`
-- NPM: `npm view qunitjs`
+- NPM: `npm view qunit`
 - CDN: visit `https://code.jquery.com/qunit/qunit-x.x.x.js`
 
 Additionally, we publish a new `git` version of QUnit after every change, so we should verify that is still working. You can do so by visiting `https://code.jquery.com/qunit/qunit-git.js` and ensuring the version number matches the latest `-pre` version. Note that it may take a couple minutes for this version to update as the build happens.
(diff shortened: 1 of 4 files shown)


rwjblue:
Seems good to me! I assume there will be another PR to https://github.com/qunitjs/qunitjs.com for the website itself?

trentmwillis:
Yes. I'll have to change those names separately and add a notice about the versions and name compatibility.

trentmwillis:
@qunitjs/qunit-team could I get a review from someone? Want to ensure at least one other maintainer is aware of these changes before proceeding.

trentmwillis:
@platinumazure the version won't change. We have already released `qunitjs@2.4.1` and after merging this I will then publish `qunit@2.4.1`.

We can then deprecate `qunitjs@2.4.1` on npm which will let users know that if they ever want to upgrade past `2.4.1` (completely opt-in) they will need to simply change `qunitjs` to `qunit` in their package.json. I think this is the cleanest upgrade path and saves us from bumping a major version when nothing is really changing in the implementation.

Let me know if you still have concerns.

platinumazure:
Personally if this were my project I would still prefer to do a semver-major bump, but what you have outlined sounds reasonable as long as we have very good documentation on this point. No concerns from my side.

jzaefferer:
Looks good to me. The important point to keep in mind, as we discussed on Gitter: The `qunit` package is currently at `1.0.0`, so the new release `2.4.1` will require a major version bump for anyone currently using the package.

PR #1235, opened 2017-12-12
All: Enable ESLint recommended rules

mc10:
This fixes some minor lint issues.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,7 @@
     "env": {
         "es6": true
     },
-    "extends": "jquery",
+    "extends": ["eslint:recommended", "jquery"],
     "parserOptions": {
         "sourceType": "module"
     },
@@ -18,7 +18,9 @@
 		"SwitchCase": 0,
 		"outerIIFEBody": 1
         }],
+        "no-console": "off",
         "no-const-assign": "error",
+        "no-constant-condition": ["error", { "checkLoops": false }],
         "no-dupe-class-members": "error",
         "no-duplicate-imports": "error",
         "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
--- a/bin/find-reporter.js
+++ b/bin/find-reporter.js
@@ -31,7 +31,7 @@ function findReporter( reporterName ) {
 
 	// If we didn't find a reporter, display the available reporters and exit
 	displayAvailableReporters( reporterName );
-};
+}
 
 function displayAvailableReporters( inputReporterName ) {
 	const message = [];
--- a/bin/run.js
+++ b/bin/run.js
@@ -100,7 +100,7 @@ function run( args, options ) {
 			process.exitCode = 0;
 		}
 	} );
-};
+}
 
 run.restart = function( args ) {
 	clearTimeout( this._restartDebounceTimer );
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -747,7 +747,7 @@ export function escapeText( s ) {
 		// Strip tags, html entity and whitespaces
 		return string
 			.replace( /<\/?[^>]+(>|$)/g, "" )
-			.replace( /\&quot;/g, "" )
+			.replace( /&quot;/g, "" )
 			.replace( /\s+/g, "" );
 	}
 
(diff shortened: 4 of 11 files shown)


mc10:
Looks like test are failing on

https://github.com/mc10/qunit/blob/ed935a1dbe7ba8c05a5a4e4a1f63e7639dc8ed7c/test/main/deepEqual.js#L510-L514

I assume it's the `u` flag that's causing it to fail. Any idea why this might be the case?

platinumazure:
Hi @mc10, thanks for contributing and sorry for the delay.

Regarding the `u` flag, I think that's only supported in fairly recent versions of Node and browsers, so any older test configurations using it might fail. If we absolutely need it, we could discuss whether we can drop support for certain runtimes, transpile around the issue, or if we need to find a different way to handle the use case.

More generally, I want to ask what problem this solving. I'm an ESLint maintainer so I'm aware of eslint: recommended and the benefits it can confer. However, this project uses `eslint-config-jquery` because that configuration is based on the [jQuery style guide](https://contribute.jquery.org/style-guide/js/), which QUnit nominally follows as a jQuery project.

So I guess I want to know why we should consider pulling in these other rules. Is there something objectively lacking in the jQuery guidelines that we should consider for QUnit now? (I think the project lead is generally amenable to the possibility, since we've taken some other steps to go away from jQuery methodology. But it would be good to explain exactly why we should make these changes so he/we can make an informed decision.)

Also, I noticed some other changes embedded in this PR, such as Travis config changes. Could those be split off into one or more separate PRs? Bonus points if they come with a good justification for the change :smile:

mc10:
Sorry for the delay! I've split off the Travis changes into a separate PR. I added the ESLint recommended rules because it didn't seem like [jquery/eslint-config-jquery](https://github.com/jquery/eslint-config-jquery) had all of the useful defaults like unnecessary semicolons, and by including the jQuery rules *after* the recommended rules, the jQuery style changes should still be in effect.

reporter/diff.js
```diff
@@ -907,7 +907,7 @@ QUnit.diff = ( function() {
 		// Performance analysis: https://neil.fraser.name/news/2010/11/04/
 		best = 0;
 		length = 1;
-		while ( true ) {
+		for ( ;; ) {
```
trentmwillis: Which rule is this change accomodating? I think the `while ( true )` form is much easier to read than the `for ( ;; )`

mc10: [no-constant-condition](https://eslint.org/docs/rules/no-constant-condition) disallows the `while (true)` form, but we can allow it by setting `"checkLoops": false`.

mc10:
I'm happy to disable `no-console` globally; would that be better?

platinumazure:
I won't stand in the way of these changes. That said, I'd love to have a discussion later about:

* Whether eslint-config-jquery (which is, nominally, the ESLint config that incorporates as much of the jQuery style guide as it can) should also be augmented with some of these changes, or not; and also,
* Whether the QUnit project should regard itself as following the jQuery style guide, or if we're going to forge our own path.

Basically, if we consider ourselves as still following the jQuery style guide but some of these rules aren't in eslint-config-jquery, we should add them (at some later point); or, if we're not following the jQuery style guide, we should update our contributor guidelines accordingly.

reporter/diff.js
Krinkle: Agreed, we should allow constant expression in loops. I imagine `for ( ;; ) {` might be faster in some sense in some engines, maybe, but not worth the cost of code readability here. If this was a really hot code path that would get faster from this, then it'd be worth considering (but not without a clarifying inline comment in that case).

mc10:
I've disabled `no-console` as it's raising too many false positives without any real benefit. Other than that, this should be good to go.

trentmwillis:
@platinumazure I don't consider us to strictly be following the jQuery style guide. It's a good starting point, but I prefer for us to have flexibility to adopt conventions and standards that fit our project and preferences. We should definitely discuss further though.

I'm going to go ahead and land this. Further adjustments can be made in follow-up PRs if necessary.

Thanks a bunch for the contribution @mc10!

platinumazure:
Thanks for clarifying, @trentmwillis!

PR #1238, opened 2017-12-17
Assert: Add `assert.rejects`.

rwjblue:
Closes https://github.com/qunitjs/qunit/issues/1204

### DIFF
--- a/docs/assert/rejects.md
+++ b/docs/assert/rejects.md
@@ -1,0 +1,83 @@
+---
+layout: default
+title: rejects
+description: Test if the provided promise rejects, and optionally compare the rejection value.
+categories:
+  - assert
+---
+
+## `rejects( promise, expectedMatcher [, message ] )`
+
+Test if the provided promise rejects, and optionally compare the rejection value.
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| `promise` (thenable) | promise to test for rejection      |
+| `expectedMatcher`  | Rejection value matcher              |
+| `message` (string) | A short description of the assertion |
+
+
+### Description
+
+When testing code that is expected to return a rejected promise based on a
+specific set of circumstances, use `assert.rejects()` for testing and
+comparison.
+
+The `expectedMatcher` argument can be:
+
+* A function that returns `true` when the assertion should be considered passing.
+* A base constructor to use ala `rejectionValue instanceof expectedMatcher`
+* A RegExp that matches (or partially matches) `rejectionValue.toString()`
+
+Note: in order to avoid confusion between the `message` and the `expectedMatcher`, the `expectedMatcher` **can not** be a string.
+
+### Example
+
+Assert the correct error message is received for a custom error object.
+
+```js
+QUnit.test( "rejects", function( assert ) {
+
+  assert.rejects(Promise.reject("some error description"));
+
+  assert.rejects(
+    Promise.reject(new Error("some error description")),
+    "rejects with just a message, not using the 'expectedMatcher' argument"
+  );
+
+  assert.rejects(
+    Promise.reject(new Error("some error description")),
+    /description/,
+    "`rejectionValue.toString()` contains `description`"
+  );
+
+  // Using a custom error like object
+  function CustomError( message ) {
+    this.message = message;
+  }
+
+  CustomError.prototype.toString = function() {
+    return this.message;
+  };
+
+  assert.rejects(
+    Promise.reject(new CustomError()),
+    CustomError,
+    "raised error is an instance of CustomError"
+  );
+
+  assert.rejects(
+    Promise.reject(new CustomError("some error description")),
+    new CustomError("some error description"),
+    "raised error instance matches the CustomError instance"
+  );
+
+  assert.rejects(
+    Promise.reject(throw new CustomError("some error description")),
+    function( err ) {
+      return err.toString() === "some error description";
+    },
+    "raised error instance satisfies the callback function"
+  );
+});
+```
--- a/src/assert.js
+++ b/src/assert.js
@@ -305,6 +305,113 @@ class Assert {
 			message
 		} );
 	}
+
+	rejects( promise, expected, message ) {
+		let result = false;
+
+		const currentTest = ( this instanceof Assert && this.test ) || config.current;
+
+		// 'expected' is optional unless doing string comparison
+		if ( objectType( expected ) === "string" ) {
+			if ( message === undefined ) {
+				message = expected;
+				expected = undefined;
+			} else {
+				message = "assert.rejects does not accept a string value for the expected " +
+					"argument.\nUse a non-string object value (e.g. validator function) instead " +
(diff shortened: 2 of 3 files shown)


rwjblue:
@platinumazure - Thanks for reviewing!  You are absolutely right, I was actually somewhat confused about how that worked as I'm used to failing assertions failing CI (which is to be avoided), but a testing framework has to creatively work around that to ensure that its failures are correct too.  I did a bit of digging after your comment to see how this was handled for things like `assert.throws` or `assert.ok`, and figured out how it works (inverting `assert.pushResult`'s `result` value).  

tldr; I just pushed an update with tests around when `assert.rejects` fails.

platinumazure:
I've taken another quick look and it looks promising. I'll give it a full review this week. Naturally, I'll also want @trentmwillis to take a look as well. Thanks for your hard work!

rwjblue:
> looks promising.

Hehe. Thank you, got a good chuckle out of this pun this morning. 😝

src/assert.js
```diff
@@ -305,6 +305,100 @@ class Assert {
...
+
+		const currentTest = ( this instanceof Assert && this.test ) || config.current;
+
+		// 'expected' is optional unless doing string comparison
+		if ( objectType( expected ) === "string" ) {
+			if ( message == null ) {
+				message = expected;
+				expected = null;
+			} else {
+				throw new Error(
```
Krinkle: I may've missed something here, but it seems reasonable to support `expected` as a string, especially given it is documented as being a comparable for the rejection value. The rejection value can be a string, the first documentation example added in this commit also suggests it:

```
Promise.reject("error")
```

Based on the documentation, I would expect that argument types Function and Regex are special-cased, and that other types (e.g. Error instance) are handled by something like propEqual somehow, given how it compares the `name` and `message` properties. I realise the implementation below doesn't use propEqual, and I'm not suggesting it does, but this is what it looks like from the outside, and I'd be nice if string, number and boolean rejection values would work. Alternatively, we'd need to stronger document that the only comparable rejection value we support is an object, and we should also strictly document in that case that that case too is a special case, unlike propEqual or deepEqual, which specifically compares the `constructor`, `name` and `message` properties only.

Krinkle:
I assume we don't want to support `expected` as string because it creates ambiguity with the optional `message` parameter. I don't have a specific suggestion here, but I do think it warrants better documentation than currently provided.

As-is, the documentation tells me the following:
* `expected` is a comparable for the rejection value.
* `Function` and `RegExp` type values are special cases, other types are compared directly to `actual` (such as an Error instance).

However, current code supports `object` as the only "regular" comparable value type. Any other value (except string) silently falls back to `result = false`, that's definitely a problem.

I understand string is a problem because of ambiguity with the optional message parameter. I also understand that we (I think?) discourage throwing things that aren't objects. But in that case we shouldn't showcase it in the documentation. The first usage example is currently `Promise.rejects("error")`.

Two minor suggestions, if we indeed want to discourage throwing non-objects:

* [ ] Remove non-object rejection example from documentation.
* [ ] Make sure that `assert.rejects` explicitly fails when expected is of unsupported type (with unit tests to confirm). The missing `else` case in the current code suggests that at least `number` and `boolean` types would result in silent or less-obvious failure.

docs/assert/rejects.md
```diff
@@ -0,0 +1,77 @@
...
+  function CustomError( message ) {
+    this.message = message;
+  }
+
+  CustomError.prototype.toString = function() {
+    return this.message;
+  };
+
+  assert.rejects(
+    Promise.reject("error"),
```
Krinkle: (*)

rwjblue:
Thanks for the review @Krinkle! I just pushed some changes which (assuming I understood your concerns properly) should address your bullet points.

Specifically:

* cleaned up the documentation a bit to make it clearer what each `expectedMatcher` type actually did
* refactored the implementation around function's for the expected value so that we can properly be aware of the difference between "invalid expected type" vs "matcher function returned non-true"

Let me know what you think!

docs/assert/rejects.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+## `rejects( promise, expected [, message ] )`
+
+Test if the provided promise rejects, and optionally compare the rejection value.
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| `promise` (thenable) | promise to test for rejection      |
+| `expectedMatcher`  | Rejection value matcher              |
```
trentmwillis: This name is inconsistent with the function signature given. I'd prefer we use `expectedMatcher` throughout as it makes more sense semantically.

docs/assert/rejects.md
```diff
@@ -0,0 +1,83 @@
...
+| `message` (string) | A short description of the assertion |
+
+
+### Description
+
+When testing code that is expected to return a rejected promise based on a
+specific set of circumstances, use `assert.rejects()` for testing and
+comparison.
+
+The `expected` argument can be:
```
trentmwillis: `expected` -> `expectedMatcher`

docs/assert/rejects.md
```diff
@@ -0,0 +1,83 @@
...
+Note: in order to avoid confusion between the `message` and the `expectedMatcher`, the `expectedMatcher` **can not** be a string.
+
+### Example
+
+Assert the correct error message is received for a custom error object.
+
+```js
+QUnit.test( "rejects", function( assert ) {
+
+  assert.rejects(new Error("some error description"));
```
trentmwillis: This is an invalid example, right?

docs/assert/rejects.md
```diff
@@ -0,0 +1,83 @@
...
+    "rejects with just a message, not using the 'expected' argument"
+  );
+
+  assert.rejects(
+    Promise.reject(new Error("some error description")),
+    /description/,
+    "`rejectionValue.toString()` contains `description`"
+  );
+
+	// Using a custom error like object
```
trentmwillis: Nit: inconsistent indentation.

src/assert.js
```diff
@@ -305,6 +305,112 @@ class Assert {
...
+
+		const currentTest = ( this instanceof Assert && this.test ) || config.current;
+
+		// 'expected' is optional unless doing string comparison
+		if ( objectType( expected ) === "string" ) {
+			if ( message === undefined ) {
+				message = expected;
+				expected = undefined;
+			} else {
+				message = "rejects does not accept a string value for the expected argument.\n" +
```
trentmwillis: Can we put `assert.rejects() does not accept a string value...`? I think it is just a bit clearer/more helpful.

trentmwillis:
@rwjblue, no pressure, but do you think you'll update this soon? I'd like to see it in the next release if possible but it can always be in the one after as well 🙂

rwjblue:
@trentmwillis - Thanks for the review! I cleaned up the docs and fixed the error message, I believe that addresses all of the outstanding review comments.

trentmwillis:
Thanks a bunch!

PR #1239, opened 2017-12-18
Tests: Update test matrix (drop Node 7, add Node 9)

rwjblue:


rwjblue:
I did this as separate commits, but I'm also happy to squash if that is preferable...

PR #1246, opened 2018-01-05
Core: Remove delay from setTimeouts in processing code

trentmwillis:
Per https://github.com/qunitjs/qunit/issues/1245 these can build up and cause larger test suites to have some significant performance hits. They don't seem to be needed and I can't find any reason in the history as to why they are needed. In fact, the one comment I can find relating to them is a suggestion to [remove it](https://github.com/qunitjs/qunit/pull/653#discussion_r17195204).

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -272,7 +272,7 @@ function scheduleBegin() {
 	if ( defined.setTimeout ) {
 		setTimeout( function() {
 			begin();
-		}, 13 );
+		} );
 	} else {
 		begin();
 	}
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -40,7 +40,7 @@ function advance() {
 
 			config.queue.shift()();
 		} else {
-			setTimeout( advance, 13 );
+			setTimeout( advance );
 			break;
 		}
 	}
--- a/src/test.js
+++ b/src/test.js
@@ -789,7 +789,7 @@ function internalStart( test ) {
 			}
 
 			begin();
-		}, 13 );
+		} );
 	} else {
 		begin();
 	}
--- a/test/main/promise.js
+++ b/test/main/promise.js
@@ -13,7 +13,7 @@ function createMockPromise( assert, reject, value ) {
 				return reject ?
 					rejectedCallback.call( thenable, value ) :
 					fulfilledCallback.call( thenable, value );
-			}, 13 );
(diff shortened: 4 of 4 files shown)


trentmwillis:
@qunitjs/qunit-team would appreciate a review. @Krinkle, I think you've been with the project the longest, do you know of any reason to not remove these?

Turbo87:
Sweet! Nice work!

jzaefferer:
I have a very vague memory of these helping the browser UI getting updated between test or even assertion runs. That might have been needed in 2008... As I apparently commented before, let's get rid of them!

wycats:
This looks great!

Does anyone know the significance of the magic value 13?

trentmwillis:
Thanks for chiming in @jzaefferer. I tend to think this may have had value at some point way back when (even the first commit to this repo had the start function using 13 ms timeout), but since then it has likely just been carried forward.

I'm going to land this after a bit more testing. If it breaks anything after the next release, we can always rollback.

wycats:
I couldn't put this down, so I kept digging. It looks like the magic number 13 came from jQuery, which had it in its initial commit 13 years ago!

https://github.com/jquery/jquery/blame/130b8a1c030ae7888934fcd8dee6695eac271fcb/fx/fx.js

Krinkle:
@trentmwillis Sounds good to me. This shouldn't be needed indeed. The implicit default of ~ 4ms should suffice. I couldn't find any evidence either as for why this would need a larger delay.

Within QUnit, the oldest commit is 9b5085a751ca (8 years ago). @wycats Nice find on tracing it back to jquery-core!

@trentmwillis Just one minor clarification in case there is any doubt, this commit will not remove the async delay. I assume we all know that, but  I'm mentioning this because [the linked comment that suggests removal](https://github.com/qunitjs/qunit/pull/653#discussion_r17195204) mentions "can run immediately" which seems potentially confusing. `setTimeout` always makes the callback run from a future tick of the event loop after the existing call stack and micro tasks have finished. With an additional upfront default and minimum delay of approximately 4ms. No parameter (not even 0) can reduce this delay.

trentmwillis:
Good points, this one, in particular, is important to call out

> ...this commit will not remove the async delay. It just removes the _delay parameter_...

I'll update the commit message to reflect that the delay isn't being removed entirely.

PR #1247, opened 2018-01-08
Docs: Add example for async test with assert.timeout

Krinkle:
Add example showing use of `assert.timeout()` in an async test.

### DIFF
--- a/docs/assert/timeout.md
+++ b/docs/assert/timeout.md
@@ -44,3 +44,12 @@ QUnit.test( "Waiting for async function", function( assert ) {
   } );
 });
 ```
+
+```js
+QUnit.test( "Waiting in an async test", async assert => {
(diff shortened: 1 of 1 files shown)


docs/assert/timeout.md
```diff
@@ -44,3 +44,12 @@ QUnit.test( "Waiting for async function", function( assert ) {
   } );
 });
 ```
+
+```js
+QUnit.test( "Waiting in an async test", async assert => {
+  assert.timeout( 500 ); // Timeout of .5 seconds
+
+  var result = await asyncFunction();
```
trentmwillis: Should we use `let`/`const` instead? I don't particularly care, just seems a bit odd to be using `async`/`await` with `var`.

Krinkle: Thanks, fixed.

PR #1249, opened 2018-01-09
Build: Update yarn.lock

Krinkle:
This should update the version of ESLint used, and thuss expose
the style violation Jenkins caught.

Assuming it does so, I'll update this PR with a fix for the style violation.

Ref https://github.com/qunitjs/qunit/issues/1248.


platinumazure:
Question: Is package.json current? Just the yarn.lock that needed to be updated?

Krinkle:
@platinumazure Yeah, the current version ranges already match. It's just that a minor version was released that contained a bug fix that found a previously unknown violation. (Either that, or it's from an indirect dependency update that the lock file would hold back).

trentmwillis:
Thanks for tracking this down and fixing it!

PR #1250, opened 2018-01-10
HTML Reporter: Ensure attributes on qunit-fixture are reset.

rwjblue:
This change leverages `Node.prototype.cloneNode` to deeply clone the original `qunit-fixture` before any tests have started, then for each test replaces (using `Node.prototype.replaceChild`) the current `qunit-fixture` with a clone of the original.

Fixes https://github.com/qunitjs/qunit/issues/1224.

### DIFF
--- a/runner/fixture.js
+++ b/runner/fixture.js
@@ -20,7 +20,7 @@ import { window, document } from "../src/globals";
 
 		var fixture = document.getElementById( "qunit-fixture" );
 		if ( fixture ) {
-			config.fixture = fixture.innerHTML;
+			config.fixture = fixture.cloneNode( true );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


test/main/test.js
```diff
@@ -11,16 +11,65 @@ if ( typeof document !== "undefined" ) {
 
 	QUnit.module( "fixture", function( hooks ) {
 		var failure = false,
-			values = [ /* initial value (see unshift below), */ "<b>ar</b>", undefined ],
+			values = [
+
+				/* initial value (see unshift below), */
+				/* initial value (see unshift below), */
```
platinumazure: Confused as to why this comment is duplicated.

rwjblue: The comments represent items in the `values` array here, and we now push the original value twice (the two `values.unshift` calls below).

runner/fixture.js
```diff
@@ -33,8 +33,14 @@ import { window, document } from "../src/globals";
 		}
 
 		var fixture = document.getElementById( "qunit-fixture" );
-		if ( fixture ) {
+		var resetFixtureType = typeof config.fixture;
+		if ( resetFixtureType === "string" ) {
+
+			// support user defined values for `config.fixture`
 			fixture.innerHTML = config.fixture;
```
trentmwillis: Seems like we would still want to replace the outerHTML as well, right? Otherwise attributes on `qunit-fixture` won't reset when using user-defined fixture markup.

rwjblue: Great point, I'll add a test case for that and fix it up.

rwjblue: Updated to address (with tests)...

test/main/test.js
platinumazure: Oh, okay, now I understand. (Kind of a weird setup for these tests, though...)

rwjblue: INDEED! It was massively confusing for a bit, I _tried_ to leave them a little easier to grok for next time but I definitely agree it is tricky as heck...

PR #1254, opened 2018-01-22
Core: Throw error if no tests are run

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/569 (finally).

We throw an error if no tests are run. Optionally includes a more specific message if a filter is being used. The error should report as a global failure via the `QUnit.onError` handler.

Need to add some new tests to verify behavior.

### DIFF
--- a/test/cli/fixtures/expected/tap-outputs.js
+++ b/test/cli/fixtures/expected/tap-outputs.js
@@ -124,5 +124,39 @@ not ok 2 global failure
 # skip 0
 # todo 0
 # fail 2
+`,
+
+	"qunit no-tests":
+`TAP version 13
+not ok 1 global failure
+  ---
+  message: "No tests were run."
+  severity: failed
+  actual: null
+  expected: undefined
+  stack: undefined:undefined
+  ...
+1..1
+# pass 0
+# skip 0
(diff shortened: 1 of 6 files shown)


trentmwillis:
@qunitjs/qunit-team this is ready to be review now.

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -124,5 +124,39 @@ not ok 2 global failure
...
+
+	"qunit no-tests":
+`TAP version 13
+not ok 1 global failure
+  ---
+  message: "No tests were run."
+  severity: failed
+  actual: null
+  expected: undefined
+  stack: undefined:undefined
```
Krinkle: The `undefined:undefined` doesn't look too good. Can we avoid that in in a fairly easy way, or something for later?

trentmwillis: I agree but would prefer to address in a follow-up PR. It has to do with the current `QUnit.onError` implementation here https://github.com/qunitjs/qunit/blob/a577183e8d4f975d822e3021cbba552c5ed77549/src/core/onerror.js#L18

PR #1256, opened 2018-01-24
add totalTestToRun to beginData.

stefanpenner:
This allows for test runners, such as testem to differentiate between the total number of tests, and the total that will actually run.

These two numbers are different, due to the fact the modules/tests can be filtered and skipped.

----

Complimentary PR to testem: https://github.com/testem/testem/pull/1208

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -305,6 +305,7 @@ export function begin() {
 		emit( "runStart", globalSuite.start( true ) );
 		runLoggingCallbacks( "begin", {
 			totalTests: Test.count,
+			totalTestsToRun: config.queue.length,
 			modules: modulesLog
 		} );
 	}
--- a/test/logs.js
+++ b/test/logs.js
@@ -1,7 +1,7 @@
 QUnit.config.reorder = false;
 
-var totalTests, moduleContext, moduleDoneContext, testContext, testDoneContext, logContext,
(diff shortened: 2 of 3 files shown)


trentmwillis:
The implementation as it stands will still count "skipped" tests as they get inserted into the queue, but are effectively no-ops.

I'm also unsure I want to add any additional properties to these old reporting callbacks. We'd like folks to move to the new event emitter interface based on the [js-reporters](https://github.com/js-reporters/js-reporters) spec. The `runStart` event should contain all the info you need (and much more) to compute this value.

stefanpenner:
> I'm also unsure I want to add any additional properties to these old reporting callbacks. We'd like folks to move to the new event emitter interface based on the js-reporters spec. The runStart event should contain all the info you need (and much more) to compute this value.

@trentmwillis  I like the new reporter spec. Lets use that.

It's not clear which property I want, is it `SuiteStart`'s` `testCounts.total` ? Suggestions?

trentmwillis:
@stefanpenner there is no explicit property for what's requested, but I think you could figure it out by walking the `tests` and `childSuites` properties of the global suite that gets emitted during `runStart`.

Krinkle:
> I'll close it in a week or so […]

I suppose some weeks last longer than others. `#covid19`

PR #1260, opened 2018-02-01
Extract tests out of queue into its own testQueue

step2yeung:
This PR separates the content of `Qunit.config.queue` which contains both tests (runTest function) and its tasks(each function added to the queue by runTest) into two separate queue: queue and testQueue.

Previously the queue was serving two purposes:
1) tracking the list of tests to run
2) when a given test is ran it unshifts that specific tests tasks to the front of the queue

This PR also goes along the lines of the first comment by `platinumazure` in this PR: https://github.com/qunitjs/qunit/pull/1124
The idea of this change is to clean up and simplify the queue logic to make it easier to reason about. The change separates the processing of tests and its tasks, giving a clear indication when all the tasks of a tests are complete, and when the next test can begin.

Before:
```
QUnit.config.queue = [
  function() { test.before() },
  function runHook(), # before
  function() { test.preserveTestEnvironment(); },
  function runHook(), #beforeEach
  function() { test.run(); },
  function runHook(), #afterEach
  function runHook(), # after
  function() { test.after(); },
  function() { test.finish(); },
  function runTest(),
  function runTest(),
  function runTest(),
  # ...
];
```

After:
```
QUnit.config.queue = [
  function() { test.before() },
  function runHook(), # before
  function() { test.preserveTestEnvironment(); },
  function runHook(), #beforeEach
  function() { test.run(); },
  function runHook(), #afterEach
  function runHook(), # after
  function() { test.after(); },
  function() { test.finish(); },
];

QUnit.config.testQueue = [
  function runTest(),
  function runTest(),
  function runTest(),
  # ...
];
```

### DIFF
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -49,6 +49,7 @@ function advanceTaskQueue() {
 		const elapsedTime = now() - start;
 
 		if ( !defined.setTimeout || config.updateRate <= 0 || elapsedTime < config.updateRate ) {
+
 			// TODO: reintroduce priorityCount decrement to start higher priority test
 			const task = taskQueue.shift();
 			task();
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -38,7 +38,7 @@ function advance() {
  */
 function advanceTaskQueue() {
 	const start = now();
-	config.depth = ( config.depth || 0 ) + 1; //[TODO] Do not know if we still need this
+	config.depth = ( config.depth || 0 ) + 1;
 
 	while ( config.queue.length && !config.blocking ) {
 		const elapsedTime = now() - start;
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -20,6 +20,7 @@ import {
 
 let priorityCount = 0;
 let unitSampler;
+const taskQueue = [];
 
 /**
  * Advances the taskQueue to the next task. If the taskQueue is empty,
@@ -28,7 +29,7 @@ let unitSampler;
 function advance() {
 	advanceTaskQueue();
 
-	if ( !config.queue.length ) {
+	if ( !taskQueue.length ) {
 		advanceTestQueue();
 	}
 }
@@ -40,11 +41,11 @@ function advanceTaskQueue() {
 	const start = now();
 	config.depth = ( config.depth || 0 ) + 1;
 
-	while ( config.queue.length && !config.blocking ) {
+	while ( taskQueue.length && !config.blocking ) {
 		const elapsedTime = now() - start;
 
 		if ( !defined.setTimeout || config.updateRate <= 0 || elapsedTime < config.updateRate ) {
-			const task = config.queue.shift();
+			const task = taskQueue.shift();
 			task();
 		} else {
 			setTimeout( advance );
@@ -59,12 +60,12 @@ function advanceTaskQueue() {
  * Advance the testQueue to the next test to process. Call done() if testQueue completes.
  */
 function advanceTestQueue() {
-	if ( !config.blocking && !config.testQueue.length && config.depth === 0 ) {
+	if ( !config.blocking && !config.queue.length && config.depth === 0 ) {
 		done();
 		return;
 	}
 
-	const testTasks = config.testQueue.shift();
+	const testTasks = config.queue.shift();
 	addToTaskQueue( testTasks() );
 	advance();
 }
@@ -74,7 +75,15 @@ function advanceTestQueue() {
  * @param {Array} tasksArray
  */
 function addToTaskQueue( tasksArray ) {
-	config.queue.push( ...tasksArray );
+	taskQueue.push( ...tasksArray );
+}
+
+/**
+ * Return the number of tasks remaining in the task queue to be processed.
+ * @return {Number}
+ */
+function taskQueueLength() {
+	return taskQueue.length;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/core/processing-queue.js
```diff
@@ -23,67 +22,83 @@ let priorityCount = 0;
...
+ */
+function advanceTaskQueue() {
 	const start = now();
 	config.depth = ( config.depth || 0 ) + 1;
 
 	while ( config.queue.length && !config.blocking ) {
 		const elapsedTime = now() - start;
 
 		if ( !defined.setTimeout || config.updateRate <= 0 || elapsedTime < config.updateRate ) {
-			if ( priorityCount > 0 ) {
```
rwjblue: Was this code unneeded?  What was its original purpose? Can you explain why we don't need it now?

src/core/processing-queue.js
```diff
@@ -23,67 +22,83 @@ let priorityCount = 0;
...
 
-		return;
+/**
+ * Enqueue the tasks for a test into the task queue.
+ * @param {Array} tasksArray
+ */
+function addToTaskQueue( tasksArray ) {
+	for ( let i = 0; i < tasksArray.length; i++ ) {
+		const taskItem = tasksArray[ i ];
+		config.queue.push( taskItem );
```
rwjblue: I wonder if this whole function can be simplified to:

```js
config.queue.push(...tasksArray)
```

What do you think?

src/core/processing-queue.js
step2yeung: In the original code, priorityCount's main purpose is to prioritize tests that was flagged to be executed first. The priorityCount was also incremented/decremented in `advance()` originally because if a tests needed to be inserted to the queue, it will be inserted after the tasks(which gets added to front of the queue) and after the already prioritized tests.
This is not necessary anymore since the task is in it's own queue, so we do not need to deal with the priortyCount when dealing with the tasks. (Only needed for the testQueue in the new implementation)

src/core/processing-queue.js
step2yeung: Yes, it certainly can. Will update that.

rwjblue:
I don’t think the intent is really related to exposing another queue, just trying to simply the queueing logic. 

In the long run, having separate queues for the list of tasks for a single test and the pending tests will make it easier to add more logic around test to test advancing (e.g. so that measurements can be done to prevent memory leaks and/or identify per-test code coverage), but that is definitely not the goal of this PR itself. This PR is only about simplifying the queuing logic (having a single queue where some items are prepended and others are appended is really confusing).

rwjblue:
If you’d prefer to avoid adding a new property on QUnit.config I think we can most likely do that.

@step2yeung - Can you confirm?

step2yeung:
Thanks for clarifying @rwjblue!
We can avoid adding it to QUnit.config, and just have the testQueue be private to processing-queue.js.
If thats the case, then is it even necessary to expose the QUnit.config.queue?

trentmwillis:
Yeah, my concern is primarily around adding a new public property. If we can keep `testQueue` internal, then I'm :+1: on this. We'll still need `QUnit.config.queue` because that is how the CLI/HTML Reporter abort/cancel test runs. We may want to add a proper public API for that behavior, but that should be a separate PR.

step2yeung:
@trentmwillis Thanks for the feedback! I modified it such that `config.queue` contains the list of tests, where as taskQueue is now private to ProcessingQueue.

src/test.js
```diff
@@ -205,7 +205,7 @@ Test.prototype = {
 
 			if ( hookName === "after" &&
 				hookOwner.unskippedTestsRun !== numberOfUnskippedTests( hookOwner ) - 1 &&
-				config.queue.length > 2 ) {
+				( config.queue.length > 0 || ProcessingQueue.taskCount() > 2 ) ) {
```
rwjblue: Not really related to this PR since it was pre-exixting, but this took me a little while to figure out (specifically, why is `> 2` important), it _may_ warrant an inline comment...

src/core/processing-queue.js
```diff
@@ -21,69 +20,91 @@ import {
 
 let priorityCount = 0;
 let unitSampler;
+const taskQueue = [];
```
trentmwillis: It would be nice to have a comment describing what a `task` is within the `taskQueue`. Something to the effect of:

> This is queue of functions that are tasks within a single test. After tests and removed from config.queue they are expanded into a set of tasks in this queue.

src/core/processing-queue.js
```diff
@@ -21,69 +20,91 @@ import {
...
+ */
+function taskQueueLength() {
+	return taskQueue.length;
 }
 
 /**
- * Adds a function to the ProcessingQueue for execution.
- * @param {Function|Array} callback
- * @param {Boolean} priority
+ * Adds a tests to the TestQueue for execution.
```
trentmwillis: Nit: Should be singular, `test`.

src/test.js
trentmwillis: Agreed...should've done that last time I refactored this code

src/core/processing-queue.js
```diff
@@ -21,69 +20,91 @@ import {
...
+ * @param {Array} testTasksArray
```
trentmwillis: I could be wrong, but isn't this always going to be a function now? We pass the `runTest` function into here, and then later `shift` them off and then pass the expanded array to `addToTaskQueue`, right?

src/core/processing-queue.js
```diff
@@ -21,69 +20,91 @@ import {
...
 	const start = now();
 	config.depth = ( config.depth || 0 ) + 1;
 
-	while ( config.queue.length && !config.blocking ) {
+	while ( taskQueue.length && !config.blocking ) {
 		const elapsedTime = now() - start;
 
 		if ( !defined.setTimeout || config.updateRate <= 0 || elapsedTime < config.updateRate ) {
-			if ( priorityCount > 0 ) {
-				priorityCount--;
```
trentmwillis: I believe we need to reintroduce this `priorityCount` decrement somewhere so that if a high priority test is added after the run starts, it should be placed in the right spot. I doubt we have a proper test for this, so we could always add that back with a test in a separate PR.

src/core/processing-queue.js
step2yeung: Correct! I renamed this to `testTasksFunc`

src/core/processing-queue.js
step2yeung: Added a TODO for that.

PR #1261, opened 2018-02-03
Docs: Note that falsy values fail assert.step()

Arkni:


platinumazure:
Hi @Arkni, thanks for contributing!

Would you mind giving me an example of a falsy value that would fail this assertion, but also be commonly understood as passing a message? Thanks!

Arkni:
Maybe it's just me, but I tend to start counting from 0. So, if I do:
```js
QUnit.test("A test", assert => {
  assert.step(0); // (A)

  // ...

  assert.step(1);

  // ...

  assert.step(2);

  // ...

  assert.verifySteps([0, 1, 2]);
});
```

I got an error saying I must provide a message for the step at line (A) which the docs doesn't describe as a failing case.

I know I should use a string value, but the implementation allows other types, e.g. numbers.

platinumazure:
That's good to know. Maybe the fix is to emphasize that steps must be
strings?

@trentmwillis Thoughts?

Arkni:
I think so. `assert.step()` can be updated accordingly to only accept a string value as its message. The error then can change depending on whether:
* The developer provided a non string value (`You must provide a string value to assert.step`), or
* They did provide an empty string or just omited the value. (`You must provide a message to assert.step`)

trentmwillis:
Good thoughts so far! I agree that we should enforce that `assert.step` only accepts string values. My reasoning is this: other than strings, numbers are the only type that really make sense as "steps". But, numbers don't give much semantic meaning and are prone to getting out of order/needing lots of shifting if the test case ever needs to be updated.

So in conclusion, @Arkni I'm on board with changes your propose in your last comment :+1:

Arkni:
@trentmwillis 

Opened #1264 to address it.

I'm closing this PR as it's not needed anymore.

PR #1263, opened 2018-02-03
Rejects docs updates

Arkni:
Hi everyone,

This PR contains the following changes:
1. Marking the second parameter of `assert.rejects()` as optional
2. Adding `error object` as possible values of `expectedMatcher`
3. Removing an unexpected `throw` statement in examples

### DIFF
--- a/docs/assert/rejects.md
+++ b/docs/assert/rejects.md
@@ -26,6 +26,7 @@ comparison.
 The `expectedMatcher` argument can be:
 
 * A function that returns `true` when the assertion should be considered passing.
+* An Error object
 * A base constructor to use ala `rejectionValue instanceof expectedMatcher`
 * A RegExp that matches (or partially matches) `rejectionValue.toString()`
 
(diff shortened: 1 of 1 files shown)


docs/assert/rejects.md
```diff
@@ -26,6 +26,7 @@ comparison.
 The `expectedMatcher` argument can be:
 
 * A function that returns `true` when the assertion should be considered passing.
+* An Error Object (instance)
```
platinumazure: I think to avoid potential confusion, it might be better to lowercase `Object` so nobody thinks it might refer to the global Object constructor. And I'm not sure the parentheses are necessary either.

So something like:

> * An Error object instance

Or:

> * An Error object

What do you think?

Arkni: Good point.

I will update it to `An Error object` as it also include plain objects with `name` and `message` properties (those are still used by some developers out there, e.g. `{name: "AuthenticationError", message: "Wrong username or password"}`)

Arkni:
@platinumazure 

Updated per your suggestion

PR #1264, opened 2018-02-11
Fail test when passing a non string value to assert.step

Arkni:
Ref: https://github.com/qunitjs/qunit/pull/1261#issuecomment-364330296

### DIFF
--- a/docs/assert/step.md
+++ b/docs/assert/step.md
@@ -16,7 +16,7 @@ A marker for progress in a given test.
 
 ### Description
 
-The `step()` assertion registers a passing assertion with a provided message. This makes it easy to check that specific portions of code are being executed, especially in asynchronous test cases and when used with `verifySteps()`. A step will always pass unless a message is not provided.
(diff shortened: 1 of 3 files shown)


Arkni:
@platinumazure 

That would make it readable indeed, but almost all assertion methods (except for `assert.async()`) are reusing `this.pushResult`. And as a result, I didn't want to diverge from that pattern.

platinumazure:
Thanks for that feedback. I would love to move away from that pattern if
possible- there are enough cryptic sections in the source code. So I will
happily work to change all those other places, if that is the direction the
team wants to move in.

Maybe the fairest thing to do would be to get a second opinion. It is in
that spirit of collaboration that I marked my review as a Comment rather
than Request Changes.

Arkni:
I updated the docs of `assert.step()` to note that non-string values also fail the step.

platinumazure:
Works for me. I don't have an objection against this PR itself. Hopefully I can find some time later to fix that anti-pattern.

PR #1267, opened 2018-03-01
 Assert: Clone actual steps array to avoid passing internal reference

trentmwillis:
Fixes https://github.com/qunitjs/qunit/issues/1266

The test setup is a bit awkward to verify this fix, so if anyone has a better idea, let me know!

### DIFF
--- a/test/main/assert/step.js
+++ b/test/main/assert/step.js
@@ -132,3 +132,30 @@ QUnit.test( "errors if not called when `assert.step` is used", function( assert 
 		assert.equal( message, "Expected assert.verifySteps() to be called before end of test after using assert.step(). Unverified steps: one" );
 	};
 } );
+
+// Testing to ensure steps array is not passed by reference: https://github.com/qunitjs/qunit/issues/1266
+QUnit.module( "assert.verifySteps value reference", function() {
+
+	var loggedAssertions = {};
+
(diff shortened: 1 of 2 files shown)


test/main/assert/step.js
```diff
@@ -132,3 +132,30 @@ QUnit.test( "errors if not called when `assert.step` is used", function( assert
 		assert.equal( message, "Expected assert.verifySteps() to be called before end of test after using assert.step(). Unverified steps: one" );
 	};
 } );
+
+// Testing to ensure steps array is not passed by reference: https://github.com/qunitjs/qunit/issues/1266
+QUnit.module( "assert.verifySteps value reference", function() {
+
+	var loggedAssertions = {};
+
+	QUnit.log( function( details ) {
```
rwjblue: Given that (AFAIK) these functions cannot be removed, this `QUnit.log` will continue to fill `loggedAssertions` for all subsequent (and unrelated) tests, right?

Does that pose a concern (or maybe it’s “fine”)?

trentmwillis: It only ever logs entries into `loggedAssertions["verification-assertion"]`, so I don't see much of an issue with it (given that I don't believe any other assertion has that message in our code base)

PR #1269, opened 2018-03-01
Core: Ensure late-add high-priority tests are inserted in proper order

trentmwillis:
As mentioned in this comment: https://github.com/qunitjs/qunit/pull/1260/files#r169779552

(Note: Test has some weird behavior in PhantomJS, but verified it works as expected in Chrome, Safari, and FireFox)

### DIFF
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -49,8 +49,6 @@ function advanceTaskQueue() {
 		const elapsedTime = now() - start;
 
 		if ( !defined.setTimeout || config.updateRate <= 0 || elapsedTime < config.updateRate ) {
-
-			// TODO: reintroduce priorityCount decrement to start higher priority test
 			const task = taskQueue.shift();
 			task();
 		} else {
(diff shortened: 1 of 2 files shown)


Arkni:
The weird behavior in PhantomJS is due to `QUnit.config.reorder` being reset to `false` inside of `grunt-contrib-qunit`'s bridge file , see [`grunt-contrib-qunit/phantomjs/bridge.js#L20-L21`](https://github.com/gruntjs/grunt-contrib-qunit/blob/52914729918b19c447ff3c98407680eaf6a297e4/phantomjs/bridge.js#L20-L21). Those 2 lines were there since the first commit.

trentmwillis:
Thanks for that insight, I would have never figured that out! Will update the comment in the test then.

trentmwillis:
Comment in the test has been updated. Should be good to go now I think.

trentmwillis:
@qunitjs/qunit-team would one of you mind reviewing this for me?

PR #1271, opened 2018-03-02
CLI: Implement --require option

trentmwillis:
This PR implements a `--require` option for the CLI. It allows for any number of dependencies or files to be `require`d prior to the test suite run actually starting.

Fixes https://github.com/qunitjs/qunit/issues/1222

### DIFF
--- a/bin/require-from-cwd.js
+++ b/bin/require-from-cwd.js
@@ -1,0 +1,6 @@
+const resolve = require( "resolve" );
+
+module.exports = function requireFromCWD( mod ) {
+	const resolvedPath = resolve.sync( mod, { basedir: process.cwd() } );
+	return require( resolvedPath );
+};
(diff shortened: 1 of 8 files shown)


bin/require-from-cwd.js
```diff
@@ -0,0 +1,6 @@
+const resolve = require( "resolve" );
```
Krinkle: Not related to this PR, but we seem to have grown a fair number of files in ‘bin/‘ that are not intended as standalone CLI entry points. We should consider moving these to ‘src/‘ instead.

trentmwillis: :+1: agreed, though I would like to ensure we keep some clear separation between CLI code and library code (just means we likely need `src/cli` or something)

Krinkle: Sounds good to me!

jdalton:
This is so cool! This should mean [`esm`](https://www.npmjs.com/package/esm) support can be enabled with `qunit --require esm` ❤️

PR #1279, opened 2018-05-03
Ensure module and test callbacks are released for GC.

rwjblue:
Prior to the changes in this PR, all module and test callbacks are retained (forever). This may not seem significant, but as folks use closure scope to store data across tests (which is very common).

For example, prior to the changes in this PR the following would retain the local variable:

```js
QUnit.module('top', function(hooks) {
  let largeThing;

  hooks.beforeEach(function() {
    largeThing = new LargeThing();
  });

  hooks.afterEach(function() {
    largeThing.destroy();
  });

  test('something that uses largeThing', function(assert) {
    // ...snip...
    largeThing.someMethod();
  });
});
```

In larger test suites, this _really_ adds up... 

---

Based on @krisselden's work in https://github.com/emberjs/ember.js/pull/16609

### DIFF
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -36,6 +36,9 @@ module.exports = function( grunt ) {
 			grunt.log.writeln( "-----" );
 			grunt.log.ok( constructMessage( totals ) );
 
+			// Refresh the QUnit global to be used in other tests
+			global.QUnit = requireFresh( "../../dist/qunit" );
+
 			done( !error );
 		} );
 	} );


test/cli/fixtures/memory-leak/index.js
```diff
@@ -0,0 +1,60 @@
...
+				const retainedFoos = new Function(
+					"foos",
+					"return %GetWeakSetValues(foos, 0)"
+				)( foos );
+
+				assert.equal( retainedFoos.length, 0 );
+			} )
+			.catch( error => {
+				if ( error instanceof SyntaxError ) {
+					console.log( "Must launch node 10 with `--expose-gc --track-retaining-path --allow-natives-syntax`" );
```
krisselden: node >= 9 with `--expose-gc --allow-natives-syntax`

src/test.js
```diff
@@ -342,6 +346,11 @@ Test.prototype = {
 		config.current = undefined;
 
 		function logSuiteEnd( module ) {
+
+			// Reset `module.hooks` to ensure that anything referenced in these hooks
+			// has been released to be garbage collected.
+			module.hooks = {};
```
krisselden: something in the suiteEnd probably was using hooks?

rwjblue: This has been a doozy digging into. I'm still somewhat uncertain, but it seems that the `test-on-node` grunt task is leaking tests (not properly waiting for the `QUnit.done` that is setup in `tests/log.js` which does a `setTimeout` _inside_ the done callback).

I've pushed a small cleanup that prevents the failure, but there is definitely some work needed in this area to clean up the leakage between tests.

krisselden:
This issue trolled us hard https://github.com/qunitjs/qunit/issues/1280

rwjblue:
FWIW, the following roughly polyfills the changes to work on older QUnit versions:

```js
  QUnit.testDone(function() {
    // release the test callback
    QUnit.config.current.callback = undefined;
  });

  QUnit.moduleDone(function() {
    // release the module hooks
    QUnit.config.current.module.hooks = {};
  });
```

Turbo87:
@trentmwillis friendly reminder :)

trentmwillis:
I just released 2.6.1. Sorry for the delay!

Turbo87:
Awesome, thanks! :)

src/test.js
Krinkle: I believe that as of this commit, when QUnit is tries to report the "No tests run" error, it is unable to do so now, because:

```js
  			if (module.hooks[handler].length) {
  			                          ^

TypeError: Cannot read property 'length' of undefined
    at processHooks
```

I ran into this as part of reproducing https://github.com/qunitjs/qunit/issues/1405 locally.

PR #1283, opened 2018-06-06
[fixes #121] chokidar -> sane

### LINKED ISSUE #121: Evented/Timed test helpers
# Evented testing

_DISCLAIMER:_  most of the ideas here may be project specific but appear, at least from where I'm standing, to be useful in a myriad of scenarios when testing event based functionality.

Timing and event based tests in QUnit and other frameworks tend to be fragile and even dependent on execution order in practice. There are few test helpers that we've found to be invaluable in working with events and time dependent functionality.
## sequence( functions, timeout )

Many tests that require a set of timeouts to verify functionality can be cleaned up by abstracting multiple calls to setTimeout into a single helper. This is just as brittle as using the timeouts themselves but at least results in cleaner code where no event is provided to guarantee the behavior.

Example:

https://github.com/jquery/jquery-mobile/blob/master/tests/unit/navigation/navigation_core.js#L49
## eventSequence( event, functions )

The event sequence helper executes a list of functions based on a series of the same event firing to trigger each. The most obvious use case where jQuery Mobile is concerned is page transitions which fire the `changepage` event upon completion.

An example from the event tests: 

https://github.com/jquery/jquery-mobile/blob/master/tests/unit/event/event_core.js#L425

An example from the navigation tests. `pageSequence` is just a wrapper around `eventSequence` that specifies `changepage` event as the first argument to `eventSequence`:

https://github.com/jquery/jquery-mobile/blob/master/tests/unit/navigation/navigation_core.js#L300
## eventCascade( event, function, event, function ... )

We haven't stumbled upon a use case here yet but it seems likely that a more flexible form of the eventSequence function would be valuable. In fact this should encourage a more pluggable design since events will be the preferred way to guarantee states of execution in the code being tested!
## waitFor

Jasmine has a nice helper in the form of `waitFor`. It holds the test suite execution until the callback returns a truthy value. Mind you this is dependent on the serialization of all the code in the tests through the `run` function. More information can be found at their wiki down on the bottom:

https://github.com/pivotal/jasmine/wiki/Asynchronous-specs

stefanpenner:
I think chokidar is great, but the ergonomics around native dependencies today in node really painful. Although chokidar pre-builds, users switching nodes, or with old yarn.lock files, or failed pre-build downloads, or behind firewall usages cause a fallback to requiring a working compiler toolchain. Requiring that, very much increases the barrier of entry, and support load for tooling/cli maintainers.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -151,18 +151,18 @@ function watcherEvent( event, args ) {
 }
 
 run.watch = function watch() {
-	const chokidar = require( "chokidar" );
+	const sane = require( "sane" );
 	const args = Array.prototype.slice.call( arguments );
 
-	const watcher = chokidar.watch( "**/*.js", {
-		ignored: IGNORED_GLOBS,
-		ignoreInitial: true
+	const watcher = sane( process.cwd(), {
+		globs: [ "**/*.js" ],
+		ignored: IGNORED_GLOBS
 	} );
 
 	watcher.on( "ready", () => run.apply( null, args ) );
 	watcher.on( "change", watcherEvent( "changed", args ) );
 	watcher.on( "add", watcherEvent( "added", args ) );
-	watcher.on( "unlink", watcherEvent( "removed", args ) );
+	watcher.on( "delete", watcherEvent( "removed", args ) );
 
 	function stop() {
 		console.log( "Stopping QUnit..." );
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6592 @@
+{
+  "name": "qunit",
+  "version": "2.6.2-pre",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
+      "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
+      "dev": true,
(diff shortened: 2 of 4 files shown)


stefanpenner:
I'll try to land: https://github.com/amasad/sane/issues/121 to make this slightly more compelling.

trentmwillis:
> ...cause a fallback to requiring a working compiler toolchain. Requiring that, very much increases the barrier of entry, and support load for tooling/cli maintainers.

What does this mean practically? Can you provide a concrete example of how this might increase the barrier of entry?

stefanpenner:
> What does this mean practically? Can you provide a concrete example of how this might increase the barrier of entry?

prebuilt dependencies work only if:

* access to AWS
* prebuilt for your OS + Node ABI is available on AWS
* you don't switch node versions.

If any of the above fall sideways, you end up requiring:

* python / node-gyp + compiler toolchain (like gcc) working

This affects the following users:

* unfamiliar with node, especially those on windows
* with firewall rules (work rules) preventing aws cdn access
* power-users which switch node versions often
* people doing trainings, where their users have "issues with their environment". Pure JS deps by default, avoid a whole host of these issues.

stefanpenner:
> Looks like all the --watch tests are currently failing in CI though. Once those are fixed I'm good to land this.

👍

stefanpenner:
Sorry about the delay, just fixed the tests locally. Let's see what CI says...

bin/run.js
```diff
@@ -151,18 +151,18 @@ function watcherEvent( event, args ) {
...
+	const watcher = sane( process.cwd(), {
+		globs: [ "**/*.js" ],
+		ignored: IGNORED_GLOBS
 	} );
 
 	watcher.on( "ready", () => run.apply( null, args ) );
 	watcher.on( "change", watcherEvent( "changed", args ) );
 	watcher.on( "add", watcherEvent( "added", args ) );
-	watcher.on( "unlink", watcherEvent( "removed", args ) );
+	watcher.on( "delete", watcherEvent( "removed", args ) );
```
trentmwillis: On a personal note, I prefer this event name so much more than "unlink".

stefanpenner:
@trentmwillis is there a release planned soon, or will it be some time? (I am unfamiliar with the QUnit release cycles)

trentmwillis:
We don't really have an official release cycle. It's pretty much after we've had some worthwhile changes and/or someone asks for it. I'll plan to do one this week.

PR #1287, opened 2018-06-15
Core: Extract QUnit.module to its own file

Arkni:
As requested in https://github.com/qunitjs/qunit/pull/1168#discussion_r117645535


Krinkle:
@Arkni Thanks!

PR #1293, opened 2018-06-18
Docs: Fix Jekyll build (remove conflicting file/dir names)

Krinkle:
Follows-up a7ccfcb.

Having an entry for both "foo" and "foo/" is supported, however
a problem happens if the "foo" part contains one or more dots.

Normally, "foo" creates `_site/foo.html` and
"foo/" creates `_site/foo/index.html`. However, if the name
contains a dot, then the file name won't have ".html" appended.

This causes the Jekyll build to fail because it cannot, for example
create "_site/QUnit.module" as both a file and a directory.

```
jekyll 3.4.3 | Error:
> File exists @ dir_s_mkdir - qunit/docs/_site/QUnit.module
```

Rather than trying to find a way to enforce `.html` in the filename,
I found out that the Jekyll plugin actually handles redirects for
slashless variants as well, so we only need the versions with a slash
and everything will continue to work fine.

### DIFF
--- a/docs/QUnit/module.md
+++ b/docs/QUnit/module.md
@@ -5,9 +5,7 @@ description: Group related tests under a single label.
 categories:
   - main
 redirect_from:
-  - "/QUnit.module"
   - "/QUnit.module/"
-  - "/module"
   - "/module/"
 ---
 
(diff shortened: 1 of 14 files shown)


Krinkle:
Oops! I did test locally, but after I added a few more entries and I didn't confirm that those build fine. Immediately after merging, got an automated e-mail from GitHub Pages saying the build failed, and I can reproduce it locally. This should fix it (confirmed locally, at least).

PR #1295, opened 2018-06-29
Update `fsevents` subdependency for Node 10 compat

Turbo87:
without this update `node-gyp` failed when running on Node 10.3.0

/cc @trentmwillis

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -35,13 +35,6 @@ ajv-keywords@^2.1.0:
   version "2.1.1"
   resolved "https://registry.yarnpkg.com/ajv-keywords/-/ajv-keywords-2.1.1.tgz#617997fc5f60576894c435f940d819e135b80762"
 
-ajv@^4.9.1:
-  version "4.11.8"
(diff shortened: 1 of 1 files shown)


platinumazure:
Should there also be a change in package.json?

rwjblue:
No, the change is already allowed by the current package.json range. But when you clone the repo and `yarn` it always gives you the previously locked down version.

platinumazure:
@rwjblue Understood, but should we set the package.json to a higher version anyway so our dependency list avoids the problematic version even if yarn.lock were removed?

Turbo87:
@platinumazure `fsevents` is just a subdependency of some other dependency so there is no change in the `package.json` file. If you don't use a lockfile you would have gotten that update automatically, but since the lockfile also locks subdependencies they sometimes need to be updated too.

Turbo87:
whoops, missed that @rwjblue had already replied.

`fsevents` is brought in by `chokidar`, and yes and update of that dependency would be good, but since their current version is 2.0.4 which has a major bump compared to the 1.7.0 we use, this is a slightly larger undertaking to upgrade.

platinumazure:
Thanks, that answers my questions.

PR #1298, opened 2018-07-06
pushResult() doesn't ever return a value: it's always undefined/void.

GerHobbelt:
This is an extract of other work and the result of a code review. Check the implementation of pushResult and all its other spots of usage: no return value is ever produced.


GerHobbelt:
CLA signed

trentmwillis:
Thanks! Sorry for the delay in merging.

PR #1299, opened 2018-07-31
CLI: Sane Bump

stefanpenner:
* modernized code-base (node 6+)
* no other changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "findup-sync": "2.0.0",
     "js-reporters": "1.2.1",
     "resolve": "1.5.0",
-    "sane": "^2.5.2",
+    "sane": "^3.0.0",
     "walk-sync": "0.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


trentmwillis:
Can you regenerate yarn.lock? Assuming tests pass after that I'll merge.

stefanpenner:
@trentmwillis oops, yup fixing

platinumazure:
@trentmwillis Is this a semver-major change (if we're dropping support for older Node versions)? Or would this be handled by our Rollup config?

trentmwillis:
@platinumazure I believe we stop supporting Node versions when they leave LTS without necessarily doing a major version bump. This is a "rolling" policy similar to what we do for browsers. Here's the original discussion for more context https://github.com/qunitjs/qunit/pull/1025#issuecomment-236702128.

We wanted to go this route so that we could drop legacy versions of Node without having to do a major version bump every time.

Let me know if you see any issue with that. We've already dropped Node 4 from our Travis config.

platinumazure:
No issue here, forgot about that discussion. In some other projects I'm on, this would be a major version bump. Thanks!

trentmwillis:
Landing this and it'll go out in the next release (which I plan to be a minor 2.7.0)

Issue #1302, opened 2018-08-07
Update development dependencies for CVE-2018-3728


Krinkle:
GitHub's vulnerability detection flags the following

> Dependencies defined in package-lock.json (556)
> * hapijs / hoek [ Known security vulnerability in 2.16.3 ]

See also:
* <https://nvd.nist.gov/vuln/detail/CVE-2018-3728>
* <https://www.npmjs.com/package/hoek>
* <https://github.com/hapijs/hoek>

Krinkle:
Tracing package-lock.json, all uses of hoek specify `"hoek": "2.x.x",`. Including `boom`, `hawk`, `sntp`. Those in turn are used by `request`. Which are is used by `coveralls` (request 2.79.0), `phantomjs-prebuilt` (request 2.81.0). Which are dependencies of `grunt-coveralls` and `grunt-contrib-qunit`.

Which we specify in [package.json#devDependencies](https://github.com/qunitjs/qunit/blob/9af04486aaf1bf23b602a8f379a7826989a9c89e/package.json#L42).

Krinkle:
I'm drafting a change to upgrade `grunt-contrib-qunit`, but ran into https://github.com/gruntjs/grunt-contrib-qunit/issues/149.

Krinkle:
@trentmwillis Thanks, I'll continue on `grunt-contrib-qunit`. Could you look into `grunt-coveralls`?

trentmwillis:
Sure, I'll take a look this weekend.

PR #1303, opened 2018-08-13
Build: Upgrade grunt-contrib-qunit to 3.x (Headless Chrome)

Krinkle:
Ref #1302.

### DIFF
--- a/build/coverage-bridge.js
+++ b/build/coverage-bridge.js
@@ -1,11 +1,16 @@
-/* global QUnit, window */
+/* global QUnit, window, define, require */
 
-( function( QUnit ) {
+( function( factory ) {
 
-	// Send messages to the parent PhantomJS process via alert! Good times!!
+	// For the amd tests we need to delay setup
+	if ( typeof define === "function" && define.amd ) {
+		require( [ "qunit" ], factory );
+	} else {
+		factory( QUnit );
+	}
+}( function( QUnit ) {
 	function sendMessage() {
-		var args = [].slice.call( arguments );
-		window.alert( JSON.stringify( args ) );
+		window.__grunt_contrib_qunit__.apply( window, arguments );
 	}
 
 	QUnit.done( function() {
@@ -15,4 +20,4 @@
 			sendMessage( "qunit.coverage", window.location.pathname, window.__coverage__ );
 		}
 	} );
-}( QUnit ) );
+} ) );
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,7 +28,13 @@ module.exports = function( grunt ) {
 	grunt.initConfig( {
 		pkg: grunt.file.readJSON( "package.json" ),
 		connect: {
-			server: {
+			nolivereload: {
+				options: {
+					port: 8000,
+					base: "."
+				}
+			},
+			livereload: {
 				options: {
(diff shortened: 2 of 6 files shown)


Krinkle:
The tests pass, except for `test/amd.html`. I'm not sure what's going on there.

Arkni:
@Krinkle 

The local bridge file used for collecting coverage should also be changed to use `[self|window].__grunt_contrib_qunit__` instead of `window.alert`:
https://github.com/qunitjs/qunit/blob/9af04486aaf1bf23b602a8f379a7826989a9c89e/build/coverage-bridge.js#L5-L9

See: https://github.com/gruntjs/grunt-contrib-qunit/blob/a4be5b5b41bc8df550cf3668be9b973a200deae9/chrome/bridge.js#L23-L25

Krinkle:
@Arkni Thank for catching that!

platinumazure:
The code here looks good, but I need to see what was done in grunt-contrib-qunit as well before I can sign off. Hopefully I'll have time to do that this week.

build/coverage-bridge.js
```diff
@@ -4,8 +4,7 @@
 
 	// Send messages to the parent PhantomJS process via alert! Good times!!
```
platinumazure: This comment seems inaccurate. Should it be removed?

Arkni:
Hi @Krinkle,

I think the amd.html test fails because the coverage bridge was not designed to work with amd just like the built-in bridge.

Also, after spending some time navigating the new code of `grunt-contrib-qunit` in the hope of making it more functional as it was with `PhantomJS` or even better, I found that it reports wrong stats for QUnit tests:
* Before upgrading:
```
287 tests completed with 0 failed, 15 skipped, and 5 todo. 
830 assertions (in 2414ms), passed: 825, failed: 5
```
* After:
```
324 tests completed with 2 failed, 18 skipped, and 9 todo. 
922 assertions (in 2414ms), passed: 907, failed: 15
```

The two failed tests are:
1. The test `Unhandled Rejections inside test context - test passes just fine, but has a rejected promise` in `/test/reporter-html/unhandled-rejection.js`. And
2. The test `error is thrown when overloading QUnit global` in `/test/overload.html` as Chrome reports the generic error `Script error.` instead of the raised error `QUnit has already been defined.` due to this rule https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#Notes but this only happens when using the file URI scheme (works fine in Firefox and PhantomJS).

For now, I wouldn't recommend landing this PR untill I or someone else found the time to fix `grunt-contrib-qunit` (it was my fault I didn't review the migration PR, but trying my best to fix it).

trentmwillis:
Thanks all for working on this! Just wanted to call out that this would also fix https://github.com/qunitjs/qunit/issues/1172.

trentmwillis:
I've rebased this branch and fixed the test failures; the AMD test failures were kind of tricky but basically we tried to set up the bridge before the QUnit global was available. Stole a solution from the `grunt-contrib-qunit` lib: https://github.com/gruntjs/grunt-contrib-qunit/blob/a4be5b5b41bc8df550cf3668be9b973a200deae9/chrome/bridge.js#L10-L16

PR #1307, opened 2018-08-22
Core: Adding promise handling on qunit callbacks: begin, moduleStart, testStart, testDone, moduleDone

step2yeung:
Addresses #1228 

This change consist of several parts to enable QUnit callbacks to handle promises like:
```
  QUnit.moduleDone(function() {
    return new Promise(function(resolve) {
      // do some async things
      resolve();
    });
  });
```

1. Make advanceTaskQueue handle promises returned from each task
Each QUnit Test consist of multiple tasks, which handles setup, test execution, teardown and reporting of test results.
"advanceTaskQueue" is the function that controls the execution flow of tasks from each test. Hence, this needs to be able to handle promises that are returned by tasks.

2. 'runLoggingCallbacks' returning a promise
'runLoggingCallbacks' is the function used to execute callback (eg. begin, moduleStart, testStart, testDone, moduleDone, done) function that is registered. This will now return a promise of the callbacks.

Note: since we are returning Promise.all(callbackPromises), it does not resolve promises in sequential order. Open to make it handle sequentially.

3. Update 'runLoggingCallbacks' usage to handle promises.
Since 'runLoggingCallbacks' returns a promise, `begin()` and `finish()` will have to return that promise chain. The promise will end up being waited on by 'advanceTaskQueue', and will move forward when the promise is resolved.

Note2: Have not enabled this for QUnit.on(), QUnit.log() and QUnit.done(). 

TODOs:
- [ ] Update Docs

### DIFF
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -34,10 +34,7 @@ export function registerLoggingCallbacks( obj ) {
 }
 
 export function runLoggingCallbacks( key, args ) {
-	var i, l, callbacks;
-
-	callbacks = config.callbacks[ key ];
-	for ( i = 0, l = callbacks.length; i < l; i++ ) {
-		callbacks[ i ]( args );
-	}
+	var callbacks = config.callbacks[ key ];
+	var promises = callbacks.map( callback => Promise.resolve( callback( args ) ) );
+	return Promise.all( promises );
 }
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -1,5 +1,6 @@
 import config from "./config";
 import { objectType } from "./utilities";
+import Promise from "../promise";
 
 // Register logging callbacks
 export function registerLoggingCallbacks( obj ) {
--- a/src/core/logging.js
+++ b/src/core/logging.js
@@ -35,6 +35,20 @@ export function registerLoggingCallbacks( obj ) {
 
 export function runLoggingCallbacks( key, args ) {
 	var callbacks = config.callbacks[ key ];
-	var promises = callbacks.map( callback => Promise.resolve( callback( args ) ) );
-	return Promise.all( promises );
+
+	// Handling 'log' callbacks separately. Unlike the other callbacks,
+	// the log callback is not controlled by the processing queue,
+	// but rather used by asserts. Hence to promisfy the 'log' callback
+	// would mean promisfying each step of a test
+	if ( key === "log" ) {
+		callbacks.map( callback => callback( args ) );
+		return;
+	}
+
+	// ensure that each callback is executed serially
+	return callbacks.reduce( ( promiseChain, callback ) => {
+		return promiseChain.then( () => {
+			return Promise.resolve( callback( args ) );
+		} );
+	}, Promise.resolve( [] ) );
 }
--- a/src/test.js
+++ b/src/test.js
@@ -4,6 +4,7 @@ import { begin } from "./core";
 import { setTimeout, clearTimeout } from "./globals";
 import { emit } from "./events";
 import Assert from "./assert";
+import Promise from "./promise";
 
 import config from "./core/config";
 import {
--- a/src/test.js
+++ b/src/test.js
@@ -112,34 +112,37 @@ Test.prototype = {
 	before: function() {
 		var i, startModule,
 			module = this.module,
+			promises = [],
 			notStartedModules = getNotStartedModules( module );
 
 		for ( i = notStartedModules.length - 1; i >= 0; i-- ) {
 			startModule = notStartedModules[ i ];
 			startModule.stats = { all: 0, bad: 0, started: now() };
 			emit( "suiteStart", startModule.suiteReport.start( true ) );
-			runLoggingCallbacks( "moduleStart", {
+			promises.push( runLoggingCallbacks( "moduleStart", {
 				name: startModule.name,
 				tests: startModule.tests
+			} ) );
+		}
+
+		return Promise.all( promises ).then( () => {
+			config.current = this;
+
+			this.testEnvironment = extend( {}, module.testEnvironment );
+
+			this.started = now();
+			emit( "testStart", this.testReport.start( true ) );
+			return runLoggingCallbacks( "testStart", {
+				name: this.testName,
+				module: module.name,
+				testId: this.testId,
+				previousFailure: this.previousFailure
 			} );
-		}
-
-		config.current = this;
-
-		this.testEnvironment = extend( {}, module.testEnvironment );
-
-		this.started = now();
-		emit( "testStart", this.testReport.start( true ) );
-		runLoggingCallbacks( "testStart", {
-			name: this.testName,
-			module: module.name,
-			testId: this.testId,
-			previousFailure: this.previousFailure
+		} ).then( () => {
+			if ( !config.pollution ) {
+				saveGlobal();
+			}
 		} );
-
-		if ( !config.pollution ) {
-			saveGlobal();
-		}
 	},
 
 	run: function() {
@@ -313,7 +316,7 @@ Test.prototype = {
 		emit( "testEnd", this.testReport.end( true ) );
 		this.testReport.slimAssertions();
 
-		runLoggingCallbacks( "testDone", {
+		return runLoggingCallbacks( "testDone", {
 			name: testName,
 			module: moduleName,
 			skipped: skipped,
@@ -329,21 +332,26 @@ Test.prototype = {
 
 			// Source of Test
 			source: this.stack
-		} );
-
-		if ( module.testsRun === numberOfTests( module ) ) {
-			logSuiteEnd( module );
-
-			// Check if the parent modules, iteratively, are done. If that the case,
-			// we emit the `suiteEnd` event and trigger `moduleDone` callback.
-			let parent = module.parentModule;
-			while ( parent && parent.testsRun === numberOfTests( parent ) ) {
-				logSuiteEnd( parent );
-				parent = parent.parentModule;
+		} ).then( function() {
+			if ( module.testsRun === numberOfTests( module ) ) {
+				return logSuiteEnd( module );
 			}
-		}
+		} ).then( function() {
+			if ( module.testsRun === numberOfTests( module ) ) {
+				const promises = [];
 
-		config.current = undefined;
+				// Check if the parent modules, iteratively, are done. If that the case,
+				// we emit the `suiteEnd` event and trigger `moduleDone` callback.
+				let parent = module.parentModule;
+				while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+					promises.push( logSuiteEnd( parent ) );
+					parent = parent.parentModule;
+				}
+				return Promise.all( promises );
+			}
+		} ).then( function() {
+			config.current = undefined;
+		} );
 
 		function logSuiteEnd( module ) {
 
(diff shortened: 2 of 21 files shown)


step2yeung:
Looks like 2 cli tests is failing, will be looking into it

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -138,8 +139,7 @@ not ok 1 global failure
     at done (.*)
     at advanceTestQueue (.*)
     at Object.advance (.*)
-    at unblockAndAdvanceQueue (.*)
-    at <anonymous>
+    at unblockAndAdvanceQueue (.*)(\n    at <anonymous>)*
```
platinumazure: Should this end with a `?` (if only zero or one of those show up)? Or could we see more than one?

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -161,8 +162,7 @@ not ok 1 global failure
...
```
platinumazure: Should this end with a `?` (if only zero or one of those show up)? Or could we see more than one?

src/test.js
```diff
@@ -329,21 +332,26 @@ Test.prototype = {
...
-				logSuiteEnd( parent );
-				parent = parent.parentModule;
+		} ).then( function() {
+			if ( module.testsRun === numberOfTests( module ) ) {
+				return logSuiteEnd( module );
 			}
-		}
-
-		config.current = undefined;
+		} ).then( function() {
```
platinumazure: I'm a bit confused:

1. Why are there multiple `.then()` calls in sequence?
1. There appear to be two areas of logic where `logSuiteEnd()` is called. Why do we have two areas of logic and why are we separating the handling for this module vs any parent modules?

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -126,15 +126,21 @@ not ok 2 global failure
...
   severity: failed
-  actual: null
+  actual: {}
   expected: undefined
-  stack: undefined:undefined
+  stack: Error: No tests were run.
+    at done (.*)
+    at advanceTestQueue (.*)
+    at Object.advance (.*)
+    at unblockAndAdvanceQueue (.*)(\n    at <anonymous>)*
```
platinumazure: Should this end with a `?` character? Or is there a chance we would see multiple levels of these anonymous stack entries?

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -126,15 +126,21 @@ not ok 2 global failure
...
 not ok 1 global failure
   ---
   message: "No tests were run."
   severity: failed
-  actual: null
+  actual: {}
   expected: undefined
-  stack: undefined:undefined
+  stack: Error: No tests were run.
+    at done (.*)
```
platinumazure: Should the parentheses on this and consequent lines be escaped?

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -143,15 +149,21 @@ not ok 1 global failure
...
   severity: failed
-  actual: null
+  actual: {}
   expected: undefined
-  stack: undefined:undefined
+  stack: Error: No tests matched the filter "no matches".
+    at done (.*)
+    at advanceTestQueue (.*)
+    at Object.advance (.*)
+    at unblockAndAdvanceQueue (.*)(\n    at <anonymous>)*
```
platinumazure: Should this end with a `?` character? Or is there a chance we would see multiple levels of these anonymous stack entries?

test/cli/fixtures/expected/tap-outputs.js
```diff
@@ -143,15 +149,21 @@ not ok 1 global failure
...
 not ok 1 global failure
   ---
   message: "No tests matched the filter "no matches"."
   severity: failed
-  actual: null
+  actual: {}
   expected: undefined
-  stack: undefined:undefined
+  stack: Error: No tests matched the filter "no matches".
+    at done (.*)
```
platinumazure: Should the parentheses on this and consequent lines be escaped?

test/cli/fixtures/expected/tap-outputs.js
step2yeung: good call, it should be!

test/cli/fixtures/expected/tap-outputs.js
step2yeung: yep, this should.

test/cli/fixtures/expected/tap-outputs.js
step2yeung: yep, this should.

test/cli/fixtures/expected/tap-outputs.js
step2yeung: yep, this should.

test/cli/fixtures/expected/tap-outputs.js
step2yeung: yep, this should.

src/test.js
step2yeung: Sorry for the delay @platinumazure.
1. Since each `.then()` handles an individual promise, which from the previous non promise based implementation, seemed to be running in sequential order. Hence I chained them together. 
2. I was going with the old logic, and didn't put too much thought into this. Do you have any suggestions?

step2yeung:
@platinumazure about the parentheses being escaped, eslint throws `no-useless-escape` linting errors when I try to escape it. So i will keep it as it is.

src/core/logging.js
```diff
@@ -34,10 +34,7 @@ export function registerLoggingCallbacks( obj ) {
...
 
 export function runLoggingCallbacks( key, args ) {
-	var i, l, callbacks;
-
-	callbacks = config.callbacks[ key ];
-	for ( i = 0, l = callbacks.length; i < l; i++ ) {
-		callbacks[ i ]( args );
-	}
+	var callbacks = config.callbacks[ key ];
+	var promises = callbacks.map( callback => Promise.resolve( callback( args ) ) );
```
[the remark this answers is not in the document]
step2yeung: Thanks for pointing this out. When I was implemented this, I contemplated this, and went with the concurrent implementation, not sure whether that was the correct choice.
I will have the callbacks run serially.

src/test.js
```diff
@@ -112,34 +112,37 @@ Test.prototype = {
...
 			module = this.module,
+			promises = [],
 			notStartedModules = getNotStartedModules( module );
 
 		for ( i = notStartedModules.length - 1; i >= 0; i-- ) {
 			startModule = notStartedModules[ i ];
 			startModule.stats = { all: 0, bad: 0, started: now() };
 			emit( "suiteStart", startModule.suiteReport.start( true ) );
-			runLoggingCallbacks( "moduleStart", {
+			promises.push( runLoggingCallbacks( "moduleStart", {
```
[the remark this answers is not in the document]
step2yeung: Agreed

src/test.js
```diff
@@ -329,21 +332,26 @@ Test.prototype = {
...
-		config.current = undefined;
+		} ).then( function() {
+			if ( module.testsRun === numberOfTests( module ) ) {
+				const promises = [];
+
+				// Check if the parent modules, iteratively, are done. If that the case,
+				// we emit the `suiteEnd` event and trigger `moduleDone` callback.
+				let parent = module.parentModule;
+				while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+					promises.push( logSuiteEnd( parent ) );
```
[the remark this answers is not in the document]
step2yeung: Agreed

src/test.js
step2yeung: I had some time to revisit that code, and made changes to both 1 and 2. 
Particularly for 2, I changed it so that all modules (this module or parent's modules) are handled in the same code path

step2yeung:
@platinumazure @trentmwillis addressed the feedbacks, would you be able to give it another go to see if there's any other sticking point?

rwjblue:
After talking a bit with @step2yeung about the various options here, we came up with two different paths forward:

1. Refactor this pull request to _allow_ promises to be returned without rewriting all of the code to assume promises are available. This would be pretty similar to the `Test.prototype.resolvePromise` code that @trentmwillis mentioned inline. Ultimately, I believe that this would satisfy the requirements of this particular PR's feature, but it would leave the codebase in a somewhat worse shape (all of the promise detection would add even more cognitive complexity).
2. Bundle a lightweight promise implementation _within_ the `qunit.js` asset (which would **not** be accessible outside of QUnit's own code) that we would fallback on if global `Promise` is not available. This would increase the payload size of the `qunit.js` asset by some non-trivial amount (anywhere from 3kb to 10kb as a _super_ rough guess), but would allow us to author this feature (and refactor other promisey features that already exist) in a much more idiomatic style.

I personally think option 2 above is better for the longer term project health and generally leaves things better off for future iteration, but am pretty confident that we can make either option work if including a fallback promise lib isn't acceptable.

@stefanpenner / @trentmwillis - What do you think?

trentmwillis:
@rwjblue / @step2yeung, in general, I'm not too concerned with the file size of QUnit given that it is a development tool. I agree with your opinion that option 2 is the better approach for the long-run, but would be fine with option 1 as well as I don't think it'll negatively affect the code base much if it is factored well (likely a function or two that can be swapped out with native Promise functions later).

I originally suggested an approach like option 1 as I thought it would be quicker to get landed. Introducing an extra library that has fallback behavior will require modifications to the build system and some special testing.

With that said, I leave the choice up to the implementer!

rwjblue:
@step2yeung - Looks like you need to rebase against master and resolve a `yarn.lock` conflict...

step2yeung:
@rwjblue rebased!

step2yeung:
Since we don't have appyoyer or browser stack tests turned on, I tested the above on a windows machine by running `yarn dev` and opening up the url tests (`test/index.html`, `test/autostart.html`, etc..) on IE11 which passes.
@trentmwillis @rwjblue mind giving another look?

src/core/logging.js
```diff
@@ -34,10 +35,21 @@ export function registerLoggingCallbacks( obj ) {
...
-	callbacks = config.callbacks[ key ];
-	for ( i = 0, l = callbacks.length; i < l; i++ ) {
-		callbacks[ i ]( args );
+	var callbacks = config.callbacks[ key ];
+
+	// Handling 'log' callbacks separately. Unlike the other callbacks,
+	// the log callback is not controlled by the processing queue,
+	// but rather used by asserts. Hence to promisfy the 'log' callback
+	// would mean promisfying each step of a test
+	if ( key === "log" ) {
```
trentmwillis: I think this is a reasonable tradeoff to make. We should add a note to the docs however that this callback must be synchronous.

trentmwillis:
Also, we should definitely add support for this to `QUnit.on()` as we eventually would like to deprecate the other callbacks in favor of `on`. This can be done in a separate PR though.

src/core/logging.js
step2yeung: Added notes to the doc for QUnit.log and QUnit.on

step2yeung:
Created #1329 to keep track of the two issues you raised.

trentmwillis:
Thank you! I know that was a lot of work, but it's appreciated; I'm excited to see how y'all use it.

I'll try to do a new minor release tomorrow morning (unlikely to get to it today).

dcombslinkedin:
Way to go, @step2yeung !!!! This is awesome stuff!

PR #1310, opened 2018-09-18
HTML Reporter: add norender flag to avoid deeply nested divs

gabrielcsapo:
What? 

- Adds a flag to avoid rendering tests after completed

Why?

When running large test suites, rendering test output for all passing tests will cause the page to become unresponsive.

```
> document.querySelector('#qunit-tests').children.length
> 24280
```

The above output is from:

```
81 tests completed in 97301 milliseconds, with 2 failed, 2 skipped, and 0 todo.
210 assertions of 217 passed, 7 failed.
```

While the tests are running, the page becomes more and more unresponsive while appending output items. `24280` only represents the immediate children, the full child tree contains `146780` nodes.

```
> const tests = document.querySelector('#qunit-tests');
> tests.getElementsByTagName("*").length
> 146780
```

Solution?

- Add a `norender` option to not render test output given the filter criteria. 

norender-example

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -754,7 +754,7 @@ export function escapeText( s ) {
 
 	QUnit.log( function( details ) {
 		var assertList, assertLi,
-			message, expected, actual, diff,
+			status, message, expected, actual, diff,
 			showDiff = false,
 			testItem = id( "qunit-test-output-" + details.testId );
 
@@ -762,8 +762,8 @@ export function escapeText( s ) {
 			return;
 		}
 
-		message = escapeText( details.message ) || ( details.result ? "okay" : "failed" );
-		message = "<span class='test-message'>" + message + "</span>";
+		status = escapeText( details.message ) || ( details.result ? "okay" : "failed" );
+		message = "<span class='test-message'>" + status + "</span>";
 		message += "<span class='runtime'>@ " + details.runtime + " ms</span>";
 
 		// The pushFailure doesn't provide details.expected
@@ -844,7 +844,7 @@ export function escapeText( s ) {
 	} );
 
 	QUnit.testDone( function( details ) {
-		var testTitle, time, testItem, assertList,
+		var testTitle, time, testItem, assertList, status,
 			good, bad, testCounts, skipped, sourceName,
 			tests = id( "qunit-tests" );
 
@@ -852,7 +852,21 @@ export function escapeText( s ) {
 			return;
 		}
 
+
 		testItem = id( "qunit-test-output-" + details.testId );
+
+		if ( details.failed > 0 ) {
+			status = "failed";
+		} else if ( details.todo ) {
+			status = "todo";
+		} else {
+			status = details.skipped ? "skipped" : "passed";
+		}
+
+		// can accept 'passed' or 'passed,skipped,todo'
+		if ( config.norender && config.norender.split( "," ).indexOf( status ) > -1 ) {
+			testItem.remove();
+		}
 
 		assertList = testItem.getElementsByTagName( "ol" )[ 0 ];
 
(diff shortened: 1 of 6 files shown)


reporter/html.js
```diff
@@ -844,16 +844,27 @@ export function escapeText( s ) {
...
 			return;
 		}
 
+
 		testItem = id( "qunit-test-output-" + details.testId );
 
+		if ( details.failed > 0 ) {
+			status = "failed";
+		} else {
+			status = details.skipped ? "skipped" : "passed";
```
step2yeung: a test could also be `todo`, defined by `details.todo`.

reporter/urlparams.js
```diff
@@ -34,6 +34,17 @@ import { window } from "../src/globals";
...
 	QUnit.config.urlConfig.push(
+		{
+			id: "norender",
+			label: "Don't render the selected status filters that match the results " +
+			"for the tests run.",
+			tooltip: "Only show tests that are not selected",
+			value: [
+				"passed",
+				"failed",
+				"skipped"
```
step2yeung: "todo" is also a value

reporter/html.js
```diff
@@ -844,16 +844,27 @@ export function escapeText( s ) {
...
+		}
+
+		if ( config.norender && config.norender === status ) {
```
step2yeung: can norender be an array of values?

gabrielcsapo: added support for being able to parse comma separated string

reporter/urlparams.js
gabrielcsapo: add todo as a possible value

reporter/html.js
gabrielcsapo: added conditional for todo

stefanpenner:
> I'd be onboard with adding a url-only norender option that completely disables the HTML reporter when running the tests. I think this could be beneficial for headless environments where the UI is never going to be used anyway.

Actually, disabling these UI elements makes the test UI extremely responsive. I actually feel, especially with long running tests, that not rendering by default with hidePassed=true should be considered. 

The dev experience improves quite a bit.

stefanpenner:
@trentmwillis obviously If i where to have to pick, norender QP over no such feature, I would totally opt for the QP only mode. But I would encourage you to evaluate the dev ergonomics with what I propose, you may be pleasantly surprised.

trentmwillis:
@stefanpenner I don't think this is an "either-or" situation. I think we should have both: not rendering passed tests when using `hidepassed=true` _and_ a `norender` query param for disabling the reporter (naming up to debate)

stefanpenner:
@trentmwillis Ah, yes I can get behind that. Thank you for the clarification.

gabrielcsapo:
@trentmwillis I pushed your suggested changes, what do you think of it?

trentmwillis:
Sorry, maybe I'm missing something, but it doesn't seem like either of the suggested changes are present. I'll try to clarify my suggestions:

1. We should update the existing `hidepassed` option/url param to remove DOM nodes when enabled (rather than the current behavior of just hiding them with CSS). I think this should be done instead of introducing an additional option that also performs some "hiding" functionality (as is currently proposed in this PR).

2. We should introduce a new url param that completely disables the HTML reporter. We could call this `norender`, but after some further thought maybe we should call it `noreporter` as that is more descriptive of what it should do.

We could split those two tasks into separate PRs as they should be largely independent changes. #1 is what I think this PR is trying to accomplish. #2 is a feature which I think will provide further benefits when running browser-based tests in headless browsers.

gabrielcsapo:
@trentmwillis so `hidepassed` will do what the PR calls norender? in the case of norender, it will hide multiple tests, `hidepassed` only hides the passed tests.

gabrielcsapo:
closing this in favor of https://github.com/qunitjs/qunit/pull/1311 will open up another rb for #2 add noreporter for running tests in a headless mode.

PR #1311, opened 2018-09-19
HTML Reporter: remove test item when hidepassed is set to true

gabrielcsapo:
What? 

- make hidepassed remove item from div rather than just hiding it

Why?

When running large test suites, rendering test output for all passing tests will cause the page to become unresponsive.

```
> document.querySelector('#qunit-tests').children.length
> 24280
```

The above output is from:

```
81 tests completed in 97301 milliseconds, with 2 failed, 2 skipped, and 0 todo.
210 assertions of 217 passed, 7 failed.
```

While the tests are running, the page becomes more and more unresponsive while appending output items. `24280` only represents the immediate children, the full child tree contains `146780` nodes.

```
> const tests = document.querySelector('#qunit-tests');
> tests.getElementsByTagName("*").length
> 146780
```

Solution?

- actually remove child item when it passes

norender-example

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -41,6 +41,7 @@ export function escapeText( s ) {
 	}
 
 	var config = QUnit.config,
+		hiddenTests = [],
 		document = window.document,
 		collapseNext = false,
 		hasOwn = Object.prototype.hasOwnProperty,
@@ -206,7 +207,26 @@ export function escapeText( s ) {
 			config[ field.name ] = value || false;
 			tests = id( "qunit-tests" );
 			if ( tests ) {
-				toggleClass( tests, "hidepass", value || false );
+				var length = tests.children.length;
+				var children = tests.children;
+
+				if ( field.checked ) {
+					for ( var i = 0; i < length; i++ ) {
+						var test = children[ i ];
+
+						if ( test && test.className.indexOf( "pass" ) > -1 ) {
+							hiddenTests.push( test );
+						}
+					}
+
+					for ( const hiddenTest of hiddenTests ) {
+						tests.removeChild( hiddenTest );
+					}
+				} else {
+					while ( ( test = hiddenTests.pop() ) != null ) {
+						tests.appendChild( test );
+					}
+				}
 			}
 			window.history.replaceState( null, "", updatedUrl );
 		} else {
@@ -606,7 +626,7 @@ export function escapeText( s ) {
 
 	// HTML Reporter initialization and load
 	QUnit.begin( function( details ) {
-		var i, moduleObj, tests;
+		var i, moduleObj;
 
 		// Sort modules by name for the picker
 		for ( i = 0; i < details.modules.length; i++ ) {
@@ -622,10 +642,6 @@ export function escapeText( s ) {
 		// Initialize QUnit elements
 		appendInterface();
 		appendTestsList( details.modules );
-		tests = id( "qunit-tests" );
-		if ( tests && config.hidepassed ) {
-			addClass( tests, "hidepass" );
-		}
 	} );
 
 	QUnit.done( function( details ) {
@@ -844,7 +860,7 @@ export function escapeText( s ) {
 	} );
 
 	QUnit.testDone( function( details ) {
-		var testTitle, time, testItem, assertList,
+		var testTitle, time, testItem, assertList, status,
 			good, bad, testCounts, skipped, sourceName,
 			tests = id( "qunit-tests" );
 
@@ -853,6 +869,14 @@ export function escapeText( s ) {
 		}
 
 		testItem = id( "qunit-test-output-" + details.testId );
+
+		if ( details.failed > 0 ) {
+			status = "failed";
+		} else if ( details.todo ) {
+			status = "todo";
+		} else {
+			status = details.skipped ? "skipped" : "passed";
+		}
 
 		assertList = testItem.getElementsByTagName( "ol" )[ 0 ];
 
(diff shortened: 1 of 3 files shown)


gabrielcsapo:
@stefanpenner @trentmwillis sorry for the delay, but this is actually removing divs instead of hiding them when `hidepassed` is enabled.

reporter/html.js
```diff
@@ -206,7 +206,13 @@ export function escapeText( s ) {
 			config[ field.name ] = value || false;
 			tests = id( "qunit-tests" );
 			if ( tests ) {
-				toggleClass( tests, "hidepass", value || false );
+				var children = tests.children;
+				for ( var i = 0; i < children.length; i++ ) {
+					var test = children[ i ];
+					if ( test && test.className.indexOf( "pass" ) > -1 ) {
+						tests.removeChild( test );
```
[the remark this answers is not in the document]
trentmwillis: Yes please. If we retain the nodes in memory, will that cause issues with a test suite of the size y'all are working with?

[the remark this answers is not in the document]
gabrielcsapo: The only way I can see this working is if we 

1. retain all the nodes in memory and basically filter what is displaying based on the checkbox value. If it is checked, we update the list with only the failed, skipped or todo tests, if it isn't we display all. 

2. have a function that creates the output items using the data in memory so that we don't have to keep a reference to the dom element, but only to the data, then create the list using a function that uses the data to make the dom element.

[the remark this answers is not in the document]
trentmwillis: I had the same thoughts. Both approaches seem reasonable to me, but retaining the nodes would be predicated on it not causing too much memory pressure for large test suites such as the one y'all have. I don't think it should be an issue though.

Basically, what I'm saying is choice is up to you if both approaches work :)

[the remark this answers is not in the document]
gabrielcsapo: Sorry for the late response, pushed code that will keep hiddenTests in memory and not disrupt the current behavior.

test/reporter-urlparams-hidepassed.js
```diff
@@ -0,0 +1,30 @@
...
+	QUnit.test( "passed", function( assert ) {
+		assert.ok( true );
+	} );
+	QUnit.test( "passed", function( assert ) {
+		assert.ok( true );
+	} );
+
+	QUnit.test( "passed", function( assert ) {
+
+		// we have to set it to 1 becuase there is currently one item being rendered, this one as it is in progress
```
step2yeung: There's only one item being rendered because all the test name are the same. Is this intentional?

reporter/html.js
```diff
@@ -844,7 +860,7 @@ export function escapeText( s ) {
 	} );
 
 	QUnit.testDone( function( details ) {
-		var testTitle, time, testItem, assertList,
+		var testTitle, time, testItem, assertList, status,
```
step2yeung: Why do we need `status` again? Seems redundant to details.<status>

test/reporter-urlparams-hidepassed.js
gabrielcsapo: Well there should be 7 divs realistically, only one should exist as it is the current one.

reporter/html.js
gabrielcsapo: It was more of a style thing, we mutate message a ton I just thought it was a needed abstraction

gabrielcsapo: I can change it back

gabrielcsapo:
@trentmwillis would you might checking this out again? 😄

jackson-dean:
just to give more info about why the appendTestsList is currently problematic. With large test base (10s of thousands of test) that behavior (iterating over each test and appending to the live dom individually) becomes a serious bottleneck. 

running a perf profile with ~20K tests indicates that the browser is locked up for some 20 seconds before any test even runs.

platinumazure:
Can we append the tests to an element while it's not attached to the DOM, then append that whole container to the DOM at once? I feel that would perform better, although I could be wrong.

jackson-dean:
I am very much sure it would perform better, but from our experiments it didn't seem that appending those tests up front was actually being used for anything. although, I don't know about the history of why that was originally added and what purpose it serves (if any).

gabrielcsapo:
Yeah it is basically a dead code path, since we attach a new dom node when a test is initialized anyway

trentmwillis:
Note waiting until I get a response to the question above ^ to merge this

gabrielcsapo:
@trentmwillis how would we merge a document fragment with a document fragment? 

```
hidden-tests <-> non-hidden-tests -> ol
``` 

> This would mandate we worry about currently running tests and make sure we don't merge during a dirty dom state, right?

I don't know if optimizing the use case of toggling between succeeded and non-succeeded cases should be done in this either.

trentmwillis:
> I’m not sure that splatting a fragments contents into the DOM would be any different since (in our use case) we are doing all the work synchronously anyways.

Ah good point.

Given the feedback, we'll ignore this for now and look at it again if someone complains :) Merging! Thanks for working on this!

It'll be in the next release (which will likely be when #1307 lands).

stefanpenner:
@trentmwillis #1307 looks like it will require some more work, is it possible for you guys to get a release out of current master sooner? I would like to avoid the extra work required to rollout an interim fork, obviously if thats not possible an interim fork on our end is viable. Please advise

trentmwillis:
Sure, I'll cut a release tomorrow. I won't have time to get to it today. If the other PR is updated before then, it could be released together, otherwise we'll just do two separate releases.

gabrielcsapo:
thank you @trentmwillis!

stefanpenner:
@trentmwillis <3 <3 <3

test/reporter-urlparams-hidepassed.js
Krinkle: Fixed in https://github.com/qunitjs/qunit/commit/f64b5a113bac52eb1bf3dfbf48f896fe2da6bfd4.

PR #1315, opened 2018-09-27
HTML Reporter: Add showskipped option

sechel:
Hi, I felt that the default option to show skipped tests should be false. So I added an option to show the skipped tests to the HTML reporter.

### DIFF
--- a/docs/config/QUnit.config.md
+++ b/docs/config/QUnit.config.md
@@ -43,6 +43,10 @@ By default QUnit will use whatever the starting content of `#qunit-fixture` is a
 
 By default, the HTML Reporter will show all the tests results. Enabling this option will make it show only the failing tests, hiding all that pass. This can also be managed by the HTML interface.
 
+### `QUnit.config.showskipped` (boolean) | default: `false`
+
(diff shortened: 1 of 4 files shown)


Krinkle:
Hi @sechel, thanks for submitting this proposal.

Sorry for the delay in reviewing this, going through a bit of a backlog now. From a quick glance, it appears this proposal would change the API of the HTML reporter. As such, I'm tagging it as `meta` to also seek input from others (from anyone, but especially core team).

trentmwillis:
A relevant issue for this feature: https://github.com/qunitjs/qunit/issues/1208

In general, I am in favor of this feature, but I think we'll need to refactor the UI some to avoid it becoming too cluttered. I mentioned this in the issue above at https://github.com/qunitjs/qunit/issues/1208#issuecomment-312980746.

Krinkle:
I'm declining this for now. I think hiding this by default has potential to hide issues or make it more difficult to figure out what's wrong when investigating an issue. The list of tests that have been run or considred to run is imho expected to be complete.

I do note that currently "hiding passed" is leaving skipped ones behind. I'll close this but leave the ticket open to ask about changing that.

PR #1319, opened 2018-10-11
Core: guards performance.measure to not throw and cause tests to fail

gabrielcsapo:
```
[17:05:03] not ok 10 Chrome 69.0 - [undefined ms] - <undefined> - Global error: Uncaught SyntaxError: Failed to execute 'measure' on 'Performance': The mark 'qunit_test_start' does not exist. at http://localhost:7357/assets/test-support.js, line 4335
 While executing test: ...
```

### DIFF
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -1,4 +1,5 @@
 import { window, setTimeout } from "../globals";
+import Logger from "../logger";
 
 export const toString = Object.prototype.toString;
 export const hasOwn = Object.prototype.hasOwnProperty;
@@ -17,6 +18,17 @@ function detectPerformanceApi() {
 		typeof window.performance !== "undefined" &&
 		typeof window.performance.mark === "function" &&
 		typeof window.performance.measure === "function";
+}
+
+export function measure( comment, startMark, endMark ) {
+
+	// `performance.measure` may fail if the mark could not be found.
+	// reasons a specific mark could not be found include: outside code invoking `performance.clearMarks()`
+	try {
+		performance.measure( comment, startMark, endMark );
+	} catch ( ex ) {
+		Logger.warn( "performance.measure could not be executed because of ", ex.message );
+	}
 }
 
 export const defined = {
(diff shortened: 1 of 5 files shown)


gabrielcsapo:
@rwjblue it happens in `ember-exam` not sure what is happening, but guarding fixes it

rwjblue:
> it happens in ember-exam not sure what is happening
> This is just calling attention to the issue, will deep dive with @step2yeung

Sounds good! Definitely needs a bit of root cause analysis before just guarding here (IMHO)...

trentmwillis:
I agree with @rwjblue. I'd like to understand how we get unbalanced as that shouldn't happen.

gabrielcsapo:
The issue occurs when the test itself triggers a navigation event and destroys the marks in the process.

gabrielcsapo:
So using ember-native-dom-helpers the issue persists because it calls browser events that will trigger reflow and the location state to change, using ember-test-helpers we call the router service and don't trigger a reflow of the document directly, but through the router service.

stefanpenner:
@trentmwillis I've encountered this before, simply put QUnit's `performance.measure` don't need to be unbalanced to cause this error, as other code can directly or indirectly call: `performance.clearMarks()`

stefanpenner:
cc @Turbo87

trentmwillis:
> simply put QUnit's performance.measure calls don't need to be unbalanced to cause this error, as other code can directly or indirectly call: performance.clearMarks()

Good point that I had not considered. Given that, I'm fine with introducing some guards.

gabrielcsapo:
@trentmwillis the code is different for suites and single tests, so you would want to abstract 

```
function measure(comment, start_mark, end_mark) {
  try {
    performance.measure(comment, start_mark, end_mark
  } catch(ex) {
    Logger.warn('performance.measure could not be executed because of ', ex.message)
  }
}
```

gabrielcsapo:
refactored out the measure code into utilities and added tests to ensure that clearing marks in tests does not cause errors to be thrown.

kratiahuja:
Can we please have a release of this change? qunit v2.7 has the issue currently.

gabrielcsapo:
@trentmwillis bump on the publish to 2.7.1?

trentmwillis:
v2.7.1 has been released

gabrielcsapo:
Thank you @trentmwillis!!!

PR #1323, opened 2018-10-22
Core: adds back running class to items and adds tests

gabrielcsapo:
the running class was removed and no existing tests catch this, I added it back and now added with it a test to ensure this doesn't happen again.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -724,7 +724,10 @@ export function escapeText( s ) {
 		appendTest( details.name, details.testId, details.module );
 
 		running = id( "qunit-testresult-display" );
+
 		if ( running ) {
+			addClass( running, "running" );
+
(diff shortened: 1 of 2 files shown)


test/reporter-html-running-class.js
```diff
@@ -0,0 +1,5 @@
+QUnit.module( "Ensure that html reporter adds running class", function() {
```
trentmwillis: Thanks for adding a test! This, however, should probably go in the existing `reporter-html` test file: https://github.com/qunitjs/qunit/blob/master/test/reporter-html/reporter-html.js

gabrielcsapo: @trentmwillis good idea, just added the test there

trentmwillis:
Thank you!

PR #1326, opened 2018-10-24
Core: Ensure inherited getters are not invoked by assert.deepEqual

rwjblue:
Previously, given the following:

```js
class Foo {
  constructor(a = 1) {
    this.a = a;
  }
}

Object.defineProperty(Foo.prototype, 'b', {
  enumerable: true,
  get() {
    return new Foo(this.a + 1);
  }
})

QUnit.test( "hello test", function( assert ) {
  assert.deepEqual(new Foo(), new Foo());
});
```

The `assert.deepEqual` invocation would never complete (ultimately crashing the tab or node process).

The changes here ensure that inherited descriptors (getter / setter _or_ value only) are compared without invoking the getter therefore preventing the issue mentioned above.

Fixes https://github.com/qunitjs/qunit/issues/1325

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -13,6 +13,31 @@ export default ( function() {
 		return obj.__proto__;
 	};
 
+	function lookupDescriptor( obj, keyName ) {
+		let current = obj;
+		do {
+			const descriptor = Object.getOwnPropertyDescriptor( current, keyName );
+			if ( descriptor !== undefined ) {
+				return descriptor;
+			}
+			current = getProto( current );
+		} while ( current !== null );
+		return null;
+	}
+
+	function hasSharedDescriptor( a, b, keyName ) {
+		var aDescriptor = lookupDescriptor( a, keyName );
+		var bDescriptor = lookupDescriptor( b, keyName );
+
+		return ( aDescriptor !== null && bDescriptor !== null ) &&
+			aDescriptor.value === bDescriptor.value &&
+			aDescriptor.get === bDescriptor.get &&
+			aDescriptor.set === bDescriptor.set &&
+			aDescriptor.writable === bDescriptor.writable &&
+			aDescriptor.configurable === bDescriptor.configurable &&
+			aDescriptor.enumerable === bDescriptor.enumerable;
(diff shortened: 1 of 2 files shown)


rwjblue:
Hmm, weird, CI is done (and green [see here](https://travis-ci.org/qunitjs/qunit/builds/445897600?utm_source=github_status&utm_medium=notification)) but still shows as in progress in the GitHub UI.

src/equiv.js
```diff
@@ -13,6 +13,31 @@ export default ( function() {
...
+				return descriptor;
+			}
+			current = getProto( current );
+		} while ( current !== null );
+		return null;
+	}
+
+	function hasSharedDescriptor( a, b, keyName ) {
+		var aDescriptor = lookupDescriptor( a, keyName );
+		var bDescriptor = lookupDescriptor( b, keyName );
```
gibson042: This ignores where on the prototype chain the keyName property is defined when checking if it is "shared"... if that is intentional, then it should be documented and tested.

Krinkle:
I'll do close/re-open trick here to let Travis CI re-merge/re-test this pull.

Krinkle:
I agree with @gibson042 that treating these as equal is not obviously better and indeed has potential to wrongly mask other genuine cases and/or cause a different kind of confusion instead.

I'm closing this for now, but I do want to acknowledge that https://github.com/qunitjs/qunit/issues/1325 is a bug that should be fixed. I'll respond there with some ideas.

wycats:
@Krinkle weirdly, I just hit this issue today. I was surprised by the behavior and it took me quite some time to track down.

I would be pretty happy with a mode that required me to specify a comparison function for instances of specific classes.

Krinkle:
@wycats

Krinkle:
@wycats Could you elaborate a bit with an example or pointer to your test? Was it a case of something where you'd like to be able to succesfully compare as equal to lazily-initialised recursive structures, or was it case were something accidentally made it into the comparison and you'd like QUnit to surface that quickly and clearly rather than with a timeout?

PR #1334, opened 2018-12-15
Build: Add --connect-port for Gruntfile to override 8000

Krinkle:
When something is locally running on port 8000 already, the
tests consistently fail. For now, spread the approach we used
for livereload and allow it to be specified by the user.

This is a pretty poor user experience, but until we find a way
to not need, or do differently, the urls in grunt-contrib-qunit,
we need to know it synchronously while calling grunt.initConfig,
which basically leaves manual CLI options as the only option.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -14,6 +14,7 @@ var isCI = process.env.CI || process.env.JENKINS_HOME;
 
 module.exports = function( grunt ) {
 	var livereloadPort = grunt.option( "livereload-port" ) || 35729;
+	var connectPort = Number( grunt.option( "connect-port" ) ) || 8000;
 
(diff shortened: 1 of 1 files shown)


Krinkle:
@platinumazure Yeah, I pushed the same commit twice under a different branch name (`test-port` and `test/port`).

I created the PR for `test/port`, and the `ci/pr` test is working. However, the `ci/push` test for the branch had already started and it seems Travis is refusing to start a a build for the other branch name.

It's stuck on associating the commit hash with the old `test-port` branch, despite it no longer existing and despite the PR not being associated with that.

I've worked around it now by using their Beta Feature for manually triggering a build and selecting the `test/port` branch. Seems to have done the trick, in that it re-attached this PR to that. Yay :)

PR #1336, opened 2018-12-15
Build: Remove 'npm link' from 'npm run test:cli' command

Krinkle:


trentmwillis:
So this doesn't work because the tests then try to execute `qunit <something>` but the `qunit` global won't be found

platinumazure:
Can we wrap the call to QUnit in an npm script and then call `npm run` on that script in Grunt?

PR #1337, opened 2018-12-15
Use GitHub Pages in development environment

DirtyF:
Should fix https://github.com/jekyll/jekyll/issues/7426

Bump dependencies to match GitHub's production environment.
fix config


Krinkle:
Closing per https://github.com/jekyll/jekyll/issues/7426#issuecomment-447607128. I am already working on a similar PR, and the problem is exactly the opposite. Without this change, things work locally. Switching to github-pages is causing the reported bug.

PR #1338, opened 2018-12-16
Docs: Switch to github-pages gem

Krinkle:
To ensure local testing accurately reflects what will be published,
use the github-pages gem instead of specifying our Jekyll version
and plugins directly.


trentmwillis:
Tests are failing due to https://github.com/qunitjs/qunit/pull/1336. Going to go ahead and merge this though as it should have no impact on the actual library.

PR #1341, opened 2018-12-20
HTML Reporter: Fix an unescaped details.source.

shlomif:
Fix an unescaped details.source  in innerHTML.

It became apparent when using .xhtml (application/xml+xhtml) but may
have other XSS issues (see
https://en.wikipedia.org/wiki/Cross-site_scripting ) in plain HTML.

The new test gets stuck before the production code fix and completes
successfully and promptly after applying it.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -162,6 +162,7 @@ module.exports = function( grunt ) {
 						"test/reporter-html/single-testid.html",
 						"test/reporter-html/window-onerror.html",
 						"test/reporter-html/window-onerror-preexisting-handler.html",
+						"test/reporter-html/xhtml-escape-details-source.xhtml",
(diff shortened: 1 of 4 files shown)


platinumazure:
@Krinkle @trentmwillis Are the test failures related to the changes in this PR? At a quick glance, I'm not sure they are, but maybe I'm missing something.

Krinkle:
They're unrelated, and have been fixed since the PR started. I'll close and re-open this PR so that GitHub creates a new merge commit and let Travis test that.

Krinkle:
My bad. The fix for that failure is still pending at <https://github.com/qunitjs/qunit/pull/1339>. I'll get that sorted now.

trentmwillis:
Merged #1339, restarting Travis...

PR #1343, opened 2018-12-24
Tests: Add case for errors while console is mocked

Krinkle:
The asserted output is not what we'd like it to be.
This is just to document what happens today.

Ref https://github.com/qunitjs/qunit/issues/1340.


Krinkle:
I'll fix it together with the added test.

PR #1344, opened 2018-12-27
CLI: Move non-bin files from bin/ to src/cli/

Krinkle:
Follows-up https://github.com/qunitjs/qunit/pull/1271.


rwjblue:
This forgot to update the `files` array in `package.json`, `bin/` was listed but `src/` is not. This results in failures in published builds.

PR #1360, opened 2018-12-30
Build: Add macOS and Windows to the Travis CI test matrix

### LINKED ISSUE #1347: Require Windows tests passing on pull requests
To avoid https://github.com/qunitjs/qunit/pull/1346 in the future, and to embrace Windows support, perhaps we should integrate AppVeyor for pull requests and require it to be passing like we do for Linux (on Travis CI).

Alternatively, we could look into adding Windows to the Travis CI build matrix, which has experimental support as of <https://blog.travis-ci.com/2018-10-11-windows-early-release>.

Thoughts?

/cc @platinumazure @trentmwillis

Krinkle:
Fixes #1347.

### grunt-contrib-qunit

Update grunt-contrib-qunit to 3.1.0, because grunt-contrib-qunit 3.0.1 uses a pinned (older) version of puppeteer. The older puppeteer used an older version of Chrome, which is incompatible with Windows Server 2016, which is what Travis CI offers. AppVeyor was using Windows Server 2012, which was not affected by that Chrome bug.

Ref https://github.com/gruntjs/grunt-contrib-qunit/issues/154.

### test:main

I've changed the `npm test` script to always run both the main and the CLI tests. I think this would improve the local development experience by having the default entry point for developers match what CI will run. (I've more than once broke the CLI tests and not found until Travis points it out..)

However, our Windows tests on AppVeyor were only passing because they ran `npm test` without `npm run test:cli`. So I've kept that as-is by introducing `npm run test:main` as a subset for that purpose. Making the CLI test pass on Windows is https://github.com/qunitjs/qunit/issues/1359.

### eol

Rename bin/qunit to bin/qunit.js, because otherwise ESLint fails:

> Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style

I would've preferred to do this as a separate change later on, and in theory it should be possible to add an entry to `.gitattributes` that will apply to bin/qunit like what we apply to *.js, but I wasn't able to find a way.


Krinkle:
@platinumazure Yeah. The `/push` build is passing without issue and the logs look clean. Although, it appears the `/pr` build is failing due to ESLint detecting bad line endings, despite the change to `gitattributes`. Not sure why it's different...

If I can't figure it out in a little while, I suppose we could start by depending on both CI systems during merges.

PR #1369, opened 2019-01-09
fix: HTML reporter should not require CSP style-src 'unsafe-inline'

jelhan:
Using `style` attribute of an Element requires Content-Security-Policy (CSP) `style-src 'unsafe-inline'`. This makes testing that a library or an application does not violate a strict CSP more difficult.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -350,12 +350,16 @@ export function escapeText( s ) {
 	}
 
 	function toolbarModuleFilter() {
-		var allCheckbox, commit, reset,
+		var commit, reset,
 			moduleFilter = document.createElement( "form" ),
 			label = document.createElement( "label" ),
 			moduleSearch = document.createElement( "input" ),
 			dropDown = document.createElement( "div" ),
 			actions = document.createElement( "span" ),
+			applyButton = document.createElement( "button" ),
+			resetButton = document.createElement( "button" ),
+			allModulesLabel = document.createElement( "label" ),
+			allCheckbox = document.createElement( "input" ),
 			dropDownList = document.createElement( "ul" ),
 			dirty = false;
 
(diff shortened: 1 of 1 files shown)


reporter/html.js
```diff
@@ -370,15 +374,27 @@ export function escapeText( s ) {
...
+		applyButton.style.display = "none";
+
+		resetButton.textContent = "Reset";
+		resetButton.type = "reset";
+		resetButton.style.display = "none";
+
+		allCheckbox.type = "checkbox";
+		allCheckbox.checked = config.moduleId.length === 0;
+
+		allModulesLabel.classList.add( "clickable" );
```
Krinkle: I suspect classList we might still support browsers in QUnit that predate `classList`. Would need to use `className` instead, if so.

reporter/html.js
```diff
@@ -370,15 +374,27 @@ export function escapeText( s ) {
 		label.innerHTML = "Module: ";
 		label.appendChild( moduleSearch );
 
+		applyButton.textContent = "Apply";
+		applyButton.style.display = "none";
```
Krinkle: I haven't confirmed, so apologies if this isn't true... I would expect the CSP policy for inline styles to disallow all three methods of dynamic style creation: `<style>` elements, `style=""` attributes, and CSSOM/style properties (like this one).

To abide the CSP strict style policy, I think we need to declare the styles in the CSS file ahead of time, and toggle them with a class only. For example, the `qunit-modulefilter-dropdown` element could have a class like `has-selection` or `has-no-selection`. And based on that, we could hide the buttons as needed.

However, the apply button isn't the only violation, so to address your use case we'll need numerous other changes as well. Again, that's based on my (potentially) outdated understanding of CSP.

jelhan: The specification is [not quite clear regarding CSSOM and CSP](https://www.w3.org/TR/CSP3/#directive-style-src) but as far as I'm aware of none of the implementation considers manipulating styles through CSSOM as a CSP violation. So while `document.create('<button style="display: none;">')` and `document.create('button').style = "display: none;"` require `style-src 'unsafe-inline'`, `document.create('button').style.display = 'none'` is not violation `style-src: 'none'`.

reporter/html.js
jelhan: Didn't expected anyone to still support IE9 and IE10 as even Microsoft has dropped support more that two years ago. But you are right, QUnit webseite says:

> QUnit currently supports the same browsers as jQuery 3.x.

And jQuery lists _Internet Explorer: 9+_.

Pushed the change.

jelhan:
@Krinkle Is there still something needed here?

PR #1370, opened 2019-01-18
Update `resolve` to v1.9.0

Turbo87:
This adds support for Yarn Plug'n'Play if Yarn 1.13+ is used. I'm not saying `qunit` supports it yet, but at least the import resolution should start to work.

/cc @arcanis @rwjblue

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "dependencies": {
     "commander": "2.12.2",
     "js-reporters": "1.2.1",
-    "resolve": "1.5.0",
+    "resolve": "1.9.0",
     "node-watch": "0.6.0",
     "minimatch": "3.0.4"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
   "dependencies": {
     "commander": "2.12.2",
     "js-reporters": "1.2.1",
-    "resolve": "1.5.0",
+    "resolve": "^1.9.0",
```
rwjblue: This swaps to using `^` whereas all of the other dependencies are specified explicitly. 

I don't really understand why we would want to pin everything the way it is currently being done. IMHO we should swap all of these to using caret, but it seems like something we should fix across the board and not per-dep...

platinumazure: I'll defer to Trent, but my personal preference would be to pin in this PR and then we can unpin everything in a future PR (if the team agrees).

Krinkle:
Thanks @Turbo87.

We already had this version installed indirectly, but not yet used by the qunit/cli invocation. I've reviewed the upstream changes at <https://github.com/browserify/resolve/compare/v1.8.1...v1.9.0>.

Always nice to reduce the install size, this removes a few duplicate copies of it.

PR #1376, opened 2019-02-21
Core: Ensuring semaphores are balanced when timeout occurs

### LINKED ISSUE #1375: Semaphore imbalance when returned promise resolves when test timed out.

* **QUnit version**:2.9.1
* **What environment are you running QUnit in? (e.g., browser, Node)**: Browser & Node
* **How are you running QUnit? (e.g., script, testem, Grunt)**: `qunit` Node CLI / `testem`

## What are you trying to do?

A test returning a promise (or an `async`) that runs longer than the allowed timeout period, causes an internal test semaphore imbalance if the returned promise ever settles.

```js
QUnit.module('group of tests', function(hooks) {
  QUnit.test('one', function(assert) {
	assert.timeout(10);

    return new Promise(resolve => setTimeout(resolve, 20));
  });

  QUnit.test('two', function(assert) {
   	return new Promise(resolve => setTimeout(resolve, 20))
      .then(() => {
        assert.ok(true, 'all is well');
      });
  });
});
```

[CodePen Demo Here](https://codepen.io/rwjblue/pen/GzLqXa)

Test `one` should fail (due to timeout), and test `two` should pass.

## What actually happened?

Test `one` fails properly (due to timeout), however test `two` _also_ fails with the following message:

```
Tried to restart test while already started (test's semaphore was 0 already)
```

scalvert:
Fixes #1375

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -728,6 +728,7 @@ export function only( testName, callback ) {
 
 // Put a hold on processing and return a function that will release it.
 export function internalStop( test ) {
+	let released = false;
 	test.semaphore += 1;
 	config.blocking = true;
 
@@ -748,13 +749,13 @@ export function internalStop( test ) {
 					`Test took longer than ${timeoutDuration}ms; test timed out.`,
 					sourceFromStacktrace( 2 )
 				);
+				released = true;
 				internalRecover( test );
 			}, timeoutDuration );
 		}
 
 	}
 
-	let released = false;
 	return function resume() {
 		if ( released ) {
 			return;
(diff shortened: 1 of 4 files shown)


stefanpenner:
restarting the 2 failed jobs @scalvert you may need to investigate the failures if the reoccur.

scalvert:
Fixed. The regex for the tap output needed to be loosened to account for different stacks in versions of node.

stefanpenner:
I don't think I can release, so other maintainers will have to help with that yet.

trentmwillis:
Thanks! I'll cut a new release tomorrow.

scalvert:
Thanks, @trentmwillis! This one has been plaguing us for a while!

PR #1378, opened 2019-02-22
Core: Added guard for if module.hooks[<type>] is undefined

### LINKED ISSUE #1377: Core: hooks invoked after a test is considered torn down can cause error

* **QUnit version**: 2.9.1
* **What environment are you running QUnit in? (e.g., browser, Node)**:* Browser & Node
* **How are you running QUnit? (e.g., script, testem, Grunt)**: * `qunit` Node CLI / `testem`

## What are you trying to do?

A test whose done hooks are attempted to be invoked after the test is considered torn down can produce an error when trying to access those hooks.

```js
QUnit.test( "hello test", function( assert ) {
  let done = assert.async()
  assert.ok( 1 == "1", "Passed!" );
  setTimeout(done,10);
  return Promise.reject();
});
```

Will result in:

```
Uncaught Error: assert.async callback called after test finished.
    at done (qunit-2.9.1.js:3773)
qunit-2.9.1.js:3107 Uncaught TypeError: Cannot read property 'length' of undefined
    at processHooks (qunit-2.9.1.js:3107)
    at Test.hooks (qunit-2.9.1.js:3116)
    at runTest (qunit-2.9.1.js:3254)
    at advanceTestQueue (qunit-2.9.1.js:2646)
    at Object.advance (qunit-2.9.1.js:2597)
    at Test.queue (qunit-2.9.1.js:3276)
    at test (qunit-2.9.1.js:3516)
    at Object.onError (qunit-2.9.1.js:4217)
    at window$1.onerror (qunit-2.9.1.js:5499)
```

Codepen: https://codepen.io/rwjblue/pen/KJLQEJ

Trying to access hooks on an already torn down test should not result in an error.

## What actually happened?

An [error was thrown when trying to access hooks on a module](https://github.com/qunitjs/qunit/blob/master/src/test.js#L245). At this point, `hooks` has already been torn down.

scalvert:
Fixes #1377

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -242,9 +242,11 @@ Test.prototype = {
 				processHooks( test, module.parentModule );
 			}
 
-			if ( module.hooks[ handler ].length ) {
-				for ( let i = 0; i < module.hooks[ handler ].length; i++ ) {
-					hooks.push( test.queueHook( module.hooks[ handler ][ i ], handler, module ) );
+			const hooksForHandler = module.hooks[ handler ];
+
+			if ( hooksForHandler && hooksForHandler.length ) {
+				for ( let i = 0; i < hooksForHandler.length; i++ ) {
+					hooks.push( test.queueHook( hooksForHandler[ i ], handler, module ) );
 				}
 			}
 		}
--- a/test/cli/fixtures/undefined-hooks/index.js
+++ b/test/cli/fixtures/undefined-hooks/index.js
@@ -1,0 +1,6 @@
+QUnit.test( "undefined hooks", function( assert ) {
+	const done = assert.async();
+	assert.ok( 1 == "1", "Passed!" ); // eslint-disable-line eqeqeq
+	setTimeout( done, 10 );
+	return Promise.reject();
+} );
--- a/test/cli/fixtures/undefined-hooks/index.js
+++ b/test/cli/fixtures/undefined-hooks/index.js
@@ -1,6 +1,4 @@
-QUnit.test( "undefined hooks", function( assert ) {
-	const done = assert.async();
-	assert.ok( 1 == "1", "Passed!" ); // eslint-disable-line eqeqeq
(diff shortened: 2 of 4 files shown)


scalvert:
Looking at the CI failure in node 8 now.

test/cli/fixtures/undefined-hooks/index.js
```diff
@@ -0,0 +1,6 @@
+QUnit.test( "undefined hooks", function( assert ) {
```
trentmwillis: As noted in the related issue (https://github.com/qunitjs/qunit/issues/1377#issuecomment-466259122), this reproduction can be smaller.

Additionally, this test description is pretty vague. I would prefer to see something to the effect of "assertion after errored/rejected tests does not cause further errors".

src/test.js
```diff
@@ -242,9 +242,11 @@ Test.prototype = {
 				processHooks( test, module.parentModule );
 			}
 
-			if ( module.hooks[ handler ].length ) {
-				for ( let i = 0; i < module.hooks[ handler ].length; i++ ) {
-					hooks.push( test.queueHook( module.hooks[ handler ][ i ], handler, module ) );
+			const hooksForHandler = module.hooks[ handler ];
+
+			if ( hooksForHandler && hooksForHandler.length ) {
```
trentmwillis: This will solve the current issue, but ideally, these hooks should never be getting invoked when `hooksForHandler` is not defined. I'd like to see if we can find a fix for that (which I would consider the root issue here).

test/cli/fixtures/undefined-hooks/index.js
scalvert: This seems much better. Will update.

src/test.js
scalvert: Ya I agree. This is a bit of a stopgap, but was the most visible issue that was addressable. Agree that further investigation is warranted.

PR #1379, opened 2019-02-25
Doc: deepEqual() checks strict inherited equality

niedzielski:
Via comments on
https://gerrit.wikimedia.org/r/#/c/mediawiki/extensions/Popups/+/488977/1//COMMIT_MSG@10.


Krinkle:
Thanks @niedzielski

PR #1382, opened 2019-03-12
Use `console.group/groupEnd()` to group console messages during test runs

Turbo87:
screenshot

When running multiple tests in the browser it can be hard to figure out what tests are causing warnings or other debug messages. The `console.group()` API allows console output to be grouped to make it easier to see what those messages belong to.

In this PR I've added `console.group/groupEnd()` calls to the test/test suite start/end handlers. This makes it a lot easier to see what console output belongs to what test.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,9 @@
     "parserOptions": {
         "sourceType": "module"
     },
+    "globals": {
+        "console": true
+    },
     "rules": {
         "arrow-spacing": ["error", { "before": true, "after": true }],
         "brace-style": ["error", "1tbs", { "allowSingleLine": true }],
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -7,6 +7,13 @@ export const now = Date.now || function() {
 	return new Date().getTime();
 };
 
+export const isNodeJS = detectNodeJS();
+
+function detectNodeJS() {
+	// eslint-disable-next-line no-undef
+	return typeof process !== "undefined" && process.release.name === "node";
+}
+
 export const hasPerformanceApi = detectPerformanceApi();
 export const performance = hasPerformanceApi ? window.performance : undefined;
 export const performanceNow = hasPerformanceApi ?
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -10,8 +10,10 @@ export const now = Date.now || function() {
 export const isNodeJS = detectNodeJS();
 
 function detectNodeJS() {
-	// eslint-disable-next-line no-undef
-	return typeof process !== "undefined" && process.release.name === "node";
+	/* eslint-disable no-undef */
+	return typeof process !== "undefined" &&
+		process.release !== undefined &&
+		process.release.name === "node";
 }
 
 export const hasPerformanceApi = detectPerformanceApi();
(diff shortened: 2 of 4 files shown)


Turbo87:
I'm currently doing the same thing manually in a few projects, like this:

```js
QUnit.on('suiteStart', function({ name }) {
  console.group(`Test Suite: ${name}`);
});

QUnit.on('suiteEnd', function() {
  console.groupEnd();
});

QUnit.on('testStart', function({ name }) {
  console.group(`Test: ${name}`);
});

QUnit.on('testEnd', function() {
  console.groupEnd();
});
```

but it would be a lot nicer if this would just be the default :)

.eslintrc.json
```diff
@@ -7,6 +7,9 @@
     "parserOptions": {
         "sourceType": "module"
     },
+    "globals": {
+        "console": true
```
rwjblue: Hmm, do we want to allow `console` everywhere or just in the specific places we expect to use it?

src/core/utilities.js
```diff
@@ -7,6 +7,13 @@ export const now = Date.now || function() {
 	return new Date().getTime();
 };
 
+export const isNodeJS = detectNodeJS();
+
+function detectNodeJS() {
+	// eslint-disable-next-line no-undef
+	return typeof process !== "undefined" && process.release.name === "node";
```
rwjblue:
```suggestion
	return typeof process !== "undefined" && process.release !== undefined && process.release.name === "node";
```

.eslintrc.json
Turbo87: I can also add `/* global console */` in those two files if you think it's better?

Turbo87:
@trentmwillis @Krinkle any thoughts on this?

Krinkle:
I've often found that heavy use of `console` can cause tests to run significantly slower than they otherwise would. I'm also concerned about what this will do when QUnit is used via Karma.

With Karma, the progress is reported from the browser process to a Node.js process via web sockets. In addition to that, any console messages and errors etc are also printed to the Node.js process output.

This would effectively cause two different reports to be live at the same time (the HTML reporter, and this "browser console" partial reporter).

I do very much like the idea of grouping the console output by test (if and when a test already causes output to be written to the console). But perhaps we can find a way to make not call into Console when there is none? I realise this is probably non-trivial, if at all possible.

If not possible, we may want to accommodate this in another way. E.g. as an actual reporter that you can optionally enable in addition to (or instead of) the HTML reporter.

Turbo87:
@Krinkle I can't say anything about Karma, but with `testem` (as used by Ember CLI) it worked great. `testem` just ignored the group calls and only recorded the actual logs.

gibson042:
+1 @Krinkle
I haven't had time for this project in quite a while, but it was always my intention (and I thought also a shared one) for reporters to be independent consumers of test results. An HTML reporter should not be invoking console methods in my opinion, but _should_ be able to coexist with a console reporter that does so.

step2yeung:
@Turbo87 as of testem 2.15.0, testem no longer ignores group calls:
https://github.com/testem/testem/commit/f1b637f85d45c173bd23a67711068bb8cd96bfff

Turbo87:
for anyone interested, I packaged this up as a plugin https://github.com/simplabs/qunit-console-grouper

for any Ember.js users: it's also an Ember addon, so you just need to add it as a dependency and then you're ready to go.

PR #1384, opened 2019-03-14
Ignore all those folders mentioned in the gitignore.

SparshithNR:
1) ignores node_modules and .git folder irrespective of list in .gitignore file
2) gitignored files/folders can be conviniently ignored.


SparshithNR:
For context about how much time this saves,
`node_modules/.bin/qunit tests/*test.js  10.60s user 16.44s system 101% cpu 26.734 total`
`node node_modules/qunit/bin/qunit tests/*test.js  0.71s user 0.11s system 109% cpu 0.750 total`

Krinkle:
@SparshithNR Thanks for this patch.

Take a look at `src/cli/run.js` which has a `IGNORED_GLOBS` constant being used for a similar purpose. Could you see if it's possible to incorporate this improvement in that file instead?

SparshithNR:
@Krinkle We can completely remove that from `src/cli/run.js` because it calls [findFiles](https://github.com/qunitjs/qunit/blob/master/src/cli/run.js#L109) which is where we have our `IGNORED_GLOBS` getting merged with ignored list passed by the caller. We can keep everything in `utils.js` than in `src/cli/run.js` as `IGNORED_GLOBS` is used in `utils.js`.

Krinkle:
@SparshithNR I understand, but `findFiles()` is written and maintained as a generic utility. This ignore information is about how it is used in `cli.js`, so if possible, I think we should move the logic there.

SparshithNR:
That makes sense. I will update the PR accordingly.

SparshithNR:
Made new changes. Everything moved to `run` now.

SparshithNR:
Please review this @Krinkle

src/cli/run.js
```diff
@@ -18,6 +20,11 @@ function run( args, options ) {
 	// Default to non-zero exit code to avoid false positives
 	process.exitCode = 1;
 
+	// Do not push content of .gitignore everytime when run is called
+	if ( IGNORED_GLOBS.length === 2 ) {
+		IGNORED_GLOBS.push.apply( IGNORED_GLOBS, utils.getIgnoreList( process.cwd() ) );
```
trentmwillis: Could we just do this when the module loads? E.g.:

```javascript
const IGNORED_GLOBS = [
  ".git",
  "node_modules"
].concat( utils.getIgnoreList( process.cwd() ) );
```

Then we can avoid hardcoding the length check.

src/cli/utils.js
```diff
@@ -12,6 +12,16 @@ function existsStat() {
 	}
 }
 
+
+function getIgnoreList( baseDir ) {
+	if ( fs.existsSync( `${baseDir}/.gitignore` ) ) {
+		let gitIgnore = fs.readFileSync( `${baseDir}/.gitignore`, "utf-8" );
+		gitIgnore = gitIgnore.trim();
+		return gitIgnore.split( "\n" );
```
trentmwillis: Can we avoid the reassignment by just doing:

```javascript
const gitIgnore = fs.readFileSync( `${baseDir}/.gitignore`, "utf-8" );
return gitIgnore.trim().split( "\n" );
```

src/cli/run.js
SparshithNR: As I mentioned in the comment above the check, I was thinking if the `run` is called more than once we will be pushing `".git",  "node_modules"` multiple time. If it is not a valid case I am fine with doing this. Just that `IGNORED_GLOBS` need to be global const as it is used in `run.restart functions` to ignore files.

src/cli/utils.js
SparshithNR: I will make this change.

src/cli/run.js
SparshithNR: And `push` is much faster than `concat`

trentmwillis: My suggestion is to do this outside of `run` (i.e., on line 9) so it'll only ever possibly run once.

And push vs. concat I don't particularly care about perf since we'll only be calling it once the different should be negligible.

SparshithNR: Sorry for the delayed update. I committed new changes as per review comments.

test/cli/fixtures/.gitignore
```diff
@@ -0,0 +1,2 @@
+/abcd
+/efgh
```
Krinkle: Include the standard line break at the end of the file :)

PR #1391, opened 2019-04-25
Core: ensure process-queue advance with rejected promises

step2yeung:
Problem: When the promise that processTaskQueue is rejected, the processing-queue does not continue to process the rest of the queue and will hang.

Fix: Make sure we `advance()` the processing-queue with `.finally`


src/core/processing-queue.js
```diff
@@ -62,12 +62,10 @@ function processTaskQueue( start ) {
...
 			const task = taskQueue.shift();
 			Promise.resolve( task() ).then( function() {
-				if ( !taskQueue.length ) {
-					advance();
-				} else {
+				if ( taskQueue.length > 0 ) {
 					processTaskQueue( start );
 				}
-			} );
+			} ).finally( advance );
```
rwjblue: Can we rely on `finally` here (I don't _think_ so, but I could be wrong)?

reporter/html.js
```diff
@@ -858,16 +858,15 @@ export function escapeText( s ) {
 	} );
 
 	QUnit.testDone( function( details ) {
-		var testTitle, time, testItem, assertList, status,
+		var testTitle, time, assertList, status,
```
arthirm: Nit: You want to change this `var` to `let` as well.

src/core/processing-queue.js
```diff
@@ -61,13 +61,19 @@ function processTaskQueue( start ) {
...
 					processTaskQueue( start );
 				}
-			} );
+			};
+			const throwAndAdvance = ( err ) => {
+				setTimeout( advance );
+				throw err;
+			};
+
+			Promise.resolve( task() ).then( processNextTaskOrAdvance, throwAndAdvance );
```
arthirm: You want to add some comments here on what will happen if the rejected promise is not handled?

src/core/logging.js
```diff
@@ -1,41 +1,73 @@
...
+	"moduleStart",
+	"moduleDone"
+];
+
+function _promisfyCallbacksSequentially( callbacks, args ) {
+	return callbacks.reduce( ( promiseChain, callback ) => {
+		return promiseChain.then(
+			() => callback( args ),
+			( err ) => {
+				setTimeout( callback( args ) );
```
mikrostew: shouldn't this take a function? like

```javascript
setTimeout( () => callback( args ) );
```

src/core/processing-queue.js
```diff
@@ -182,29 +177,46 @@ function done() {
...
+ * This function is called when the ProcessingQueue is done processing all
+ * items. It handles emitting the final run events.
+ */
+function done() {
+	ProcessingQueue.finished = true;
+
+	try {
+		throwErrorIfNoTestExecuted();
+	} catch ( err ) {
+		throw err;
```
mikrostew: I don't understand why you're catching the error just to re-throw it here? Maybe a comment about why this is needed

src/test.js
```diff
@@ -113,24 +111,30 @@ function getNotStartedModules( startModule ) {
...
 				startModule.stats = { all: 0, bad: 0, started: now() };
 				emit( "suiteStart", startModule.suiteReport.start( true ) );
 				return runLoggingCallbacks( "moduleStart", {
 					name: startModule.name,
 					tests: startModule.tests
 				} );
-			} );
+			};
+
+			return promiseChain.then( moduleStartCallback, moduleStartCallback );
```
mikrostew: Looks like this will swallow any errors at this point - is that the desired behavior here?

src/core/processing-queue.js
step2yeung: I don't understand what I was thinking either, totally unnecessary :)

src/core/processing-queue.js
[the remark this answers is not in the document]
step2yeung: Would love to understand more. I tried a small snippet to compare the difference:
```
function throwErr() {
    throw new Error('err');
}

Promise.resolve(throwErr());
    Uncaught Error: err
    at throwErr (<anonymous>:2:11)
    at <anonymous>:1:17

new Promise(resolve => resolve(throwErr()));
    Uncaught (in promise) Error: err
    at throwErr (<anonymous>:2:11)
    at <anonymous>:1:32
    at new Promise (<anonymous>)
    at <anonymous>:1:1
```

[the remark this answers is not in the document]
step2yeung: I would expect an error to be thrown and continue advancing the queue.

src/test.js
step2yeung: @mikrostew I assuming you are talking about `const callbackPromises = notStartedModules.reduce(...);` ?

if any of the promise from callbackPromises reject.. the whole promise should reject.. which should be handled on line 154:
```
return callbackPromises.then( testStartHandler, ( err ) => {
		return Promise.reject( err ).catch( ( err ) => {
			setTimeout( testStartHandler );
			throw err;
		} );
	} );
```

rwjblue:
@step2yeung - Would it be possible to split this into two changes (one for the actual bug being fixed along with its test, and another for the refactorings)?

src/test.js
step2yeung: Read this again, this shouldn't be a problem, but I can understand why it looks like the error is swallowed.
For a bit more context.. `const callbackPromises = notStartedModules.reduce(...);` ensures the callbacks from notStartedModules are promisfied sequentually, so each callback that was made into a promise, an awaited for until the next callback is executed and made into a promise.
so You will get an chain of promises. In that chain, if there is one rejected promise, then the whole promise chain is rejected. 

`return promiseChain.then( moduleStartCallback, moduleStartCallback );` in this case, only ensures the next callback is promisfied even if the previous promise rejects. But as stated above, lien 154 will ensure the error is thrown.

I hope that makes more sense.. I think I should likely find a better way to code the explanation above up

step2yeung:
@stefanpenner Updated the description, hope it helps clarify the user facing issue. 

@rwjblue Separated out the fix and the refactor (will submit that afterwards)

src/core/logging.js
```diff
@@ -1,27 +1,42 @@
...
+	setTimeout
+} from "../globals";
+
+function _promisfyCallbacksSequentially( callbacks, args ) {
+	return callbacks.reduce( ( promiseChain, callback ) => {
+		const executeCallback = () => callback( args );
+		return promiseChain.then(
+			executeCallback,
+			( err ) => {
+				setTimeout( executeCallback );
```
Krinkle, line 14: Use of `setTimeout` must be conditional for non-browser environments such as SpiderMonkey (see other references). An alternate approach here can be to use `Promise.resolve()` as fallback for `setTimeout`, or to avoid this alltogether by using `Promise.finally()` for the executeCallback, which naturally forwards the original rejection/error.

src/core/logging.js
```diff
@@ -1,27 +1,42 @@
...
+				throw err;
```
Krinkle, line 15: I suspect that this particular thenable chain might be better without a local error handler as otherwise it means that the last error will be reported instead of the first one. I general QUnit's test logic is modelled to continue after assertion failures, but for uncaught exceptions I think we generally want to stop after the first for any given sequence.

Krinkle:
You may ignore these inline comments. I've uncovered a couple of other things that I'll summarise at https://github.com/qunitjs/qunit/issues/1446.

reporter/html.js
```diff
@@ -858,16 +858,15 @@ export function escapeText( s ) {
...
 			good, bad, testCounts, skipped, sourceName,
-			tests = id( "qunit-tests" );
+			tests = id( "qunit-tests" ),
+			testItem = id( "qunit-test-output-" + details.testId );
 
-		if ( !tests ) {
+		if ( !tests || !testItem ) {
```
Krinkle, line 866: I've extracted and applied this, with credit, via https://github.com/qunitjs/qunit/pull/1562 (released since then).

src/core.js
```diff
@@ -176,7 +176,10 @@ export function begin() {
 		runLoggingCallbacks( "begin", {
 			totalTests: Test.count,
 			modules: modulesLog
-		} ).then( unblockAndAdvanceQueue );
+		} ).then( unblockAndAdvanceQueue, function( err ) {
+			setTimeout( unblockAndAdvanceQueue );
```
Krinkle, line 180: I've added tests for this scenario in https://github.com/qunitjs/qunit/pull/1634, and addressed the bug this fixes (https://github.com/qunitjs/qunit/issues/1446) via a different approach in https://github.com/qunitjs/qunit/pull/1638. In a nutshell, we handle this via the `unhandledrejection` event in the browser and Node.js, same as before, but previously that didn't work right due to it creating a fake test which we no longer do.

Krinkle:
The one part I believe we still haven't fixed is advancing the process-queue. It makes sense to advance the queue, but, I'd also like to make sure (to avoid future regressions) that the negative impact of this is quantified in an integration test.

For the QUnit CLI (TAP) this is probably not particularly significant as Node.js processes naturally end at this point, and the error is reported with TAP-formatted bail out. For browsers, I imagine the process would remain stuck pending timeout, exactly as before, as we'd not proceed and thus not reach the `runEnd` event or `done()` callback.

PR #1395, opened 2019-05-24
CSS: Tweak CSS to address dropdown and button styles

mfeckie:
The tested site can override the styling of buttons, which causes the
displayed content to 'jump' around while tests are running, so ensure
that all buttons in qunit elements are set to inital font size, border
and background.

With modules that have long names, the filter list becomes un-useable,
so allow them to grow using min-width rather than fixed width

### DIFF
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -11,6 +11,13 @@
 
 /** Font Family and Sizes */
 
+
+[id^=qunit] button {
+	font-size: initial;
+	border: initial;
+	background-color: buttonface;
+}
+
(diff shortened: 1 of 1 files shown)


rwjblue:
Seems like a nice improvement to me.

src/qunit.css
```diff
@@ -11,6 +11,13 @@
 
 /** Font Family and Sizes */
 
+
+[id^=qunit] button {
+	font-size: initial;
+	border: initial;
+	background-color: buttonface;
```
trentmwillis: Was not familiar with this! Cool, TIL.

mfeckie: ❤️

stefanpenner: this is an amazing color :P

PR #1400, opened 2019-06-24
Core: reset testTimeout on assert.timeout if config.timeout has already been set

step2yeung:
This reset any current timeouts by canceling and setting up a new timer.

Usecase: https://github.com/emberjs/ember-qunit/issues/496


trentmwillis:
Sorry for the delay in reviewing this; on first glance it looks good! I'll try to do a proper review this weekend.

PR #1401, opened 2019-06-27
Core: Defer getting the test source until we need it

misterbyrne:
We have a very large test suite (11K tests) and I noticed that a lot of time was spent generating stack traces when defining the tests on loading the page. This PR still captures the source of the stack trace, but defers generating the stack trace itself until we need it.

For our test suite, this reduced the time spent in the `QUnit.test` function by about 2 seconds, which makes a big difference to the developer feedback loop.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -34,9 +34,9 @@ export default function Test( settings ) {
 	this.assertions = [];
 	this.semaphore = 0;
 	this.module = config.currentModule;
-	this.stack = sourceFromStacktrace( 3 );
 	this.steps = [];
 	this.timeout = undefined;
+	this.errorForStack = new Error();
 
 	// If a module is skipped, all its tests and the tests of the child suites
 	// should be treated as skipped even if they are defined as `only` or `todo`.
@@ -112,6 +112,12 @@ function getNotStartedModules( startModule ) {
 }
 
 Test.prototype = {
+
+	// generating a stack trace can be expensive, so using a getter defers this until we need it
+	get stack() {
+		return extractStacktrace( this.errorForStack, 2 );
+	},
+
 	before: function() {
 		var module = this.module,
 			notStartedModules = getNotStartedModules( module );
@@ -318,6 +324,7 @@ Test.prototype = {
 		// avoid leaking it. It is not used by the legacy testDone callbacks.
 		emit( "testEnd", this.testReport.end( true ) );
 		this.testReport.slimAssertions();
+		const test = this;
 
 		return runLoggingCallbacks( "testDone", {
 			name: testName,
@@ -334,7 +341,8 @@ Test.prototype = {
 			testId: this.testId,
 
 			// Source of Test
-			source: this.stack
+			// generating stack trace is expensive, so using a getter will help defer this until we need it
+			get source() { return test.stack; }
 		} ).then( function() {
 			if ( module.testsRun === numberOfTests( module ) ) {
 				const completedModules = [ module ];


src/test.js
```diff
@@ -34,10 +34,15 @@ export default function Test( settings ) {
 	this.assertions = [];
 	this.semaphore = 0;
 	this.module = config.currentModule;
-	this.stack = sourceFromStacktrace( 3 );
 	this.steps = [];
 	this.timeout = undefined;
 
+	// evaluating error.stack can be expensive, so we should defer this if possible
+	var errorForStack = new Error();
+	this.getStack = function() {
```
rwjblue: Have you considered using a getter here instead of adding a `.getStack()` method?

```js
// here
this._error = new Error();
this._stack = null;

// lower in the class body
get stack () {
  if (this._stack) { return this._stack; }

  return this._stack = extractStacktrace( errorForStack, 2 );
}
```

This would have the benefit of not actually changing the expected shape/api of the `Test` instance. Not sure we _really_ care about preserving the API, but it seems vaguely easier...

src/test.js
```diff
@@ -334,7 +339,7 @@ Test.prototype = {
 			testId: this.testId,
 
 			// Source of Test
-			source: this.stack
+			source: this.getStack()
```
rwjblue: Hmm, do we need the source of the test here? Seems like this particular usage of the stack is not conditional and would ultimately force all tests to generate the stack even if it were unused.

Thoughts on:

```js
return runLoggingCallbacks('testDone', {
  //...snip...
  get source() { return this.getStack(); }
});
```

That way if the source _wasn't_ used by the logging callbacks, the stack would never have to be reified...

stefanpenner:
This works in chrome, does this work in safari/firefox as well?

src/test.js
scalvert: 👍 to this suggestion. This is how we've handled deferring stack prop access in other projects.

stefanpenner:
Good idea!

src/test.js
misterbyrne: That’s what I first thought of but I figured this was more compatible with old skool browsers. If that’s not an issue (because there’s some minimum version requirement or some compilation step that sorts this out) then it makes sense to implement this as a getter. Happy to take direction here.

rwjblue: We do transpile with Babel, but QUnit also [currently supports](https://qunitjs.com/) the browsers [supported by jQuery 3.x](https://jquery.com/browser-support/), the oldest of which is IE9. 

[IE9 supports defining getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#Browser_compatibility).

misterbyrne:
Thanks! I've changed it to use a getter - much cleaner 😄

rwjblue:
Looks awesome, thank you @misterbyrne!

src/test.js
```diff
@@ -334,7 +341,8 @@ Test.prototype = {
 			testId: this.testId,
 
 			// Source of Test
-			source: this.stack
+			// generating stack trace is expensive, so using a getter will help defer this until we need it
+			get source() { return test.stack; }
```
misterbyrne: I was incorrectly using `this` before. Anyway, should be passing now (and I've probably invalidated your approval with that last push 😬).

misterbyrne:
👋Anything else I need to do here?

misterbyrne:
Yes is seems to be tested, although indirectly (in the `reporter-html` test.

I changed the `stack` getter to always return a static string to see what tests would fail, and this one did:

https://github.com/qunitjs/qunit/blob/cc1fdb62423cddbf77d51b5bf33dd47126a82d94/test/reporter-html/reporter-html.js#L146

patocallaghan:
@platinumazure Any idea when this will merged and released?

patocallaghan:
@rwjblue @scalvert @stefanpenner any chance we could get this change merged? it would make a big difference to the reload time of our test suite.

platinumazure:
I'd like at least one more review from a team member before merging. To that end, I have request reviews from two other team members (I'll be happy to merge with just one more).

I apologize for the lengthy delay in merging this.

ro0gr:
Awesome! This has sped up our test suite on page reload even with ~1k tests 🚀

PR #1408, opened 2019-10-05
Docs: rename "Main assertions" page to "Main Methods"

### LINKED ISSUE #1396: Consider renaming "Main assertions" page
Almost every time I navigate to <https://api.qunitjs.com/> I find myself clicking "Main assertions" when looking up one of the assert methods, like `propEqual` or some such.

Then after a brief moment of confusing and a failed attempt at expanding the page further, I realise I should have clicked "Assertions" instead.

I don't have any good ideas off the top of my head, but we may want to consider renaming this page.

trentmwillis:
Fixes #1396

Also fixes some discrepancies between the sidebar and the top navigation.

### DIFF
--- a/docs/QUnit/index.md
+++ b/docs/QUnit/index.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Main assertions
+title: Main Methods
 category: main
 categories:
   - topics
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -3,7 +3,7 @@ layout: wrapper
 ---
 
 <section class="sidebar">
-    <h2>Main methods:</h2>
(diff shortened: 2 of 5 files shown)


trentmwillis:
I used title case because they're the titles of the various pages/sections. That said, I can change it if you feel it helps usability in distinguishing the nav items. At some point, I'd like to invest in the overall design of the API docs, but likely won't have time for that for a while.

Krinkle:
Yeah, might just be the Wikipedian in me having an aversion to the classical Title Case. I don't can't think of a rational reason other than (subjective) tradition and indeed the navigation ambiguity, but it's pretty small. Let's just go with this for now :)

PR #1411, opened 2019-10-05
Docs: Update Gemfile.lock

Krinkle:
This only affects local dev since for the live site, GitHub
ignores the Gemfile (it has its own Jekyll install).
This basically just regenerates it to match whatever GitHub
has published as RubyGem.

Fixes two security warnings:
* nokogiri 1.8.5 < 1.10.4
* rubyzip 1.2.2 < 1.3.0


platinumazure:
Looks good outside of build failures. Are those expected?

Krinkle:
Haven't seen that macOS failure before. Seems flaky indeed. Will re-try.

PR #1417, opened 2019-11-21
Add testsNotRan to keep count of testsNotRan to ensure moduleDone is called

### LINKED ISSUE #1416: Mismatched "suiteEnd" event when a filter is applied

* **QUnit version**: 2.9.4-pre
* **How are you running QUnit? (e.g., script, testem, Grunt)**: grunt

## What are you trying to do?

1. `git clone qunit` & `yarn dev`
2. Add the following code to `dist/qunit.js`
  a. console.logs after `function runLoggingCallbacks(key, args) {`
```
function runLoggingCallbacks(key, args) {
    console.log('runLoggingCallbacks called for '+key)
    var callbacks = config.callbacks[key];
```
  b. console.logs before `if (module.testsRun === numberOfTests(module)) {`
```
}).then(function () {
    console.log('module.testsRun === numberOfTests(module)');
    console.log(module.testsRun +' === ' +numberOfTests(module))
    if (module.testsRun === numberOfTests(module)) {
        var completedModules = [module];
```
3. Point a browser to `http://localhost:8000/test/string-filter.html`
4. Open up browser console

Expected `moduleDone` callback (from the `suiteEnd` event) to be called.

## What actually happened?

`moduleDone` was not called because [module.testsRun does not equal numberOfTests(module)](https://github.com/qunitjs/qunit/blob/5bea87b0f2974b49e8bf907aa2b197ee469fe99b/src/test.js#L347)
1 of the 3 test was filtered out and not ran. Leaving only 2 test to be ran. 
So `module.testsRun = 2` while  `numberOfTests(module) = 3`

Screen Shot 2019-11-19 at 9 09 58 AM

Regardless, I believe `moduleDone` should still be called.

step2yeung:
Fixes #1416 

This PR adds `testsNotRan` to a module. This is used to keep track of any tests where `isValid()` returns false.
This attribute will be used in this condition: 
```
if ( ( module.testsRun + module.testsNotRan ) === numberOfTests( module ) ) {
```
checking for whether the module has completed running its tests.

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -46,7 +46,7 @@ const config = {
 		tests: [],
 		childModules: [],
 		testsRun: 0,
-		unskippedTestsRun: 0,
+		testsIgnored: 0,
 		hooks: {
 			before: [],
 			beforeEach: [],
--- a/src/test.js
+++ b/src/test.js
@@ -208,7 +208,7 @@ Test.prototype = {
 
 		const runHook = () => {
 			if ( hookName === "before" ) {
-				if ( hookOwner.unskippedTestsRun !== 0 ) {
+				if ( hookOwner.testsRun !== 0 ) {
 					return;
 				}
 
@@ -217,8 +217,7 @@ Test.prototype = {
 
 			// The 'after' hook should only execute when there are not tests left and
 			// when the 'after' and 'finish' tasks are the only tasks left to process
-			if ( hookName === "after" &&
-				hookOwner.unskippedTestsRun !== numberOfUnskippedTests( hookOwner ) - 1 &&
+			if ( hookName === "after" && !lastTestWithinModuleExecuted( hookOwner ) &&
 				( config.queue.length > 0 || ProcessingQueue.taskCount() > 2 ) ) {
 				return;
 			}
@@ -309,7 +308,11 @@ Test.prototype = {
 			}
 		}
 
-		notifyTestsRan( module, skipped );
+		if ( skipped ) {
+			incrementTestsIgnored( module );
+		} else {
+			incrementTestsRun( module );
+		}
 
 		// Store result when possible
 		if ( storage ) {
@@ -344,13 +347,13 @@ Test.prototype = {
 			// generating stack trace is expensive, so using a getter will help defer this until we need it
 			get source() { return test.stack; }
 		} ).then( function() {
-			if ( module.testsRun === numberOfTests( module ) ) {
+			if ( allTestsExecuted( module ) ) {
 				const completedModules = [ module ];
 
 				// Check if the parent modules, iteratively, are done. If that the case,
 				// we emit the `suiteEnd` event and trigger `moduleDone` callback.
 				let parent = module.parentModule;
-				while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+				while (	parent && allTestsExecuted( parent ) ) {
 					completedModules.push( parent );
 					parent = parent.parentModule;
 				}
@@ -394,6 +397,7 @@ Test.prototype = {
 		const test = this;
 
 		if ( !this.valid() ) {
+			incrementTestsIgnored( this.module );
 			return;
 		}
 
@@ -551,10 +555,7 @@ Test.prototype = {
 	},
 
 	valid: function() {
-		var filter = config.filter,
-			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
-			module = config.module && config.module.toLowerCase(),
-			fullName = ( this.module.name + ": " + this.testName );
+		var module = config.module && config.module.toLowerCase();
 
 		function moduleChainNameMatch( testModule ) {
 			var testModuleName = testModule.name ? testModule.name.toLowerCase() : null;
@@ -592,6 +593,14 @@ Test.prototype = {
 		if ( module && !moduleChainNameMatch( this.module ) ) {
 			return false;
 		}
+
+		return this.filterMatch();
+	},
+
+	filterMatch: function() {
+		var filter = config.filter,
+			regexFilter = /^(!?)\/([\w\W]*)\/(i?$)/.exec( filter ),
+			fullName = ( this.module.name + ": " + this.testName );
 
 		if ( !filter ) {
 			return true;
@@ -857,23 +866,24 @@ function collectTests( module ) {
 	return tests;
 }
 
-function numberOfTests( module ) {
-	return collectTests( module ).length;
+function allTestsExecuted( module ) {
+	return module.testsRun + module.testsIgnored === collectTests( module ).length;
 }
 
-function numberOfUnskippedTests( module ) {
-	return collectTests( module ).filter( test => !test.skip ).length;
+function lastTestWithinModuleExecuted( module ) {
+	return module.testsRun + module.testsIgnored === collectTests( module ).length - 1;
 }
 
-function notifyTestsRan( module, skipped ) {
+function incrementTestsRun( module ) {
 	module.testsRun++;
-	if ( !skipped ) {
-		module.unskippedTestsRun++;
-	}
 	while ( ( module = module.parentModule ) ) {
 		module.testsRun++;
-		if ( !skipped ) {
-			module.unskippedTestsRun++;
-		}
+	}
+}
+
+function incrementTestsIgnored( module ) {
+	module.testsIgnored++;
+	while ( ( module = module.parentModule ) ) {
+		module.testsIgnored++;
 	}
 }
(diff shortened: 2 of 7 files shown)


src/test.js
```diff
@@ -344,7 +344,7 @@ Test.prototype = {
 			// generating stack trace is expensive, so using a getter will help defer this until we need it
 			get source() { return test.stack; }
 		} ).then( function() {
-			if ( module.testsRun === numberOfTests( module ) ) {
+			if ( ( module.testsRun + module.testsNotRan ) === numberOfTests( module ) ) {
```
choheekim: do we also need to check for parent module as well (#353)?
`while ( parent && parent.testsRun + parent.testNotRan === numberOfTests( parent ))`

step2yeung: @choheekim Do you have a concrete example to help me understand the question?
 I am not sure if I can visualize the problem yet

step2yeung: Talked to @choheekim offline, she pointed out the nested modules scenario, which pointed out something missing in the solution. Revised it to include this usecase.

src/test.js
```diff
@@ -344,13 +344,15 @@ Test.prototype = {
...
 				const completedModules = [ module ];
 
 				// Check if the parent modules, iteratively, are done. If that the case,
 				// we emit the `suiteEnd` event and trigger `moduleDone` callback.
 				let parent = module.parentModule;
-				while ( parent && parent.testsRun === numberOfTests( parent ) ) {
+				while (	parent &&
+					( parent.testsRun + parent.testsNotRan ) === numberOfTests( parent )
+				) {
```
scalvert: Do you think it's odd to signal that this module is completed when it really hasn't? I understand the reasoning, but it feels like you may be giving folks a false sense of security. I also understand the intent is to notify that some tests weren't run. Do we think this will mitigate the concern here?

src/test.js
```diff
@@ -877,3 +885,10 @@ function notifyTestsRan( module, skipped ) {
 		}
 	}
 }
+
+function notifyTestsNotRan( module ) {
```
scalvert: Should this function have a more accurate name? It's not really notifying anything. It's accumulating the total count of tests not run, isn't it?

step2yeung: Changed to `updateTestsNotRan`

src/test.js
step2yeung: @scalvert this is a good question. 
It depends on what is the intent of `moduleDone`:
1) done executing all the tests listed in the module.
2) done executing all tests picked to be ran in the module.

I implemented with the latter understanding. 

From the qunit documentation:
https://api.qunitjs.com/callbacks/QUnit.moduleDone
`Register a callback to fire whenever a module ends. The callback can return a promise that will be waited for before the next callback is handled.`

rwjblue: I agree with @step2yeung's interpretation here, IMHO `moduleDone` should be called when the module is finished running tests (regardless if all tests were ran, some were skipped, some were filtered, etc). This should conceptually be not that different from the `hooks.after` callback timing.

rwjblue: My last comment made me wonder if we should be creating another mechanism for tracking the same thing. Why can't we use the same technique that we use to detect when to run `after` hooks for the module? Does it have the same issue (if yes, we should include that fix here...)?

step2yeung: @rwjblue Good call on the `after` hook, i found that it suffers from the same issue. Added this fix too now!

davecombs:
could 'testsNotRan' be changed to 'testsNotRun' to match 'testsRun'? It's inconsistent as it is now.

step2yeung:
@davecombs updated.
@scalvert @rwjblue any additional comments?

src/test.js
```diff
@@ -218,7 +218,8 @@ Test.prototype = {
 			// The 'after' hook should only execute when there are not tests left and
 			// when the 'after' and 'finish' tasks are the only tasks left to process
 			if ( hookName === "after" &&
-				hookOwner.unskippedTestsRun !== numberOfUnskippedTests( hookOwner ) - 1 &&
+				( hookOwner.unskippedTestsRun + hookOwner.testsNotRun ) !==
+					numberOfUnskippedTests( hookOwner ) - 1 &&
```
rwjblue: Should we still have this condition? Why wouldn't skipped tests be treated the same as ones that were filtered out (and therefore be included in `testsNotRun` directly)?

src/core/config.js
```diff
@@ -47,6 +47,7 @@ const config = {
 		childModules: [],
 		testsRun: 0,
 		unskippedTestsRun: 0,
+		testsNotRun: 0,
```
rwjblue: I'm finding this terminology to be quite difficult to parse. As far as I can tell, this is actually representing the tests that _should_ not be ran (not those that happen to have not ran yet). 

Assuming that I've understood the logic here, here are some possible other names:

* `testsToBeIgnored`
* `excludedTests`

Also, why do we need both `testsRun`, `unskippedTestsRun`, and `testsNotRun` (or whatever name we decide to add). Do we actually need to track all three of these separately? It seems like we should only have two concepts:

* the count of tests that we have actually ran (currently `testsRun`)
* the count of tests that we plan to ignore (currently `testsNotRun`, but hopefully we can find a better name)

In other words, skipped tests are just another form of test to be ignored (just the same as tests that are filtered out).

step2yeung: Renamed this `testsNotRun ` to `testsIgnored` and removed `unskippedTestsRun`.
Tests that are skipped or filtered out will be accounted for by `testsIgnored`.

src/test.js
step2yeung: This check was still needed to make sure we only run the `after` hook ON the last test of the module. (hence the `- 1` from `numberOfUnskippedTests( hookOwner ) - 1`

I have added an equivalent check with
```
function lastTestWithinModuleExecuted( module ) {
	return module.testsRun + module.testsIgnored === collectTests( module ).length - 1;
}
```

step2yeung:
@rwjblue care to take another look?
fyi @Krinkle

src/test.js
```diff
@@ -208,7 +208,7 @@ Test.prototype = {
 
 		const runHook = () => {
 			if ( hookName === "before" ) {
-				if ( hookOwner.unskippedTestsRun !== 0 ) {
+				if ( hookOwner.testsRun !== 0 ) {
```
Krinkle, line 211: I'm wrapping my head around the bug now and want to make sure I understand each part of the change.

I believe this line is not a change in behaviour but an update to keep the same behaviour, meaning that the value we used to keep in `unskippedTestsRun` is now effectively kept in `testsRun`, only incremented for tests that were "really" run. Is that right?

src/test.js
```diff
@@ -593,6 +594,14 @@ Test.prototype = {
 			return false;
 		}
 
+		return this.filterMatch();
+	},
+
+	filterMatch: function() {
```
Krinkle, line 600: I believe this is unrelated clean up, is that right? It's fine, but I just want to check if I may have missed something here that was meant to change behaviour (e.g. the order is not significant).

step2yeung: Correct, this was a clean up to split up the responsibility of the function. No behavior change.

src/test.js
step2yeung: Yep, thats the correct understanding!
And to make the reverse clear, f tests that were not run(eg. skip, or tests where `valid()` returns false) would be kept using `testsIgnored`

test/main/modules.js
```diff
@@ -81,14 +81,10 @@ QUnit.module( "after (skip)", {
 	}
 } );
 
-QUnit.test( "does not run after initial tests", function( assert ) {
```
Krinkle, line 84: I'm not sure that this is correct. Looking at it locally, this means the `after` hook never runs, which would be a regression. During the first/last actually running test, it previously checked if we're currently executing the last non-ignored test. Which meant that if the last registered test is skipped, the after hook will run correctly after the last actually run test.

However, that logic was removed as part of this patch, and thus we now get a similar bug but affecting the `after` hook rather than the `moduleDone` callback.

I think for this to work we need to continue to "look ahead" like it did before, so that it knows during this test that it will be the last.

Krinkle (changes requested):

test/main/modules.js
step2yeung: Agreed, good catch @Krinkle! Let me see what I can do

Krinkle: @step2yeung Friendly ping to check if you're able to find time for this. If not, I might give it a go next week for the next patch release. Thanks again!

step2yeung: Thanks for the reminder Krinkle! Will get back to it this week!

Krinkle: One more reminder 🙂. No problem at all, but I'll give this a look this Saturday/Sunday if I don't hear back. Thanks again.

PR #1422, opened 2020-01-08
Build: Bumps package.lock verstion to 2.9.4-pre

kltan:
This happens automatically when `npm install` runs.


kltan:
@Krinkle Can you look into this? This change is needed so that we don't get unintentional version bump whenever we run `npm install`.

PR #1424, opened 2020-01-08
Core: Create config.testIds to expose all available test ids

kltan:
What this does

Currently each testId is calculated as each test is executed.
Config.testIds pre-hashes all the loaded tests on demand and returns all
testIds.

How is this helpful

Currently test runners can get access to test modules via config.modules
and partition the modules based on some distribution strategy for running
the tests in parallel. This is fine but it can lead to unbalanced
distribution as some modules have a lot of tests and some modules are
testing integration (longer running) test instead of unit test.

Partitioning the tests by modules caused some instance of test runners
to run way longer than others as some heavier running modules might be
clumped together in the same instance.

To mitigate and minimize the unbalanced test runs, the proposed solution
is to partition based on each individual test with testId. The test
runners can now access testIds beforehand and can distribute those ids
with the testId url param. This will minimize the test execution time
descrepencies between each individual instance of the test runners.


smcclure15:
Does this pre-hashing match the existing unique hashes during the run?
The existing hash mechanism modifies clashing test-names (white whitespace suffixes) to get unique IDs.

I want to ensure these pre- and post-ids match up, and if there's any way to avoid hashing the same thing twice (we're pre-allocating, so maybe the latter can reuse that cache and not recalculate?).

kltan:
> Does this pre-hashing match the existing unique hashes during the run?
> The existing hash mechanism modifies clashing test-names (white whitespace suffixes) to get unique IDs.

You are right, the new code did not account for that. I am going to abstract out the hashing collision management to a util and fix the code.

> I want to ensure these pre- and post-ids match up, and if there's any way to avoid hashing the same thing twice (we're pre-allocating, so maybe the latter can reuse that cache and not recalculate?).

The hashing is incredibly fast, I had 10k tests and it returned the whole array of hashes immediately upon calling the getter. There's a storage tradeoff for caching for something that might not improve performance measurably. 

Also, QUnit.config.testIds only evaluates on demand for certain cases that need the testIds in advance. I would argue not to add caching complexities.

kltan:
@smcclure15 Thanks for the review. Here's a screenshot of it working.

image

kltan:
@Krinkle and @trentmwillis appreciate some feedback here.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "qunit",
-  "version": "2.9.3-pre",
```
Krinkle: Unrelated change

Krinkle: ^ Unrelated change.

src/core/config.js
```diff
@@ -58,7 +58,25 @@ const config = {
...
+	// Get the testIds for all tests, value is read only
+	get testIds() {
+		var i, j, ml, tl, testIds = [], testNames, testName;
+
+		// Loop through all modules and tests and generate testIds
+		for ( i = 0, ml = config.modules.length; i < ml; i++ ) {
+			testNames = [];
+			for ( j = 0, tl = config.modules[ i ].tests.length; j < tl; j++ ) {
+				testName = uniqueTestName( testNames,
+					config.modules[ i ].tests[ j ].name );
```
Krinkle: Ideally we'd avoid this duplication. But at least worth leaving a comment here to point to the authoratative place in test.js where the same code is duplicated from. And in the other direction as well.

kltan: @Krinkle Added comments in both places.

package-lock.json
kltan: Removed accidental check in.

src/core/config.js
```diff
@@ -58,7 +58,27 @@ const config = {
 	callbacks: {},
 
 	// The storage module to use for reordering tests
-	storage: localSessionStorage
+	storage: localSessionStorage,
+
+	// Get the testIds for all tests
+	// Simulates how test.js generates testIds by
+	// iteratively pushing testNames to get uniqueTestName
+	get testIds() {
```
Krinkle: @trentmwillis I have a vague sense that using a getter here may bite us long-term in terms of confusion and flexibility. What's your take on this? Would making this a regular function be easier to understand/document/maintain long-term? Drawback is that it introduces a non-serialisable property to `QUnit.config`. On the other hand, if config does get serialised, the value it would currently serialise is likely wrong by the time it is read.

Maybe that's a reason to mount it somewhere else entirely, e.g. on `currentModule`.

kltan: Just trying to get some traction here.

kltan: @trentmwillis Any thoughts? I'd love to see this submitted to help tests run faster in parallel.

trentmwillis: Sorry, I've been busy working through other OSS tickets. From a brief glance, I think I'm inclined to agree that it doesn't belong on `QUnit.config`. I'll review more thoroughly this Friday.

trentmwillis: So after reviewing more thoroughly, I think this should just be a normal property that is an array.

The `testId` for each test is already generated eagerly (i.e., when you call `QUnit.test`). The information is already available within the (undocumented) `QUnit.config.modules` property but would require some munging. So instead, we could set this up such that as each testId gets initialized, we push it into a new `QUnit.config.testIds` property.

src/core/config.js
```diff
@@ -1,5 +1,5 @@
 import { window, localSessionStorage } from "../globals";
```
captain-hoity-toity, line 1: Here are the recommended changes for the PR commit:

1. **Consider adding comments:** While the code is quite readable, adding comments to explain what certain parts of the code do can be beneficial. This is especially true for complex operations or when you're using less common JavaScript features.

2. **Error handling:** There doesn't seem to be any error handling in the code. Consider adding `try-catch` blocks or error handling functions to handle any potential errors.

3. **Use `let` and `const` instead of `var`:** It's recommended to use `let` and `const` for variable declaration instead of `var` to reduce the scope in which the variable lives and to prevent redeclaration of the same variable.

4. **Avoid too many nested loops:** The nested loops in the `get testIds()` function could potentially be refactored for better readability and performance. Consider using higher-order functions like `map` or `reduce` instead.

5. **Code formatting:** Consider using a code formatter like Prettier to ensure consistent code style and formatting. This includes consistent use of semicolons, indentation, and whitespace.

6. **Use meaningful variable names:** While `i`, `j`, `ml`, and `tl` may make sense in the context of the loop, they don't provide much information about what they represent. Consider using more descriptive variable names for better readability.

7. **Use destructuring:** When accessing multiple properties from an object, consider using destructuring. For example, you can destructure `window.QUnit` as follows: `const { QUnit } = window;` and then use `QUnit` directly in the code.

Please note, these are just recommendations and the necessity of these changes depends on the overall context and use case of the code.

Issue #1428, opened 2020-01-31
Tab order for "Module" and "Filter" items is reversed in toolbar


eddylu94:
## Tell us about your runtime:

* **QUnit version**: v2.9.2
* **What environment are you running QUnit in? (e.g., browser, Node)**: Edge, Chrome
* **How are you running QUnit? (e.g., script, testem, Grunt)**: Script

## What are you trying to do?

Code that reproduces the problem:

```js
.qunit-filter,
#qunit-modulefilter {
	float: right;
	position: relative;
	margin-left: 1em;
}
```

If you have any relevant configuration information, please include that here:

The toolbar includes two filters, **Module** and **Filter**:

![image](https://user-images.githubusercontent.com/6400501/73509505-2ac51000-4394-11ea-92c7-bc299b502a40.png)

One will notice that, while tabbing through the application, the **Filter** item will be focused upon first before the **Module** item. This is because the **Filter** item (class="qunit-filter") is listed before the **Module** filter (class="qunit-modulefilter") in the DOM structure:

```
<form class="qunit-modulefilter">...</form>
<form class="qunit-filter">...</form>
```

This reverse ordering is a consequence of using CSS 'float' style, where visual order is inverted from that of the DOM order.

## What did you expect to happen?

The **Module** toolbar item should be focused upon before the **Filter** toolbar item.

## What actually happened?

The tab order is reversed.

eddylu94:
Closing issue as it has been resolved in commit 95226b78702633354a13556404cde381eebab10d

PR #1429, opened 2020-01-31
Correct tab order for toolbar filters

eddylu94:
The objective of this pull request is to correct the tab order for the toolbar filters. Currently, the tab order for the **Module** and **Filter** items in the toolbar are in reverse order (right-to-left for English) as described by Issue #1428:

Issue #1428 - Tab order for "Module" and "Filter" items is reversed in toolbar

The changes applied in this pull request wrap the two toolbar filters into a single container SPAN element. This new container has a `float: right` style that acts as a substitute for the original `float` styles assigned to each of the two toolbar filters. When the individual `float` styles are removed from the toolbar filters, then the order of the corresponding DOM elements (also the tab order) for these filters will no longer be the reverse of the visual order.

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -502,13 +502,22 @@ export function escapeText( s ) {
 		return moduleFilter;
 	}
 
+	function toolbarFilters() {
+		var toolbarFilters = document.createElement( "span" );
+
+		toolbarFilters.id = "qunit-toolbar-filters";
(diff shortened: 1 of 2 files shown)


eddylu94:
@trentmwillis From local testing, the visual appearance of the UI remains unaltered when the changes are applied

trentmwillis:
Thanks!

PR #1431, opened 2020-02-19
Proc Queue/Test: add "started", "ended" to stats from tests/modules

davecombs:
Added 'started' and 'ended' fields in the stats information alongside 'runtime'. Both values were already available in internal code, just not returned in the test and module statistics information. They're very useful for diagnosing the precise timing of individual tests.

### DIFF
--- a/src/core/processing-queue.js
+++ b/src/core/processing-queue.js
@@ -162,7 +162,8 @@ function done() {
 
 	ProcessingQueue.finished = true;
 
-	const runtime = now() - config.started;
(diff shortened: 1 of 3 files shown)


davecombs:
Tests failing due to deep comparisons of stats object now containing started/ended values. Working out how to provide a meaningful value (since it is "live") without compromising the test.

davecombs:
I'm going to remove this PR, figure out what to do about the tests more generally, and resubmit.

Issue #1434, opened 2020-04-17
New test method proposal to run a subset of tests


ventuno:
[`only`](https://api.qunitjs.com/QUnit/only) currently allows a test to exclusively run, preventing all other tests from running.
It is currently impossible to use `only` or any other existing methods to run _only_ a subset of tests.
Changing the way `only` works would be a breaking change, so I recommend to add a new method `some`. All tests added with `some` will run and other tests will be skipped.

### Possible implementation
https://github.com/qunitjs/qunit/pull/1435

### Alternative approaches
It is technically possible to group tests in a module and then use [`module.only`](https://api.qunitjs.com/QUnit/module#exclusive-tests), but it is not quite as ergonomic as having the ability to selectively run a number of tests across modules and across files.

### References
Jest allows multiple tests to run with [`test.only`](https://jestjs.io/docs/en/api#testonlyname-fn-timeout).

trentmwillis:
Thanks for opening this and providing a possible implementation. I'd prefer to do something like Jest and just expand what `only` does (rather than introducing additional methods).

I'm not sure I'd actually qualify this as a breaking change. Technically, it changes behavior, but it is not backwards-incompatible if a test suite were to be using `.only` currently. Additionally, `.only` isn't intended to be committed to codebases (there's even [an eslint rule for it in the eslint-plugin-qunit](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-only.md)).

ventuno:
Thanks for the feedback @trentmwillis. I agree with you that in general this _should not_ be a breaking change, but in my experience things that _should not_ happen always end up happening and people might be counting on `only` to keep working the way it is documented.
On top of that, please consider that [`no-only` seems to have issues at the moment](https://github.com/platinumazure/eslint-plugin-qunit/issues/79).

As a way to mitigate possible issues, perhaps we could introduce a configuration parameter like `legacyOnlyBehavior` so that people can bring back the old behavior if needed?

Regardless, I trust your judgement as you definitely know more than I do about QUnit and its users.
I can easily update https://github.com/qunitjs/qunit/pull/1435 to modify the way `only` behaves instead of adding a new method. Just let me know what is the preferred approach.

PR #1435, opened 2020-04-17
[WIP] Core: Connected to #1434: add `some` method.

ventuno:
WIP implementation of #1434.

TODO:
- [ ] docs;

PR #1436, opened 2020-04-19
Core: Connected to #1434: run all `only` tests

ventuno:
WIP alternate implementation of #1434: modifying the way `only` behaves instead of adding a new method.

TODO:
- [x] docs;

### DIFF
--- a/docs/QUnit/module.md
+++ b/docs/QUnit/module.md
@@ -59,7 +59,7 @@ The module's callback is invoked with the test environment as its `this` context
 
 ## Exclusive tests
 
-When you are debugging your code, you will often need to _only_ run a subset of tests. You can append `only` to `QUnit.module` to specify which module you want to run.
+When you are debugging your code, you will often need to _only_ run a subset of tests. You can append `only` to `QUnit.module` to specify which module(s) you want to run.
 
 It works the same way `QUnit.only()` does for tests.
 
--- a/docs/QUnit/module.md
+++ b/docs/QUnit/module.md
@@ -61,8 +61,6 @@ The module's callback is invoked with the test environment as its `this` context
 
 When you are debugging your code, you will often need to _only_ run a subset of tests. You can append `only` to `QUnit.module` to specify which module you want to run.
 
-Note that if more than one module was defined using `QUnit.module.only()`, only the first one will run.
-
 It works the same way `QUnit.only()` does for tests.
 
 ## Inclusive tests
--- a/docs/QUnit/only.md
+++ b/docs/QUnit/only.md
@@ -8,7 +8,7 @@ categories:
 
 ## `QUnit.only( name, callback )`
 
-Adds a test to exclusively run, preventing all other tests from running.
+Add a set of tests to exclusively run, preventing other tests from running.
 
 | parameter | description |
 |-----------|-------------|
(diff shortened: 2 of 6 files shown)


ventuno:
Thanks @trentmwillis, let me know which approach is preferred and I'll complete this by adding docs :-).

smcclure15:
I'm in favor of this "running all only's" behavior. The new logic is clear ("run only these!"), where the current algorithm was a little fuzzy on which on won. The "run some" approach with a different API has a lot of behavioral overlap (two ways of doing the same thing). It helps close this gap, but introduces more fuzziness with it.

I don't think this change should be cause for a major semver bump, just a minor.

Krinkle:
> Thanks for implementing this approach as well! @qunitjs/core any of you have thoughts on this approach over #1435?

I'm +1 on improving `module.only` over introducing another method. To me this is a "Do what I mean" kind of thing where I see no harm from letting multiple run, if the flag is applied to module test suites.

As I understand it, its design was limited to one as a way to to avoid mistakes, not per se because it would be undesirable to support multiple.

I do note that if we change this behaviour, the docs should be updated as well. In particular, at <https://api.qunitjs.com/QUnit/only> it currently says:

> Note that if more than one `QUnit.only` call is present, only the first instance will run.

ventuno:
Thanks @Krinkle, as we seem to be leaning towards changing `only`'s approach I just pushed related doc changes. Please let me know if that looks good :-).

docs/QUnit/module.md
```diff
@@ -61,8 +61,6 @@ The module's callback is invoked with the test environment as its `this` context
 
 When you are debugging your code, you will often need to _only_ run a subset of tests. You can append `only` to `QUnit.module` to specify which module you want to run.
```
trentmwillis, line 62: To make this a bit clearer that it supports more than one module, maybe this line could be updated to say: `...specify which module(s) you want to run.`

docs/QUnit/only.md
```diff
@@ -8,7 +8,7 @@ categories:
 
 ## `QUnit.only( name, callback )`
 
-Adds a test to exclusively run, preventing all other tests from running.
+Add a set of tests to exclusively run, preventing other tests from running.
```
trentmwillis, line 11: This sounds like it allows you to add a set of tests with a single invocation. Maybe it could instead say:

```markdown
Adds a test to exclusively run, preventing any other tests not defined with `QUnit.only()` from running
```

docs/QUnit/only.md
```diff
@@ -23,9 +23,7 @@ Adds a test to exclusively run, preventing all other tests from running.
 
 ### Description
 
-Use this method to focus your test suite on a specific test. `QUnit.only` will cause any other tests in your suite to be ignored.
-
-Note that if more than one `QUnit.only` is present only the first instance will run.
+Use this method to focus your test suite on a specific set of tests. `QUnit.only` will cause any other tests in your suite to be ignored.
```
trentmwillis, line 26: After the change above, the first part of this could probably be updated to `...to focus your test suite on specific tests.`

docs/QUnit/module.md
ventuno: Done. Thanks :-).

docs/QUnit/only.md
ventuno: Updated.

ventuno:
> But I think if a `module.only` is followed by another `module.only` with the second one having nested modules, then the nested modules won't run (even though I think a user would expect them to).

Thanks @trentmwillis, I agree that modules nested inside a `module.only` should run as well. Thanks for catching this. I addressed (and tested) this, let me know if my solution looks good: I'm looking at the list of parent modules and if any is in the queue then I assume this module should run as well.

docs/QUnit/only.md
ventuno: Updated.

ventuno:
Thanks @trentmwillis. Looking forward to it :-).

trentmwillis:
Released in v2.10.0!

ventuno:
Thanks!

PR #1441, opened 2020-05-16
WIP HTML Reporter: Fuzzy search using fuse

ventuno:
Sample implementation of: https://github.com/qunitjs/qunit/issues/1438 ([`Fuse.js`](https://fusejs.io/)-based).
Below search results with different keywords (on `test/index.html`):

Search: "assert".
fuse assert

Search: "beforeEach".
fuse beforeEach

Search: "promiseaware".
fuse promiseaware

Search: "script".
fuse script


.babelrc
```diff
@@ -3,5 +3,5 @@
         // Don't transpile modules since Rollup does that
         [ "env", { modules: false } ]
     ],
-    "plugins": [ "external-helpers" ]
+    "plugins": [ "external-helpers", "babel-plugin-transform-object-rest-spread" ]
```
trentmwillis, line 6: Does the Fuse library use these ES features internally? Just want to know why this is required.

reporter/html.js
```diff
@@ -365,6 +366,15 @@ export function escapeText( s ) {
 			dropDownList = document.createElement( "ul" ),
 			dirty = false;
 
+		const options = {
+			includeScore: true,
```
trentmwillis, line 370: I don't see score being used. Is this needed?

reporter/html.js
```diff
@@ -365,6 +366,15 @@ export function escapeText( s ) {
...
+
+			// Search in `author` and in `tags` array
```
trentmwillis, line 372: This comment doesn't seem to apply to our data set. Is it needed?

.babelrc
ventuno: Yes [`Fuse.js` uses the spread operator](https://github.com/krisk/Fuse/blob/master/dist/fuse.esm.js#L215).

reporter/html.js
ventuno: Nope, removed :-).

reporter/html.js
ventuno: Removed.

ventuno:
@trentmwillis closing this in favor of https://github.com/qunitjs/qunit/pull/1442 as I was having some trouble after fetching the latest upstream changes.

Also, I'm a little partial to the [`fuzzysort` implementation](https://github.com/qunitjs/qunit/pull/1442) just because the set of results looks a lot more predictable than `Fuse.js` (see attached screenshots). See for example the search for "beforeEach" where results almost looks sorted "at random" or "script" which returns a number of unrelated results.

If you feel strongly about Fuse.js, then let me try and play a little more with its [configuration parameters](https://fusejs.io/concepts/scoring-theory.html) and see if I can come up with better results. I would strongly recommend that you fetch both locally and try them and see the difference for yourself.

PR #1445, opened 2020-06-06
Assert: Introduce strict boolean assertions.

ventuno:
Closes https://github.com/qunitjs/qunit/issues/1444.

TODO:
- [x] docs;
- [x] naming?

### DIFF
--- a/docs/assert/false.md
+++ b/docs/assert/false.md
@@ -1,0 +1,36 @@
+---
+layout: default
+title: false
+description: A boolean check. Passes if the first argument is false.
+categories:
+  - assert
+---
+
+## `false( state [, message ] )`
+
+A boolean check, inverse of `true()` equivalent to Chais's `assert.isFalse()`, and JUnit's `assertFalse()`. Passes if the first argument is false.
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| state              | Expression being tested              |
+| message (string)   | A short description of the assertion |
+
+### Description
+
+`false()` requires just one argument. If the argument evaluates to false, the assertion passes; otherwise, it fails. If a second message argument is provided, it will be displayed in place of the result.
+
+### Examples:
+
+```js
+QUnit.test( "false test", function( assert ) {
+  assert.false( false, "false succeeds" );
+
+  assert.false( "", "empty string fails" );
+  assert.false( NaN, "NaN fails" );
+  assert.false( null, "null fails" );
+  assert.false( undefined, "undefined fails" );
+  assert.false( true, "true fails" );
+  assert.false( 1, "1 fails" );
+  assert.false( "not-empty", "not-empty string fails" );
+});
+```
--- a/docs/assert/false.md
+++ b/docs/assert/false.md
@@ -8,7 +8,7 @@ categories:
 
 ## `false( state [, message ] )`
 
-A boolean check, inverse of `true()` equivalent to Chais's `assert.isFalse()`, and JUnit's `assertFalse()`. Passes if the first argument is false.
+A boolean check, inverse of `true()` equivalent to Chai's `assert.isFalse()`, and JUnit's `assertFalse()`. Passes if the first argument is false.
 
 | name               | description                          |
 |--------------------|--------------------------------------|
@@ -19,7 +19,9 @@ A boolean check, inverse of `true()` equivalent to Chais's `assert.isFalse()`, a
 
 `false()` requires just one argument. If the argument evaluates to false, the assertion passes; otherwise, it fails. If a second message argument is provided, it will be displayed in place of the result.
 
-### Examples:
+[`true()`](/assert/true) can be used to explicitly test for a true value.
+
+### Examples
 
 ```js
 QUnit.test( "false test", function( assert ) {
--- a/docs/assert/true.md
+++ b/docs/assert/true.md
@@ -1,0 +1,37 @@
+---
+layout: default
+title: true
+description: A boolean check, equivalent to Chai's assert.isTrue() and JUnit's assertTrue(). Passes if the first argument is true.
+categories:
+  - assert
+redirect_from:
+  - "/true/"
+---
+
+## `true( state [, message ] )`
+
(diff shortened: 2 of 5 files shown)


Krinkle:
Looks good to me. I'll do a more close review of the source code later on, but let's get the docs in meanwhile as well?

ventuno:
Thanks @Krinkle, just added docs for `assert.true` and `assert.false`.

Let me know if you'd like me to make some changes to the documentation for `ok`/`notOk` as well.
There are a number of references and usages of `assert.ok`/`assert.notOk` all over the documentation so if we decide to discourage its usage those should probably be changed to something else instead. I would suggest we take this up in a separate PR and perhaps discuss this in a new issue first?

Krinkle:
Yep, my thoughts exactly. Let's tackle that separately. I'll try to review this PR soon :) Thanks!

ventuno:
Thanks @Krinkle: created https://github.com/qunitjs/qunit/issues/1450.

docs/assert/false.md
```diff
@@ -0,0 +1,36 @@
...
+layout: default
+title: false
+description: A boolean check. Passes if the first argument is false.
+categories:
+  - assert
+---
+
+## `false( state [, message ] )`
+
+A boolean check, inverse of `true()` equivalent to Chais's `assert.isFalse()`, and JUnit's `assertFalse()`. Passes if the first argument is false.
```
Krinkle, line 11: `Chai's` :)

docs/assert/false.md
```diff
@@ -0,0 +1,36 @@
...
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| state              | Expression being tested              |
+| message (string)   | A short description of the assertion |
+
+### Description
+
+`false()` requires just one argument. If the argument evaluates to false, the assertion passes; otherwise, it fails. If a second message argument is provided, it will be displayed in place of the result.
```
Krinkle, line 20: Would be good I think to have an extra line here that links to `true()`, see <https://api.qunitjs.com/assert/equal> for example which links to notEqual and strictEqual,

docs/assert/false.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Examples:
```
Krinkle, line 22: We usually have our sections without colon.

src/assert.js
```diff
@@ -173,6 +173,38 @@ class Assert {
 		} );
 	}
 
+	true( result, message ) {
+		const isTrue = result === true;
+		if ( !message ) {
+			message = isTrue ?
+				"true" :
+				`failed, expected argument to be true, was: ${dump.parse( result )}`;
```
Krinkle, line 181: I haven't run this yet to see what the output looks like in the HTML and CLI/Tap reporter, but I suspect this is redundant given we don't do this for equal etc either. I believe `ok()` did it mainly because it's comparison target was technically unknown, so we did not want to put too much emphasis in the actual/expected part where it would say `true` given that `ok()` also allows e.g. random objects and functions as truthy. But this does not apply to assert.true/false, I'd expect them to behave the same as `strictEqual(, true/false)`.

ventuno: Makes sense. I removed the custom message. See a screenshot below and let me know if that works for you.

output

docs/assert/false.md
ventuno: My bad: this is coming from copy-pasting the [`notOk` docs](https://api.qunitjs.com/assert/notOk). We should make sure to fix that as part of https://github.com/qunitjs/qunit/issues/1450. I also added the (correct) header to `true.md` which was missing (as it's missing in [`ok` docs](https://api.qunitjs.com/assert/ok)).

docs/assert/false.md
ventuno: Added a reference from `false.md` to `true.md` and one from `true.md` to `false.md`.

docs/assert/false.md
ventuno: Good catch: fixed.

docs/assert/true.md
```diff
@@ -0,0 +1,41 @@
+---
+layout: default
+title: true
+description: A boolean check, equivalent to Chai's assert.isTrue() and JUnit's assertTrue(). Passes if the first argument is true.
+categories:
+  - assert
+redirect_from:
+  - "/true/"
```
Krinkle, line 8: Many files contain this for compat with URLs from an older iteration of the API doc site (which was based on WordPress and jquer-wp-content). Not needed for new pages. I'll omit this while landing.

docs/assert/true.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+## `true( state [, message ] )`
+
+A boolean check, equivalent to Chai's assert.true() and JUnit's assertTrue(). Passes if the first argument is `true`.
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| `state`            | Expression being tested              |
```
Krinkle, line 17: This is a left-over from `ok()`, for which in turn I actually don't know where it originates, but in general we call this the `actual` value. I'll use that here as well when landing the commit.

docs/assert/false.md
```diff
@@ -0,0 +1,38 @@
...
+---
+
+## `false( state [, message ] )`
+
+A boolean check, inverse of `true()` equivalent to Chai's `assert.isFalse()`, and JUnit's `assertFalse()`. Passes if the first argument is false.
+
+| name               | description                          |
+|--------------------|--------------------------------------|
+| state              | Expression being tested              |
+| message (string)   | A short description of the assertion |
```
Krinkle, line 16: missing backticks :)

ventuno:
Thanks @Krinkle any idea when these changes will be released?

Krinkle:
@ventuno I'm aiming for within the next few weeks, in July?

ventuno:
Sounds good. Thanks!

bmish:
I'd like to add types for these new assertions in @types/qunit: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49911

Krinkle:
@bmish Thanks. I've left a review over there, LGTM. Would you be interested in contributing these into the qunit project directly?

PR #1447, opened 2020-06-13
WIP: HTML Reporter: Fuzzy search using fuse@6.1.0-alpha.1

ventuno:
Follow up on: https://github.com/qunitjs/qunit/pull/1442#issuecomment-643550086. Testing `Fuse.js`@6.1.0-alpha.1.

#### `Fuse.js`@6.1.0-alpha.1.

Search: "assert".
fuse@6 1 0-alpha 1  - assert

Search: "beforeEach".
fuse@6 1 0-alpha 1  - beforeEach

Search: "promiseaware".
fuse@6 1 0-alpha 1  - promiseaware

Search: "promiseawarebeforeEach".
fuse@6 1 0-alpha 1  - promiseawarebeforeEach

Search: "script".
fuse@6 1 0-alpha 1  - script

#### `Fuse.js`@6.1.0-alpha.0.

Search: "assert".
fuse@6 1 0-alpha 0  - assert

Search: "beforeEach".
fuse@6 1 0-alpha 0  - beforeEach

Search: "promiseaware".
fuse@6 1 0-alpha 0  - promiseaware

Search: "promiseawarebeforeEach".
fuse@6 1 0-alpha 0  - promiseawarebeforeEach

Search: "script".
fuse@6 1 0-alpha 0  - script

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,5 +3,5 @@
         // Don't transpile modules since Rollup does that
         [ "env", { modules: false } ]
     ],
-    "plugins": [ "external-helpers" ]
+    "plugins": [ "external-helpers", "babel-plugin-transform-object-rest-spread"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "qunit",
-  "version": "2.10.0-pre",
+  "version": "2.10.1-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -423,6 +423,12 @@
       "integrity": "sha1-nufoM3KQ2pUoggGmpX9BcDF4MN4=",
       "dev": true
     },
+    "babel-plugin-syntax-object-rest-spread": {
+      "version": "6.13.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-syntax-object-rest-spread/-/babel-plugin-syntax-object-rest-spread-6.13.0.tgz",
+      "integrity": "sha1-/WU28rzhODb/o6VFjEkDpZe7O/U=",
+      "dev": true
+    },
     "babel-plugin-syntax-trailing-function-commas": {
       "version": "6.22.0",
       "resolved": "https://registry.npmjs.org/babel-plugin-syntax-trailing-function-commas/-/babel-plugin-syntax-trailing-function-commas-6.22.0.tgz",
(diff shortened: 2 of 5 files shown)


krisk:
Re: https://github.com/qunitjs/qunit/pull/1442#issuecomment-643686228

The reason is because of the [field-length norm](https://fusejs.io/concepts/scoring-theory.html#field-length-norm), where the shorter the field is, the higher the relevance of a match will be.

However, it sounds like what you may be looking for then is a basic relevance scoring. In this case, I'd recommend pulling `fuse.js@6.1.0-alpha.1`, and setting these options:

- `ignoreFieldNorm` => `true`
- `ignoreLocation` => `true`

By doing this, when I search for `"promiseawarebeforeEach"`, I get the following:

```js
[
  ({
    item: 'Module with Promise-aware beforeEach',
    refIndex: 3,
    score: 0.09090909090909091
  },
  {
    item: 'Module with Promise-aware afterEach',
    refIndex: 5,
    score: 0.2727272727272727
  },
  {
    item: 'Promise-aware return values without beforeEach/afterEach',
    refIndex: 4,
    score: 0.4090909090909091
  },
  { item: 'async beforeEach test', refIndex: 1, score: 0.5 },
  { item: 'assert.async in beforeEach', refIndex: 2, score: 0.5 },
  { item: 'beforeEach', refIndex: 0, score: 0.5454545454545454 })
]
```

Hopefully, this settles the use case? 😅

ventuno:
Thanks @krisk for further looking into this. I updated to `fuse.js@6.1.0-alpha.1` and although results look better I think they still look a little odd (see updated screenshots in the description. I kept both sets for comparison).

[Here's the list of modules we are filtering](https://pastebin.com/RGzFjqgq) I removed all other properties and only left `name`, which is the property we use for filtering. It's a little hard for me to provide the expected sorting, but my expectation would be that if I search "promiseawarebeforeEach" all results containing "Promise-aware" and "beforeEach" (together) should be closer to the top, so for example seeing "Module with Promise-aware afterEach" ranked higher than "Promise-aware return values without beforeEach/afterEach" is not really what I expect (`fuzzysort` is very close to what I expect [example 1](https://github.com/qunitjs/qunit/pull/1440#issue-418914860) and [example 2](https://github.com/qunitjs/qunit/pull/1442#issuecomment-636652146)).

As a I user, I know what my module name is and I'm just trying to get to it using the filter.
As far as I understand (just based on empirical observation) `Fuse.js` is optimized to surface matches when I _don't know_ what's in the content, while [`fuzzysort`](https://github.com/farzher/fuzzysort)'s goal seems more optimal to our goal as it strives to match SublimeText's file finder.
Does that make sense?

ventuno:
Closing based on the above.

PR #1449, opened 2020-06-21
Allow test runs with zero assertions to pass

jayjayjpg:
Fixes https://github.com/qunitjs/qunit/issues/1405

### Previous Behaviour

Before this change, test runs that included tests which had in total zero assertions, would fail with a global error. This made it, f.e. not possible to run test cases leveraging other means of assertions than the QUnit built-ins (`assert.ok`, `assert.equals`, etc.)
For a reproduction example, see also the description of the original issue: https://github.com/qunitjs/qunit/issues/1405

### New Behaviour

With this change, QUnit will decide if the current test run is "empty" and should therefore fail based on the number of tests included in the run rather than the number of included assertions. This should allow users to run test sets containing zero assertions, as long as respective test cases are marked as such using the `assert.expect(0)` declaration.

### How to Test

Run the new example test case containing no assertions to verify the new behaviourL

```bash
bin/qunit.js test/cli/fixtures/zero-assertions.js
```

| before | after |
|---|---|
| image | image |

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -299,6 +299,7 @@ Test.prototype = {
 		this.runtime = now() - this.started;
 
 		config.stats.all += this.assertions.length;
+		config.stats.testCount += 1;
 		module.stats.all += this.assertions.length;
 
 		for ( i = 0; i < this.assertions.length; i++ ) {
--- a/src/core.js
+++ b/src/core.js
@@ -96,7 +96,7 @@ extend( QUnit, {
 
 		// Initialize the configuration options
 		extend( config, {
-			stats: { all: 0, bad: 0 },
+			stats: { all: 0, bad: 0, testCount: 0 },
 			started: 0,
 			updateRate: 1000,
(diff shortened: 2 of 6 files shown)


src/test.js
```diff
@@ -125,7 +125,7 @@ Test.prototype = {
 		// ensure the callbacks are executed serially for each module
 		var callbackPromises = notStartedModules.reduce( ( promiseChain, startModule ) => {
 			return promiseChain.then( () => {
-				startModule.stats = { all: 0, bad: 0, started: now() };
```
Krinkle, line 128: I believe these are different state, they are per-module ones. I think this would be unused currently (not incremented, not read). Can you double check?

jayjayjpg: Awesome, I didn't catch that yet. I just checked locally and it seems to work as you suggested!

jayjayjpg:
:sparkles:

PR #1451, opened 2020-06-25
CSS: Scope QUnit UI button style to not affect #qunit-fixture (2.10.x backport)

Krinkle:
There are currently two places where we use buttons:
* The toolbar, with the filter "Go" button and the "Apply" and "Reset"
  buttons for the module dropdown.
* The test results, with the temporal "Abort" button.

Fixes https://github.com/qunitjs/qunit/issues/1437.

### DIFF
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -11,8 +11,12 @@
 
 /** Font Family and Sizes */
 
-
-[id^=qunit] button {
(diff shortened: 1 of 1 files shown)


Krinkle:
This is the 2.10.x backport of https://github.com/qunitjs/qunit/pull/1452.

PR #1459, opened 2020-07-09
HTML Reporter: Hide skipped tests when hidepassed is set

### LINKED ISSUE #1208: Hide skipped tests
I'd like to propose a `hideskipped` tests url parameter that QUnit understands.  We run our tests with saucelabs which has videos of the tests running.  Without being able to hide skipped tests, we are unable to see which test failed.  

cursor_and_sauce_labs___test_details

This would work just like `hidepassed`.  Both should be able to be active at the same time.  

Let me know if this sounds reasonable and I can probably submit a pull request.  Thanks!

raycohen:
Fixes #1208

Also adds the test file with the for hidepassed to the list of tests to be run.

I don't think there is test coverage on my changes to the `toolbarChanged` method.


rwjblue:
Nice, thanks for picking this up @raycohen!

PR #1464, opened 2020-07-21
Core: Drop support for Node 6 and Node 8, require Node 10+

Krinkle:
Per <https://github.com/nodejs/Release>, Node 6 and 8 have
reached EOL in 2019.

This allows us to drop the 'resolve' package as dependency,
which is no longer needed as of Node 8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
-dist
-node_modules
-build/report
-browserstack-run.pid
-temp/
-docs/_site/
+/dist
+/node_modules
+/build/report
+/browserstack-run.pid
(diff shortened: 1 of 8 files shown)


Krinkle:
| Before | After
|--|--
|  | 
| 9 dependencies | 7 dependencies
| <https://snyk.io/test/npm/qunit/2.10.1?tab=dependencies> | <https://snyk.io/test/npm/qunit/2.11.0?tab=dependencies>

See also https://github.com/qunitjs/qunit/issues/1342#issuecomment-450455620.

PR #1466, opened 2020-08-04
Test: Clean up usages of assert.ok/notOk where possible

ventuno:
Implements: https://github.com/qunitjs/qunit/issues/1450.

@Krinkle / @trentmwillis: I'm leaving a few usages of `assert.ok`/`assert.notOk` as their usage seems legitimate as they are actually testing something related to the assertions themselves.
Remaining legitimate usages of `assert.ok`:
- [`test/stack-errors.js`](https://github.com/qunitjs/qunit/blob/ae78fa04351306f213b73518ff9b186b41a1a4cf/test/stack-errors.js#L49);
- [`test/logs.js`](https://github.com/qunitjs/qunit/blob/master/test/logs.js#L156);
- [`test/main/assert.js`](https://github.com/qunitjs/qunit/blob/master/test/main/assert.js#L23)

Remaining legitimate usages of `assert.notOk`:
- [`test/main/assert.js`](https://github.com/qunitjs/qunit/blob/master/test/main/assert.js#L32)

TODO:
- [x] Docs.

### DIFF
--- a/docs/config/QUnit.extend.md
+++ b/docs/config/QUnit.extend.md
@@ -37,6 +37,6 @@ QUnit.test( "QUnit.extend", function( assert ) {
   assert.strictEqual( base.a, 1, "Unspecified values are not modified" );
   assert.strictEqual( base.b, 2.5, "Existing values are updated" );
   assert.strictEqual( base.c, 3, "New values are defined" );
-  assert.strictEqual( base.z, undefined, "Values specified as `undefined` are removed" );
+  assert.false( "z" in base, "Values specified as `undefined` are removed" );
 });
 ```
--- a/docs/config/QUnit.extend.md
+++ b/docs/config/QUnit.extend.md
@@ -34,9 +34,9 @@ QUnit.test( "QUnit.extend", function( assert ) {
     z: undefined
   } );
 
-  assert.equal( base.a, 1, "Unspecified values are not modified" );
-  assert.equal( base.b, 2.5, "Existing values are updated" );
-  assert.equal( base.c, 3, "New values are defined" );
-  assert.ok( !( "z" in base ), "Values specified as `undefined` are removed" );
+  assert.strictEqual( base.a, 1, "Unspecified values are not modified" );
+  assert.strictEqual( base.b, 2.5, "Existing values are updated" );
+  assert.strictEqual( base.c, 3, "New values are defined" );
+  assert.strictEqual( base.z, undefined, "Values specified as `undefined` are removed" );
 });
 ```
--- a/test/cli/fixtures/memory-leak/index.js
+++ b/test/cli/fixtures/memory-leak/index.js
@@ -56,7 +56,7 @@ QUnit.module( "later thing", function() {
 
 		let snapshot = await streamToString( v8.getHeapSnapshot() );
 		let matches = snapshot.match( reHeap );
-		assert.true( matches && matches.length, "found local Foo in heap" );
+		assert.strictEqual( matches.length, 2, "found local Foo in heap" );
 
 		snapshot = matches = null;
 		foo2.destroy();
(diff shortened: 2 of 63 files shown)


ventuno:
@Krinkle updated the docs as well. I wonder if I should update the docs for `assert.ok` and `assert.notOk` to mention that their usage is not recommended, or if we should wait until they are deprecated instead.

Krinkle:
> […] I'm also seeing [some test failures](https://travis-ci.com/github/qunitjs/qunit/jobs/368452443) which seem unrelated to my changes. Let me know if that's a known issue or if it's something I should be looking into.

I've not seen this before. I've triggered a rebuild of several times and the failure appears to be deterministic. I've also triggered a rebuild of the same job on the latest master branch commit, and it still consistently passes there.

test/cli/fixtures/memory-leak/index.js
```diff
@@ -56,7 +56,7 @@ QUnit.module( "later thing", function() {
 
 		let snapshot = await streamToString( v8.getHeapSnapshot() );
 		let matches = snapshot.match( reHeap );
-		assert.ok( matches && matches.length, "found local Foo in heap" );
+		assert.true( matches && matches.length, "found local Foo in heap" );
```
Krinkle, line 59:
```
  message: "found local Foo in heap"
  severity: failed

  actual: 2
  expected: true
```

The  `matches.length` array property returns a number, which is `2` in this case.

You can keep the same loose semantics by changing it to `!!matches && matches.length > 0`, or preferably, use `strictEqual` expecting `2`.

This test interacts with the internal GC interface of Node.js, and because this changes between versions we only assert the behaviour of Node 14 and the test is skipped entirely on other versions. This is why the other builds are passing.

ventuno: Fixed. Thanks for looking into this!

ventuno:
@Krinkle there's a [new failure](https://travis-ci.com/github/qunitjs/qunit/jobs/369215677) complaining about a "Missing API key", is it something I've changed?

Krinkle:
@ventuno You may ignore that one. It's only authorized outside pull requests. I must've not filtered that job well enough, I'll get that fixed.

docs/config/QUnit.extend.md
```diff
@@ -34,9 +34,9 @@ QUnit.test( "QUnit.extend", function( assert ) {
...
   } );
 
-  assert.equal( base.a, 1, "Unspecified values are not modified" );
-  assert.equal( base.b, 2.5, "Existing values are updated" );
-  assert.equal( base.c, 3, "New values are defined" );
-  assert.ok( !( "z" in base ), "Values specified as `undefined` are removed" );
+  assert.strictEqual( base.a, 1, "Unspecified values are not modified" );
+  assert.strictEqual( base.b, 2.5, "Existing values are updated" );
+  assert.strictEqual( base.c, 3, "New values are defined" );
+  assert.strictEqual( base.z, undefined, "Values specified as `undefined` are removed" );
```
Krinkle, line 40: This change is in my opinion not correct. 

The first assertions demonstrate that values can be updated or replaced.

This last assertion demonstrates that a property can be deleted. An existing property with the value `undefined` and a property not existing, are two different things, which is why this assertion was here.

```js
var x = {
  foo: undefined
};

'foo' in x; // true, yes - it does exist as local or inherited property.
Object.hasOwnProperty(x, 'foo'); // does true, yes, it exists as local property.
x.foo === undefined; // true, yes it explicitly yields the undefined value.

'foo' in x; // false, property does not exist at all.
Object.hasOwnProperty(x, 'foo'); // false, property does not exist locally.
x.bar === undefined; // true, it implicitly yields the undefined value.
```

I suggest using `assert.false( "z" in base, … )`.

PR #1467, opened 2020-08-11
Build: New release process with branch protection and npm 2FA support

Krinkle:
* Remove use of <https://github.com/qunitjs/qunit-release>.

* Separate commands that run arbitrary npm JS from commands that need
  npm or github credentials so that these can be run in a container
  with only the current directory mounted and nothing else.

* Let the release manager run the `npm publish` and `git push`
  commands directly.

* Keep a small amount of automation with a simple, dependency-free,
  local script in the build/ directory.

* Use port 4000 instead of 8000 since 8000 is commonly used for
  long-runing web servers unrelated to QUnit, and 4000 is more
  used by ad-hoc dev tooling, such as `jekyll serve`, or `php -S`.

* Add npm run command for `grunt authors`

  This reomves the need to install grunt-cli globally.

Ref https://github.com/qunitjs/qunit/issues/1350.


trentmwillis:
Sorry, I'm a bit late. This seems reasonable to me!

PR #1476, opened 2020-08-21
CLI: Further improve startup performance

Krinkle:
When using a glob or directory argument, we currently recursively traverse the entire working directory (except for `.git` and `node_modules`) and then filter out results later with Minimatch. This is inefficient to say the least, and is especially noticable on large repositories.

I looked into a bunch of glob libraries but couldn't find any that uphold our security and simplicity principles, in so far that most of them bring in far too many dependencies that I can't feasibly audit or keep up with and (unfortunately) most likely the library author hasn't audited or trusts deeply either to e.g. not get compromised or taken over in subtle ways.

Except for one, `tiny-glob`, this seems to fit the bill nicely and would be a great library to boost through the QUnit community. I instrumented it with some checks and confirmed that it isn't just very fast (which some globbers achieve by spawning processes rather than being efficient), but indeed it is also efficient in not traversing directories that it knows cannot match any part of the pattern. e.g. `src/` will be entered for `**/*.js` but will not be entered for `test/**/*.js`.

@SparshithNR @rwjblue I tried this out on a few small code basis and an artificially bloated directory tree and it seems to work fine. I'd be great if you could give this a try and let me know if you find any issues and/or could measure improvements. Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1616,7 +1616,8 @@
     "balanced-match": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
-      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
+      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
+      "dev": true
     },
     "basic-auth": {
       "version": "2.0.1",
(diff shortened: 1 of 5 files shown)


Krinkle:
4 of 7 dependencies for the QUnit CLI are `minimatch` and its three dependencies. A fairly popular alternative is micromatch's [picomatch](https://www.npmjs.com/package/picomatch) library, which is a drop-in replacement for what we need, and with a decent [npm score](https://npms.io/search?q=picomatch). I've also audited the code and its tests, and it seems pretty solid.

| <http://npm.broofa.com/?q=qunit@2.11.0>
|--
| NPMGraph: qunit 2.11.0

gabrielcsapo:
I just wanted to say thank you so much for this! the CLI is so much faster!!!

PR #1489, opened 2020-10-03
Clean up internal access to document and performance globals

Krinkle:
The pattern of checking `A` to decide whether `B` is safe to use is imho an anti-pattern and reduces confidence in the code. This was previously the case for `setTimeout`, `performance` and `document`.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,4 +1,4 @@
-import { window, setTimeout } from "./globals";
+import { window, document, setTimeout } from "./globals";
 
 import equiv from "./equiv";
 import dump from "./dump";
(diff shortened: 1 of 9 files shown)


src/core.js
```diff
@@ -31,7 +31,7 @@ var globalStartCalled = false;
 var runStarted = false;
 
 // Figure out if we're running the tests from a server or not
-QUnit.isLocal = !( defined.document && window.location.protocol !== "file:" );
+QUnit.isLocal = ( window && window.location && window.location.protocol === "file:" );
```
rwjblue, line 34: If we update our babel setup we could use 🤔 

`window?.location?.protocol === 'file:'`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/core/utilities.js
```diff
@@ -7,33 +7,35 @@ export const now = Date.now || function() {
...
-	performance.now.bind( performance ) :
-	now;
+const nativePerf = getNativePerf();
 
-function detectPerformanceApi() {
-	return window &&
+function getNativePerf() {
+	if ( window &&
 		typeof window.performance !== "undefined" &&
 		typeof window.performance.mark === "function" &&
```
rwjblue, line 15: `if (typeof window?.performance?.mark === 'function') {`

rwjblue (approved):
Nice bit of cleanup here @Krinkle!

PR #1492, opened 2020-11-04
Assert: Allow fat arrow functions in throws expected handler

### LINKED ISSUE #1389: assert.throws fails when expected is an arrow function: 'prototype' property of expected is not an object

* **QUnit version**: 2.9.2 & 2.9.3-pre
Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:66.0) Gecko/20100101 Firefox/66.0
* **How are you running QUnit? (e.g., script, testem, Grunt)**: script

## What are you trying to do?

[run in jsbin](https://jsbin.com/dafijuqepi/edit?html,js,output)
```js
function f() {
  throw new Error('err');
}
function errCheck(e) {
  return e.message == 'err';
}
QUnit.test('foo', (t) => {
  t.throws(f, errCheck); // this assert is passed
  t.throws(f, (e) => e.message == 'err'); // this assert fails
})
```

test passed

## What actually happened?
1st assert passed,
but 2nd assert fails:
```
'prototype' property of expected is not an object
```

smcclure15:
fixes #1389
added positive and negative unittests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -175,7 +175,8 @@ module.exports = function( grunt ) {
 				"test/module-only",
 				"test/module-skip",
 				"test/module-todo",
-				"test/es2017/async-functions"
+				"test/es2017/async-functions",
+				"test/es2017/throws"
 			]
 		},
 		"watch-repeatable": {
--- a/src/assert.js
+++ b/src/assert.js
@@ -322,7 +322,9 @@ class Assert {
 				// Log the string form of the regexp
 				expected = String( expected );
 
-			// Expected is a constructor, maybe an Error constructor
+			// Expected is a constructor, maybe an Error constructor.
+			// Note the extra check on its prototype - this is an implicit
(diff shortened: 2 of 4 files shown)


Krinkle:
It took me a while to figure out why this works. In particular, I was trying to find how absence of this condition leads to ` TypeError: 'prototype' property of x is not an object`.

But I see now that the issue is coming from the native code behind the `instanceof` operator. I... wasn't expecting that!

```js
arr = () => {};
function Foo() {}

x = new Foo();
x instanceof arr;
//> TypeError: 'prototype' property of arr is not an object
```

Krinkle (approved):
LGTM. I'll get review this once more tomorrow and then land it.

Could you add an inline comment about the `instanceof` caveat for future eyes?

smcclure15:
Agreed, this also took me a bit.
Kudos to @ohze for supplying the solution in their issue.
This was indeed a "Good first issue" - glad to help!

PR #1493, opened 2020-11-05
Build: include more tests to coverage phase

smcclure15:
The code coverage we were getting was only from running the CLI tests.
This change runs all existing tests.

Locally, overall line coverage went from 61% to 70%.
In particular, the "src" folder went from 65% to 89%.


Krinkle:
The coverage step is (supposed to?) already be running both the CLI and main browser tests, and combining their reports:

```
     "coverage": "npm run build:coverage && npm run coverage:_cli && npm run coverage:_main && nyc report", 
```

Is this not working? If I understand correctly, this patch will mean it now runs all tests twice.

smcclure15:
I started this endeavor wanting to top-off the coverage for the assert.js src, wanting it near 100%. But then studying the missed lines, there was no way that they weren't hit, and it turns out it was because the unittests weren't being counted in the coverage. My understanding is that we had CLI tests and main browser tests, but still needed that "grunt test" phase to hit.

smcclure15:
You're absolutely right - there is some double-running going on.
I started fiddling with test/index.html and measuring coverage, and realized that if I _remove_ some suites, certain lines then start to show as "hit".
There's definitely some weird interaction with at least one "done" not being processed that I found. Without it, everything passes, but we somehow lose coverage from other tests. I'm combing through the suite now and will will either update this PR, or start a fresh one since this solution could go in a very different direction.

Issue #1413, opened 2019-10-12
dependencies should use version ranges


XhmikosR:
You shouldn't lock the dependencies. devDependencies is another thing that affects only your usage.

But locking dependencies means that package managers cannot flatten the deps as good as they could. Also, if any security patch releases are out, you will need to release a new version for consumers to get them.

I know some people might have a different opinion, but after all, that's why lock files exist.

rwjblue:
> But locking dependencies means that package managers cannot flatten the deps as good as they could

Ya, this is definitely true, but I think that the larger issue (from my personal perspective) is that consumers can't get compatibility (e.g. for new Node versions) or security updates.

> I know some people might have a different opinion, but after all, that's why lock files exist.

As far as I know, the current system predates modern package managers with lock files. Now that (AFAICT) all major package managers have a lockfile by default (for quite some time now), I think it makes sense to reconsider...

ondrejsevcik:
Second that. Would you accept a PR for this issue?

Krinkle:
We choose to pin dependencies as part of the CLI project in 2017 (https://github.com/qunitjs/qunit/pull/1115).

I still tend to avoid version ranges for dependencies in most projects, however I think it might be fine for QUnit to lax its dependencies to a caret range like `^1.2.0`.

@trentmwillis What do you think?

trentmwillis:
I don't have a strong opinion. I wasn't around when the original decision was made to keep things pinned, but I imagine it might be similar to what @rwjblue mentioned with regards to it being before package managers had lockfiles.

PR #1494, opened 2020-11-05
Build: Use version ranges in package dependencies

smcclure15:
Use caret-ranges on all dependencies
Addresses #1413


package.json
```diff
@@ -41,45 +41,45 @@
     "node": ">=10"
   },
   "dependencies": {
-    "commander": "6.0.0",
-    "js-reporters": "1.2.3",
-    "node-watch": "0.6.4",
-    "tiny-glob": "0.2.6"
+    "commander": "^6.0.0",
```
Krinkle, line 44: For the production dependencies, we prefer to pin and audit individual packages directly. These should stay as before.

The rest of the patch is a-OK :)

Remember to re-run npm install after changing this so that the lock file is synced again.

Krinkle (approved):

PR #1495, opened 2020-11-06
Deprecate QUnit.extend

### LINKED ISSUE #1491: Deprecate QUnit.extend
I was suprised to see the `QUnit.extend` API still fully supported, or at least not marked deprecated.
Would the native `Object.assign` not suffice, and QUnit can back away from this redundant implementation/support?

* [ ] Mark `QUnit.extend` as deprecated in code.
* [ ] Update documentation page.

smcclure15:
Fixes #1491 

Backfilled a few unittests for sanity-checks and coverage
QUnit.extend issues a warning that it is to be deprecated in QUnit 3.0, and recommends `Object.assign`. (do we want an MDN hyperlink?)
The core `extend` utility is still warning-free, and used internally in many places. In QUnit 3.0, those can move to `Object.assign` as well.
Added deprecated status on docs, guessing that this would land in 2.12.

### DIFF
--- a/docs/config/QUnit.extend.md
+++ b/docs/config/QUnit.extend.md
@@ -1,9 +1,13 @@
 ---
 layout: default
-categories: [config]
+categories:
+- config
+- deprecated
 title: QUnit.extend()
+status: deprecated
 description: Copy the properties defined by the <code>mixin</code> object into the <code>target</code> object.
 version_added: "1.0"
+version_deprecated: "2.12"
 ---
 
 `QUnit.extend( target, mixin )`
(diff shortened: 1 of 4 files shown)


docs/config/QUnit.extend.md
```diff
@@ -1,9 +1,13 @@
...
 layout: default
-categories: [config]
+categories:
+- config
+- deprecated
 title: QUnit.extend()
+status: deprecated
 description: Copy the properties defined by the <code>mixin</code> object into the <code>target</code> object.
 version_added: "1.0"
+version_deprecated: "2.12"
```
Krinkle, line 10: I've only recently added support for these version pairs, and only once before a release. The way I did it that one time is setting it to [unreleased](https://github.com/qunitjs/qunit/search?q=unreleased) (since it'll go live on the main site as soon I land this PR), and then as part of the release prep commit, replace with the version.

I'm not worried about it being unknown which version it'll end up in, I'm sure it'll end up in 2.12. But I want to avoid broken links or it appearing as if it's already in effect when it isn't yet.

Krinkle: Actually, I'll do a release today so I'm landing this as-is for now.

Krinkle (approved):

Krinkle:
>  QUnit.extend issues a warning that it is to be deprecated in QUnit 3.0, and recommends Object.assign. (do we want an MDN hyperlink?)

Good idea. I'm doing a commit now with some release preparation clean up, and will add the link in the same go. Thanks!

PR #1496, opened 2020-11-06
Provide QUnit.test aliases for only/todo/skip

### LINKED ISSUE #1402: Add QUnit.test.only as alias for QUnit.only (and QUnit.test.todo, QUnit.test.skip)
Hello!

I'm wondering if it would be a good idea to chain module methods together to describe each test. Here's what I'm thinking.

Instead of...

```js
import { module, test, only } from 'qunit';

module('something', function() {
  test('first', function() {
    // the first test
  })

  only('the test im working on', function() {
    // the second test
  })
})
```

We could allow the following syntax...

```js
import { module, test } from 'qunit';

module('something', function() {
  test('first', function() {
    // the first test
  })

  test.only('the test im working on', function() {
    // the second test
  })
})
```

The main reasoning behind this change is as tests files grow in size I don't have to manage the imports. As I toggle `only` on and off I have to add and remove the import, which gets cumbersome. Chaining only off of test would avoid this.

This is a half baked thought, but I figured I'd start some discussion around this.

smcclure15:
Fixes #1402 
```
QUnit.test.only === QUnit.only
QUnit.test.todo === QUnit.todo
QUnit.test.skip === QUnit.skip
```

#1402 called out these QUnit.test variants as aliases _to_ the main QUnit functions, but to keep the code modular, I reversed that in the implementation to make the _main_ QUnit functions aliases to the test implementations. Those are really the "shortcuts" at that level, and Test holds onto the real implementations (as it always had). These methods are now encapsulated and exposed only through the Test function/object, no longer as exported functions.

Our existing tests still use the main QUnit methods, which I think is perfectly valid.

I doc'ed these new aliases in Test.md, pointing to the existing main QUnit methods. I did _not_ touch the main only/todo/skip docs to point to anything about these new aliases, but I could add a "Changelog" section in each to mention these? I didn't want to clutter.

### DIFF
--- a/docs/QUnit/test.md
+++ b/docs/QUnit/test.md
@@ -14,6 +14,9 @@ version_added: "1.0"
 ---
 
 `QUnit.test( name, callback )`
+`QUnit.test.only( name, callback )`<br>
+`QUnit.test.skip( name, callback )`<br>
+`QUnit.test.todo( name, callback )`
 
 Add a test to run.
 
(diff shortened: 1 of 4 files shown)


smcclure15:
Hmm, got a failure in "CLI Watch > properly watches files after initial run", only on Mac/Node10. Looks unrelated.
I've seen that before on a different change, maybe a flaky test...?

Krinkle:
> Hmm, got a failure in "CLI Watch > properly watches files after initial run", only on Mac/Node10. Looks unrelated.
> I've seen that before on a different change, maybe a flaky test...?

Yeah, it started happening a few weeks ago. Possibly due to a change in Node.js or in Travis' Mac systems. I've re-run those builds, and they pass now.

PR #1501, opened 2020-11-09
Build: Prepare 2.12.0 release

Krinkle:


Krinkle:
Tested in latest Firefox, Safari, and Chrome; and in IE9, IE10 and IE11.

PR #1502, opened 2020-11-09
Core: deprecate QUnit.load

### LINKED ISSUE #1084: Deprecate QUnit.load
This is not even documented, but it's used everywhere.

We need to make it an unnecessary call and then warn if it's called.

### Subtasks

* [x] Understand the practical use cases for this undocumented method.
* [x] Create a documentation page for `QUnit.load` and add examples of how to accomplish these use cases instead.
* [x] In QUnit 2.x, deprecate the method with a warning and link to these examples.

smcclure15:
Fixes #1084
Modularize and export load implementation
The load impl remains as-is
The QUnit.load entry point issues deprecation warning

There's more untangling we can do here among `start`/`config.autostart`/`load`, but this will allow us to drop the `QUnit.load` entry point in 3.0 while we are free to refactor internally.

In the deprecation message, a recommended workflow can get a little complicated; it's not just "use `QUnit.start` instead", because there are some gotchas (can't call it more than once, etc). Given that this was never documented, I don't know that it's a must-have to suggest a replacement. Feedback welcomed


Krinkle (changes requested):
Before this lands, I think we should make sure that all relevant use cases have a known path forward and documented on `QUnit.load` page for people to find. Let's continue that conversation on https://github.com/qunitjs/qunit/issues/1084.

smcclure15:
This is severely outdated and there's been a lot of churn to merge back in.
This change is very easily re-implemented if that's something that is needed in the future.
Closing this stale PR to keep a clean state.

PR #1503, opened 2020-11-09
Build: fix missing coverage from main browser tests

smcclure15:
A continuation of what #1493 was getting at - the right test _phases_ were being run, but many testpoints were _missed_.

I isolated that the "sandboxed-iframe.html" test was the one causing issues in the browser tests, though only in coverage mode. I admit I don't understand what that iframe and injection step is doing, but it was leaking into the other tests queued up, and resulting in a lot of testpoints being omitted.

With it listed first, we get 129 tests completed with 210 assertions, and overall code coverage at 61%.
Moving it to the end, we get **348** tests completed with **949** assertions, and overall code coverage at **88%**.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -110,7 +110,6 @@ module.exports = function( grunt ) {
 						require.resolve( "grunt-contrib-qunit/chrome/bridge" )
 					],
 					urls: [
-						"test/sandboxed-iframe.html",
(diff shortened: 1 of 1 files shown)


Krinkle (approved):

Krinkle:
Thanks, that must've been broken for a while. Nice find!

PR #1504, opened 2020-11-11
Tests: Add unittests to increase code coverage

smcclure15:
Adding unit tests to increase code coverage.
Was able to get a few areas to 100% statement coverage.

### DIFF
--- a/test/main/assert.js
+++ b/test/main/assert.js
@@ -189,13 +189,11 @@ QUnit.test( "throws", function( assert ) {
 	assert.throws( function() {
 		// eslint-disable-next-line qunit/no-throws-string
 		assert.throws(
-			function() {
-				throw "my error";
-			},
-			"expected",
-			"throws fail when expected is string and message is non-null"
+			undefined, // irrelevant - errors before even verifying this
+			"expected is a string",
+			"message is non-null"
 		);
-	} );
+	}, /throws\/raises does not accept a string value for the expected argument/ );
 
 	// This test is for IE 7 and prior which does not properly
 	// implement Error.prototype.toString
@@ -317,9 +315,6 @@ QUnit.test( "throws", function( assert ) {
 	);
 
 	var errorObjectLiteral = {
-		toString: function() {
-			return "[object";
-		},
 		name: { // bad name
 			toString: function() { }
 		},
@@ -327,11 +322,12 @@ QUnit.test( "throws", function( assert ) {
 			toString: function() { }
 		}
 	};
+
 	assert.throws(
 		function() {
 			throw errorObjectLiteral;
 		},
-		errorObjectLiteral
+		/^Error$/
 	);
 
 	errorObjectLiteral.message.toString = function() {
@@ -341,7 +337,7 @@ QUnit.test( "throws", function( assert ) {
 		function() {
 			throw errorObjectLiteral;
 		},
-		errorObjectLiteral
+		/^dummy message$/
 	);
 } );
 
--- a/test/main/assert.js
+++ b/test/main/assert.js
@@ -27,6 +27,8 @@ QUnit.test( "ok", function( assert ) {
 	assert.ok( Infinity );
 	assert.ok( {} );
 	assert.ok( [] );
+
+	assert.ok( true, "with message" );
 } );
 
 QUnit.test( "notOk", function( assert ) {
@@ -36,6 +38,8 @@ QUnit.test( "notOk", function( assert ) {
 	assert.notOk( null );
 	assert.notOk( undefined );
 	assert.notOk( NaN );
+
+	assert.notOk( false, "with message" );
 } );
 
 QUnit.test( "true", function( assert ) {
@@ -160,7 +164,7 @@ QUnit.test( "propEqual", function( assert ) {
 } );
 
 QUnit.test( "throws", function( assert ) {
-	assert.expect( 15 );
+	assert.expect( 18 );
 	function CustomError( message ) {
 		this.message = message;
 	}
@@ -181,6 +185,17 @@ QUnit.test( "throws", function( assert ) {
 		},
 		"simple string throw, no 'expected' value given"
 	);
+
+	assert.throws( function() {
+		// eslint-disable-next-line qunit/no-throws-string
+		assert.throws(
+			function() {
+				throw "my error";
+			},
+			"expected",
+			"throws fail when expected is string and message is non-null"
+		);
+	} );
 
 	// This test is for IE 7 and prior which does not properly
 	// implement Error.prototype.toString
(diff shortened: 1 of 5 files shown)


test/main/assert.js
```diff
@@ -182,6 +186,17 @@ QUnit.test( "throws", function( assert ) {
...
 	);
 
+	assert.throws( function() {
+		// eslint-disable-next-line qunit/no-throws-string
+		assert.throws(
+			function() {
+				throw "my error";
+			},
+			"expected",
+			"throws fail when expected is string and message is non-null"
```
Krinkle, line 196: Is this meant to test throwing of a string, or the unsupported matching against an expected string? I suspect the latter, in which case we can use a throw an Error object instead.

If I understand correctly, the message is currently lost -  t should be on the outer `throws()`, right? Could you also add an expectation match for that one, e.g. `/does not accept a string value/`.

test/main/assert.js
```diff
@@ -300,6 +315,34 @@ QUnit.test( "throws", function( assert ) {
...
+		name: { // bad name
+			toString: function() { }
+		},
+		message: { // bad message
+			toString: function() { }
+		}
+	};
+	assert.throws(
+		function() {
+			throw errorObjectLiteral;
```
Krinkle, line 332: Interseting.. It seems like it should behave the same as the plain `{ name: "SomeName", message: "some message" }` case earlier. Which branches does this cover?

Krinkle (changes requested):

test/main/assert.js
smcclure15: I agree, this nested throw should be clarified. It's intent is to verify an exception is thrown when "expected" is a string and "message' is non-null. This is the relevant branches in src/assert.js that this covers:

```
// 'expected' is optional unless doing string comparison
if ( objectType( expected ) === "string" ) {
	if ( message == null ) {
		message = expected;
		expected = null;
	} else {
		throw new Error(
			"throws/raises does not accept a string value for the expected argument.\n" +
			"Use a non-string object value (e.g. regExp) instead if it's necessary."
		);
	}
}
```

test/main/assert.js
smcclure15: This hits later branches within the "errorString" helper, where name and message can each have falsey values:

```
function errorString( error ) {
	const resultErrorString = error.toString();
 
	// If the error wasn't a subclass of Error but something like
	// an object literal with name and message properties...
	if ( resultErrorString.substring( 0, 7 ) === "[object" ) {
		const name = error.name ? error.name.toString() : "Error";
		const message = error.message ? error.message.toString() : "";
 
		if ( name && message ) {
			return `${name}: ${message}`;
		} else Eif ( name ) {
			return name;
		} else if ( message ) {
			return message;
		} else {
			return "Error";
		}
	} else {
		return resultErrorString;
	}
}
```

I wasn't convinced that `(new CustomError()).toString()` lead down the `[object` path, but I see now that it does. I can make this testpoint simpler and more direct...

Krinkle:
The "toString" method isn't usually something that people should define directly. You're right that it does work, but that is not the kind of value that this code is supposed to check.

The code is actually supposed to help detect when values _don't_ have a toString method.

Apologies if you know the rest already, but I'll show an example to illustrate it. It does seem kind of odd that calling toString helps us determine that a class does not have that method implemented. Here's an example:

```js
var x = { foo: "bar" };
var a = "" + x;
var b = x.toString();
console.log(a, b); // "[object Object]" "[object Object]"

class Foo {
}
var x = new Foo();
console.log("" + x); // "[object Object]
```

Both the plain object and the Foo class do not have "toString" method defined. In that case, the default we get is `[object Object]`. This default comes from the `Object.prototype.toString` method, which all objects inherit from. The `Error` class in JavaScript, on the other hand, does have its own "toString" method defined. It works like this:

```js
class MyError {
  constructor(message = "") {
    this.name = "MyError";
    this.message = message;
  }
  toString() {
    return this.name + ": " + this.message;
  }
}

var x = new MyError("Foo!");
console.log("" + x); // "MyError: Foo!"       not "[object Object]"
console.log(x.toString()); // "MyError: Foo!"       not "[object Object]"
```

I hope that made sense. Again, apologies if you already knew that. I don't mind :)

smcclure15:
Yup, I'm with you. That's a good demonstration.

I was trying to exercise the truthy branch of this:
```
const message = error.message ? error.message.toString() : "";
```
and yet still hit this later branch:
```
} else if ( message ) {
	return message;
} else {
	return "Error";
}
```

That's why I was really sabotaging the toString method. Maybe the falsy expression should be an `undefined` instead of `""`? That might even improve the
```
if ( name && message ) {
	return `${name}: ${message}`;
```
branch to be more readable.

That's an edge of behavior and I'm up for looking into. I'm just glad this deep dive in coverage exposed that kind of stuff, though this (code coverage topoff) and that (cleaner messaging?) could be considered independent changes.

Krinkle:
Ah... I see. I'm looking at the [coverage report](https://coveralls.io/builds/34974174/source?filename=src/assert.js) for [assert.js](https://github.com/qunitjs/qunit/blob/10728df0a22050928780613d79a2288df1e2542f/src/assert.js#L488) now:

cap

Indeed, it does not appear to be possible (under what I would consider normal conditions) for `name` to not have a truthy value here. Either `error.name` itself yields a non-empty string, or we use `Error`. Indeed the edge case that woudl need to unfold is where `error.name` exists as something truthy but then casts to an empty string. This seems very unlikely indeed.

Perhaps a simpler way would be to drop the odd explicit call here and use something like this instead:

```js
const name = String( error.name ) || "Error";
```

Here `String()` is called as a casting function (not a constructor), like what `"" + error.name` would do. This will natively call any toString method as needed, and then let us put in the fallback after both of those things have been taken care of.

Could you give that a go? It might also be worth a try to dig up the original change that created this to see if perhaps there was something more a useful idea here at some point that may have gotten lost.

smcclure15:
That sounds great - and I can dig into that next.
Would you agree that these are somewhat independent? I'd feel better about getting solid unit test coverage in, then as a separate PR, change source code behavior (and hopefully some newly failing tests!). Just trying to keep PR's and commits nice and crisp.

Krinkle:
I'm used to a patch-driven workflow (as opposed to branch-driven) and I agree these would indeed be separate commits. In the GitHub model, I can preserve as-is in the repo (not squash), and could be up for review together. Or I can review/merge one before the other, if the second isn't ready yet.

The main concern I have is that I'd rather not add test cases for code that isn't meant to be supported or has no expected behaviour. I do generally agree we should write tests before refactoring, and I think this code is fairly well-covered currently. If there are more supported cases we can add that are missing, that'd be fine to do first as well, then it would become 100% covered by the next commit removing the uncovered code.

Whichever you prefer 👍

smcclure15:
Committed in place here. I refactored the error formatting logic, with unittests that pass before and after (no incompatibilities, except for those extreme edge cases I was forced to test before).

smcclure15:
Added one last(?) commit with super lean logic.
I found some old context from the original implementation:
https://github.com/qunitjs/qunit/blob/6f8f6c3deaea5ae836f1941c96356cbc9e0b5a1c/src/core.js#L35

Those if/else's were the verbatim logic of this spec: http://es5.github.io/#x15.11.4.4
The simplified code retains the same behavior with less complexity. Glad this coverage work exposed this opportunity to clean up!

PR #1505, opened 2020-11-12
HTML Reporter: deprecate PhantomJS

smcclure15:
Use console.warn to message that PhantomJS is deprecated

I do have reservations about how visible this `console.warn` messaging may be (looks like different setups may suppress such outputs?), but it's a soft, opportunistic message before removal in 3.0.


Krinkle (approved):

Krinkle:
Ref https://github.com/qunitjs/qunitjs.com/issues/172.

PR #1510, opened 2020-11-16
CLI: Add native support for ESM .mjs files on Node 12+

Krinkle:
Ref https://github.com/qunitjs/qunit/issues/1465.


Krinkle:
Note to self: Update TAP output to use a regex and make the following line optional so that Node 12 and 14 both pass:

````
(node:5775) ExperimentalWarning: The ESM module loader is experimental.\n
```

https://travis-ci.com/github/qunitjs/qunit/jobs/440548904
https://github.com/nodejs/node/blob/v14.15.0/doc/changelogs/CHANGELOG_V14.md#ecmascript-modules---experimental-warning-removal

test/es2017/.eslintrc.json
```diff
@@ -1,5 +1,6 @@
 {
 		"parserOptions": {
-				"ecmaVersion": 2017
+				"ecmaVersion": 2018,
```
smcclure15, line 3: worth keeping separate 2017 and 2018 test suites? or just renaming/rev'ing the folder name?

rwjblue: Seems vaguely odd to allow ES2018 in a folder named es2017

rwjblue (commented):
We should also add support for `.cjs` files. Depending on your default setting in `package.json` you would either have to use `.cjs` or `.mjs` to use the other format (or to be explicit).

test/es2017/.eslintrc.json
Krinkle: Agreed, I'll rename the directory. In a follow-up I'll try to reorganize the tests more generally since it's quite a mess at the moment.

Issue #1511, opened 2020-11-17
Clean up internal test/ directory structure


Krinkle:
We currently have many ways we run unit tests:

1. In a browser (via Headless Chromium, grunt-contrib-qunit).
2. In a browser with a Web Worker (via Headless Chromium, grunt-contrib-qunit).
3. In Node.js, plainly (via a custom Grunt task).
4. In Node.js using the QUnit CLI.
5. In SpiderMonkey (via mozjs).

Outcome for this issue:

* To the extent possible, have the same these three test runners automatically share a common base list of test suites. Or, have build script that can keep them in sync. Or, have a small test that just tells us whether they are in sync and add that to `npm test`.
* In the repository, organise tests such that it is easy to spot which tests are common, which specific to browser, and which specific to CLI. Perhaps by giving those three use cases their own subdirectory, or by a mark in file name, or file comment.

For later:
* Walk through the tests that don't run in all three and determine whether there is a good reason for it, and document that reason. And if there isn't one, have it run in all three. From a quick check it seems that a big chunk of tests only ever run in the browser and in plain Node. The CLI test run seems limited to only CLI-specific stuff mostly. It'd be good to have the bulk of the tests run through there as well.

Krinkle:
From https://github.com/qunitjs/qunit/pull/1516:

> `test-on-node` is still the first task, […]. Once we've added the "main" tests to the CLI runs […], we may want to re-arrange it so that the CLI and browser test run first.

Krinkle:
https://github.com/qunitjs/qunit/blob/66081d66051d75f2ca10fed868eb512f5d3a6095/Gruntfile.js#L139-L151

As of https://github.com/qunitjs/qunit/pull/1629, the `module-skip.js` and `module-todo.js` test suites might pass on plain node.

PR #1516, opened 2020-11-20
 Build: Rewrite test-on-node task to use QUnit.on()

Krinkle:
_Depends on: https://github.com/qunitjs/qunit/pull/1515_

-------

Ref #1511.

### DIFF
--- a/build/tasks/test-on-node.js
+++ b/build/tasks/test-on-node.js
@@ -5,62 +5,23 @@ module.exports = function( grunt ) {
 	grunt.registerMultiTask( "test-on-node", async function() {
 		var done = this.async();
 
-		const stats = [];
+		const totals = { files: 0, failed: 0 };
 		for ( const file of this.data ) {
-			const stat = await new Promise( resolve => runQUnit( file, resolve ) );
-			stats.push( stat );
+			totals.failed += await new Promise( resolve => runQUnit( file, resolve ) );
+			totals.files++;
 		}
 
-		var totals = stats.reduce( function( totals, stats ) {
-			totals.passed += stats.passed;
-			totals.failed += stats.failed;
-			totals.skipped += stats.skipped;
-			totals.todo += stats.todo;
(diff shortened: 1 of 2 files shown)


build/tasks/test-on-node.js
```diff
@@ -5,63 +5,24 @@ module.exports = function( grunt ) {
...
-			todo: 0,
-			runtime: 0,
-			passedAssertions: 0,
-			failedAssertions: 0
-		} );
-
-		grunt.log.writeln( "-----" );
-		grunt.log.ok( constructMessage( totals ) );
+		if ( totals.failed ) {
+			grunt.log.error( `Ran ${totals.files} test files, ${totals.failed} failed` );
```
smcclure15, line 15: I always have to double-check: is "failed" the number of files failed, or the number of assertions failed across all tests/files?
If it's the latter, maybe rewording to "Ran X test files, Y assertions failed". Otherwise it reads fine, and it just my own hesitation.

smcclure15 (approved):
I like this - especially the cleaner/flatter logic inside src/test.js

build/tasks/test-on-node.js
Krinkle: Neither assertions nor files, it's the number of tests.

Updated to:
```js
Ran ${totals.files} files, ${totals.failed} failed tests
```

Krinkle:
- Fixed test counter.
- Merged https://github.com/qunitjs/qunit/pull/1515 and rebased on master.

Krinkle:
I realized this task runs before the headless chrome and CLI tests, which provide greater debugging details. Which meant that in the case of failures in the "main" tests, the minimal test-on-node task was the first failure and thus prevented the others from running.

It is still the first one, but at least as of this commit it includes a stack trace. Once we've added the "main" tests to the CLI runs (ref https://github.com/qunitjs/qunit/issues/1511), we may want to re-arrange it so that the CLI and browser test run first.

PR #1517, opened 2020-11-20
feat: better logging for assertions after a test

brandonocasey:
Was running into assertions happening after tests for one of our unit tests and locally made this change. It seems useful to know what test and what assertion are running after the test are supposed to have completed so I think that It would be nice to have this change in qunit.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -450,7 +450,10 @@ Test.prototype = {
 
 	pushResult: function( resultInfo ) {
 		if ( this !== config.current ) {
-			throw new Error( "Assertion occurred after test had finished." );
+			const message = resultInfo && resultInfo.message || '';
+			const testName = this && this.testName || '';
+
+			throw new Error( "Assertion '" + message + "' occurred after test '" + testName + "' had finished." );
 		}
 
 		// Destructure of resultInfo = { result, actual, expected, message, negative }
--- a/src/test.js
+++ b/src/test.js
@@ -450,10 +450,12 @@ Test.prototype = {
 
 	pushResult: function( resultInfo ) {
 		if ( this !== config.current ) {
-			const message = resultInfo && resultInfo.message || '';
-			const testName = this && this.testName || '';
+			var message = resultInfo && resultInfo.message || "";
+			var testName = this && this.testName || "";
+			var error = "Assertion " + message +
+				" occurred after test " + testName + " had finished.";
 
-			throw new Error( "Assertion '" + message + "' occurred after test '" + testName + "' had finished." );
+			throw new Error( error );
 		}
 
 		// Destructure of resultInfo = { result, actual, expected, message, negative }
--- a/src/test.js
+++ b/src/test.js
@@ -452,8 +452,9 @@ Test.prototype = {
 		if ( this !== config.current ) {
 			var message = resultInfo && resultInfo.message || "";
 			var testName = this && this.testName || "";
-			var error = "Assertion " + message +
-				" occurred after test " + testName + " had finished.";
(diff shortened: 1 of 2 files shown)


test/main/assert.js
```diff
@@ -667,18 +667,19 @@ QUnit.test( "rejects", function( assert ) {
...
 		previousTestAssert = assert;
 	} );
 
 	QUnit.test( "assertions after test finishes throws an error - part 2", function( assert ) {
 		assert.expect( 1 );
 		assert.throws( function() {
-			previousTestAssert.ok( true );
-		}, /Assertion occurred after test had finished/ );
+			previousTestAssert.ok( true, "message here" );
+		}, "Assertion message here occurred after test " + firstTestName + " had finished" );
```
smcclure15, line 683: This might read better with a few more delimiters, maybe something like
```
Assertion 'message here' occurred after test 'my test' had finished.
```
?

src/test.js
```diff
@@ -446,7 +446,12 @@ Test.prototype = {
 
 	pushResult: function( resultInfo ) {
 		if ( this !== config.current ) {
-			throw new Error( "Assertion occurred after test had finished." );
+			var message = resultInfo && resultInfo.message || "";
+			var testName = this && this.testName || "";
+			var error = "Assertion " + message +
+				" occurred after test " + testName + " had finished.";
```
smcclure15, line 452: What's a default message look like, say something like `assert.ok(actual)`? Is that just going to be "Assertion okay occurred..." ?
I try to optimize for the typical cases, but it's hard to say what those are. Plus, I don't have much to suggest on how we could improve it - maybe that's as simple as we need to make it.

smcclure15 (commented):
Seems like a helpful enhancement, being more specific.
I left a few comments on maybe where we could get the messaging more clear

test/main/assert.js
Krinkle: Agreed. Although test names themselves can also contain quotes at times. I think it would be okay to use multiple lines here. That's somewhat unusual for exception messages, but should be okay here.

For example:

```
Error: Assertion occurred after test finished.
> Test: assertions after test finish cause an error
> Assertion: message here

File: src/test.js (hunk 2 above)
Comment (Krinkle) [reply src/test.js:452]: Indeed. Every assertion should have a default message faik, and for `ok()` that is "okay".

File: test/main/assert.js (hunk 1 above)
Comment (brandonocasey) [reply test/main/assert.js:683]: Sounds great to me, will make the change and rebase as I saw that this file just changed.

File: test/main/test.js [hunk 3]
```diff
@@ -234,19 +234,23 @@ QUnit.module( "test", function() {
...
 			previousTestAssert = assert;
 		} );
 
 		QUnit.test( "assertions after test finishes throws an error - part 2", function( assert ) {
 			assert.expect( 1 );
 			assert.throws( function() {
-				previousTestAssert.ok( true );
-			}, /Assertion occurred after test had finished/ );
+				previousTestAssert.ok( true, "message here" );
+			}, "Assertion occurred after test had finished.\n" +
```
Krinkle, line 250: This string is used as message, not as expected value matcher. I apologize, this is ambiguous indeed due to both second and third parameter being optional. Support for literal string matching against `error.message` was removed for that reason. Either use a regex for a partial match, or use an Error object for a literal match.

E.g.

```js
assert.throws( function () {
  /* … */
}, new Error( "foo" ), "optional message" );
```

<https://api.qunitjs.com/assert/throws/>

brandonocasey: Ahh entirely my bad, I think I have that fixed now.

Krinkle (approved):

PR #1526, opened 2020-12-08
Core: Simplify start() and begin() logic to improve readability

smcclure15:
Flattening logic and pushing local vars down in scope
Inverted some if/else's to be simpler and avoid as much nesting as possible.

Incremental cleanup towards deprecating QUnit.load (#1502) that I wanted to isolate and keep driving forward

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -78,20 +78,23 @@ extend( QUnit, {
 				"QUnit.config.autostart was true" );
 		}
 
-		if ( config.pageLoaded ) {
-			scheduleBegin();
+		if ( !config.pageLoaded ) {
+
+			// The page isn't completely loaded yet, so we set autostart and then
+			// load if we're in Node or wait for the browser's load event.
+			config.autostart = true;
+
+			// Starts from Node even if .load was not previously called. We still return
+			// early otherwise we'll wind up "beginning" twice.
+			if ( !document ) {
+				QUnit.load();
+			}
+
 			return;
 		}
 
-		// The page isn't completely loaded yet, so we set autostart and then
-		// load if we're in Node or wait for the browser's load event.
-		config.autostart = true;
+		scheduleBegin();
 
-		// Starts from Node even if .load was not previously called. We still return
-		// early otherwise we'll wind up "beginning" twice.
-		if ( !document ) {
-			QUnit.load();
-		}
 	},
 
 	extend: function( ...args ) {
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -38,56 +38,61 @@ QUnit.isLocal = ( window && window.location && window.location.protocol === "fil
...
+		}
+		if ( config.autostart ) {
+			throw new Error( "Called start() outside of a test context when " +
+				"QUnit.config.autostart was true" );
+		}
 
-	is: is,
+		if ( config.pageLoaded ) {
+			scheduleBegin();
+			return;
```
Krinkle, line 83: I'm a fan of early returns for handling errors and unusual scenarios, but I think this might have taken it a bit too far as it is it now losing structure of the logic as well. Browser and Node js handling are both the "main" purpose of this function and I would expect those to be part of the same non-early part of the function body.

When extending the function it should be easy to add something to the end that both branches share, which would be difficult in this case as its natural end is now copied in two places. I don't really mind how much "work" it would be to add something to this function, that's easy to refactor if needed. The concern is that when reading this code now, it doesnt' resemble the logical decision tree that the function is responsible for, I think?

Anyway, my suggestion would be to omit this last one and use an else branch. Possibly reversing the order so that the non-ready case remains handled "first" like before.

src/core.js
```diff
@@ -155,37 +151,37 @@ function unblockAndAdvanceQueue() {
...
-		emit( "runStart", globalSuite.start( true ) );
-		runLoggingCallbacks( "begin", {
-			totalTests: Test.count,
-			modules: modulesLog
-		} ).then( unblockAndAdvanceQueue );
-	} else {
-		unblockAndAdvanceQueue();
+	// Avoid unnecessary information by not logging modules' test environments
+	const l = config.modules.length;
+	var modulesLog = [];
```
Krinkle, line 171: Feel free to convert individual functions from var to const/let in their entirety.

Krinkle (changes requested):
LGTM, but for two minor points

smcclure15:
Those changes sound reasonable to me - another commit coming

Krinkle:
Nice minimal diff now as well (if set to ignore whitespace). LGTM at glance, will try to land this weekend. Thanks!

Krinkle (approved):

PR #1527, opened 2020-12-08
Core: fix module.only logic for "flat" structures

### LINKED ISSUE #1272: `QUnit.module.only` behaves strange
`QUnit.module.only` behaves strange when using the module syntax that assigns tests to modules using the order of invocation, e.g.:

```javascript
QUnit.module.only('module 1');

QUnit.test('test in module 1', assert => {
  assert.ok(true, 'a successful assertion in module 1');
});

QUnit.module('module 2');

QUnit.test('test in module 2', assert => {
  assert.ok(true, 'a successful assertion in module 2');
});
```

Without `only` `test in module 1` belongs to `module 1` and `test in module 2` belongs to `module 2`. Using `only` in `module 1` the second test also seems to belong to `module 1` and is run (applying the wrong hooks if there would be any).

You can check this in this plunk: https://embed.plnkr.co/mh4VgJ0O6Dhp4H5APd3S/

smcclure15:
Fixes #1272

I'm still a little worried about more edge cases, but I wanted to get this out there and am very open to feedback on the approach.

For every module _after_ a `module.only` call, we need to say that the current module is then closed for appending, so that later tests can stay out of them. They are still a no-op for performance, but they need to communicate this extra piece.


test/module-only-flat.js
```diff
@@ -0,0 +1,82 @@
...
+
+	tests.push( {
+		name: details.name,
+		module: details.module,
+		skipped: details.skipped,
+		todo: details.todo
+	} );
+} );
+
+QUnit.done( function() {
```
Krinkle, line 19: I understand this follows the existing patterns for other tests here, but I've been thinking lately that perhaps we should phase this out since it seems to rely re-opening test suites which is something we actively don't support and afaik already doesn't work in many cases when a JS reporter is used. Incl from the command-line.

Could you see if this would work as a CLI test instead? Then we can simply assert the TAP output instead, which would also make it a more complete assertion for anything unexpected.

smcclure15: Great flag; I was wrestling with this verification for a bit, and was still uneasy that it was doing the right thing. I’ll get this into a CLI test, and with that I should be able to cover this area pretty exhaustively.

test/only.js
```diff
@@ -1,25 +0,0 @@
-QUnit.module( "QUnit.only", function( hooks ) {
```
smcclure15, line 1: This deleted file is causing a build failure (I see the 404's print in my local dev, but it didn't indicate a failure like Travis did).
It was being used in the "sandboxed-iframe" test, I _think_ to try to isolate it as much as possible. That should be resolved now with the CLI test suite, and I think some or all of the "bridge" injections can go away now. I'm still grasping at the intent of that sandbox test, so I'm curious of any historical knowledge others can share.

Krinkle: Based on the commit history for these files, I believe its purpose is not so much testing of the `QUnit.only` functionality, but rather it is a higher-level integration test for running QUnit from inside an iframe. Specifically, a sandbox iframe that has limited capabilitities. Ref https://github.com/qunitjs/qunit/pull/1091.

As such, some form of this should remain. See https://github.com/qunitjs/qunit/pull/1091#issuecomment-274123666 for more information. The test needs to cause use of code that tries to use `QUnit.config.storage` (aka localSessionStorage, aka window.sessionStorage).

Krinkle: Based on the commit history for these files, I believe its purpose is not so much testing of the `QUnit.only` functionality, but rather it is a higher-level integration test for running QUnit from inside an iframe. Specifically, a sandbox iframe that has limited capabilitities. Ref https://github.com/qunitjs/qunit/pull/1091.

As such, some form of this should remain. See https://github.com/qunitjs/qunit/pull/1091#issuecomment-274123666 for more information. The test needs to trigger code that tries to use `QUnit.config.storage` (aka localSessionStorage, aka window.sessionStorage).

Krinkle (changes requested):

test/only.js
smcclure15: I've restored this sandboxed-iframe test.
I brought back the only.js test logic, albeit renamed to `sandboxed-iframe.js` to not be confused with the `only` tests.
(also, I now see some whitespace changes in the HTMLs that I can correct right away)

It sounds like this might be able to be improved, but this is admittedly beyond my scope, so I'm restoring to status quo to keep issues independent.

src/module.js
```diff
@@ -100,27 +98,31 @@ function processModule( name, options, executeNow, modifiers = {} ) {
...
+		config.currentModule.closed = true;
 		return;
 	}
 
 	processModule( name, options, executeNow );
 }
 
-module.only = function() {
+module.only = function( ...args ) {
 	if ( !focused ) {
```
Krinkle, line 113: Based on how `test.only` is written, I would expect `focused = true;` to be set in this block instead of after calling `processModule()`.

If changing this would not affect the issue that this patch addresses, then feel free to leave it as-is and we can look at it later (to keep this patch clean).

src/module.js
```diff
@@ -100,27 +98,31 @@ function processModule( name, options, executeNow, modifiers = {} ) {
 	}
 }
 
+let focused = false;
+
 export default function module( name, options, executeNow ) {
 	if ( focused && !isParentModuleInQueue() ) {
+		config.currentModule.closed = true;
```
Krinkle, line 105: It took me a while to realize that this is actually mutating the "previous" module. The one we are processing now, never makes it into the registry. This feels a bit fragile as it means `test()` calls are thus interacting with a closed module and I can see that leading to some confusing failures in the future.

On the other hand, given the way that `only` is meant to act like a `filter`, the module isn't meant to be visible in any way. Perhaps, though, it would be best to still let it enter the registry, but as an empty module that (if all goes well) will never start or stop, because it is empty. Afaik that is how empty modules are treated.

If empty modules are exposed to reporters or callbacks in some way, then we'd need some additional logic to make sure e.g. empty modules with `currentModule.ignored` get bypassed in our events/callback layer.

Krinkle: Let me know if you like that direction. I haven't thought it through fully, so it's not a strong suggestion, just an idea.

For now my only suggestion would be to add an inline comment indicating what we have done, which is to keep the previous module on the stack and that "closed" means it was implicitly closed by a subsequent `module()` call.

Krinkle:
I notice that when I modify the Plnkr test case in #1272 to use a scoped module, the issue does not happen. I guess that's because we just skip execution of the closure, and not actually due to any explicit logic making that happen.

For `skip` and `todo`, we currently inherit the state from inside the `Test` constructor, which is kind of interesting. Although that approach probably wouldn't work well here since `only` is really an extension of `filter` and not of `skip` (that is, when using `only` the other tests shouldn't be registered at all, not even skipped).

I like the approach this takes. Just two minor nit picks:

* The new property should be explicitly declared and set to false by default.
* The purpose of this property, and the new behaviour of keeping the previous module on the stack, is worth documenting.

src/module.js
smcclure15: This is agnostic to _flat_ modules, but here's a tangible test to study with _nested_ modules:
```
QUnit.module.only( "Module A", () => {
    QUnit.module( "Module B", () => {
        QUnit.test( "Test A", function( assert ) {
            assert.true( true );
        } );
    } );
    QUnit.module.skip( "Module C", () => {
        QUnit.test( "Test B", function( assert ) {
            assert.true( true );
        } );
    } );
} );
```

With setting `focused = true` where it is today, `Test A` is run and `Test B` is skipped:
```
ok 1 Module A > Module B > Test A
ok 2 # SKIP Module A > Module C > Test B
1..2
# pass 1
# skip 1
# todo 0
# fail 0
```
Pushing the `focused = true` to before `processModule()` gives just `Test A`, and `Test B` is never even registered:
```
ok 1 Module A > Module B > Test A
1..1
# pass 1
# skip 0
# todo 0
# fail 0
```

The behavior is maybe debatable, but for the scope of this PR. I'd prefer we keep it as-is for compatibility, and maybe consider separately in the future.

src/module.js
smcclure15: I'm certainly intrigued by that suggestion, and the notion of `currentModule.ignored` has a much more natural feel and intuition.

I think it's inside `module.only` where we set `config.modules.length = 0;` to effectively wipe out the others that were registered before. I'm trying to make that work with something like
```
config.modules.forEach( module => {
   module.ignored = true;
} );
```
but I'm definitely wrestling with the unittests - I don't have everything quite right.

I'll keep this open and continue noodling. I agree we should document the exposed property, and `ignored` is just so much easier to digest than `closed` (or alternative).

smcclure15:
Updated PR with `ignored` property instead of the misleading `closed`.
Some of the logic definitely simplified, and it feels more intuitive.

If that direction looks good, I'll move on to documenting that new property!

Krinkle (approved):
LGTM. Will try to land soon. Thanks!

Krinkle:
Also confirmed the fix at <https://plnkr.co/edit/bBpWs5X2vKp2YPTu> (forked from [original](https://embed.plnkr.co/mh4VgJ0O6Dhp4H5APd3S/) in #1272).

cap

Issue #1427, opened 2020-01-30
HTML Report should meet WCAG compliance


smcclure15:
We'd like to use a tool like [Axe](https://github.com/dequelabs/axe-core) to help lockdown [WCAG](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG) rules for accessibility, but the QUnit page itself has some violations that we are forced to filter out. Ideally, yes, tests would only attempt to lockdown their particular widget/section, likely placed in the qunit-fixture, but this shouldn't be a limitation.

Aside from our specific intent, meeting these guidelines can independently benefit testing workflows with increased industry-standard accessibility.

Code that reproduces the problem:

```html
<html>
<head>
  <link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.9.3.css">
</head>
<body>
  <div id="qunit"></div>
  <div id="qunit-fixture">
  <script src="https://code.jquery.com/qunit/qunit-2.9.3.js"></script>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/axe-core/3.4.1/axe.min.js"></script>
  <script type="text/javascript">
    QUnit.test('Hello World', async assert => {
      await axe.run(document.body, (err, results) => {
        assert.equal(results.violations.length, 0);
        results.violations.forEach(v => {
          assert.notOk(v);
        })
      })
    })
  </script>
</body>
</html>
```

## What did you expect to happen?

No violations to be found

## What actually happened?

This currently finds 3 violations, failing the following rules:
* color-contrast (impact: serious)
* empty-heading (impact: minor)
* link-name (impact: serious)

### Checklist

See https://github.com/qunitjs/qunit/issues/1427#issuecomment-683370797

trentmwillis:
Thanks for the report. I definitely agree that this should be fixed.

Krinkle:
Merging from https://github.com/qunitjs/qunit/issues/1469:

> When scanning the default example at https://qunitjs.com/resources/example-index.html with the AXE tool in Firefox, there are a couple of issues detected currently. (Thanks to @zachleat for including these in [Speedlify](https://www.speedlify.dev/test-runners/). We looked at similar tools in the past, but it's been years, and a ton of low-hanging fruit has crept in.)
> 
>     * QUnit 2.10.1
> 
>     * axe v4.5.3 (axe-core 3.5.5)
> 
> 
> Most of these seem easy to fix:
> 
[...]

PR #1528, opened 2020-12-09
Reporter: fix several WCAG violations

smcclure15:
Resolves some of the more straightforward violations (1, 2, and 3) captured in #1427

### DIFF
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -599,7 +599,7 @@ export function escapeText( s ) {
 	}
 
 	function appendInterface() {
-		const qunit = id( "qunit" );
+		var qunit = id( "qunit" );
 
 		// For compat with QUnit 1.2, and to support fully custom theme HTML,
 		// we will use any existing elements if no id="qunit" element exists.
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -599,7 +599,7 @@ export function escapeText( s ) {
 	}
 
 	function appendInterface() {
-		var qunit = id( "qunit" );
+		const qunit = id( "qunit" );
 
 		// For compat with QUnit 1.2, and to support fully custom theme HTML,
 		// we will use any existing elements if no id="qunit" element exists.
@@ -611,12 +611,14 @@ export function escapeText( s ) {
 		// having the HTML reporter actively render anything.
 		if ( qunit ) {
 
+			qunit.setAttribute( "role", "main" );
+
 			// Since QUnit 1.3, these are created automatically if the page
 			// contains id="qunit".
 			qunit.innerHTML =
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
-			"<h2 id='qunit-banner'></h2>" +
-			"<div id='qunit-testrunner-toolbar'></div>" +
+			"<div id='qunit-banner'></div>" +
+			"<div id='qunit-testrunner-toolbar' role='navigation'></div>" +
 			appendFilteredTest() +
 			"<h2 id='qunit-userAgent'></h2>" +
 			"<ol id='qunit-tests'></ol>";
--- a/reporter/html.js
+++ b/reporter/html.js
@@ -617,7 +617,7 @@ export function escapeText( s ) {
 			// contains id="qunit".
 			qunit.innerHTML =
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
-			"<div id='qunit-banner'></div>" +
+			"<h2 id='qunit-banner'></h2>" +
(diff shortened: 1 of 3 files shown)


reporter/html.js
```diff
@@ -599,7 +599,7 @@ export function escapeText( s ) {
 	}
 
 	function appendInterface() {
-		var qunit = id( "qunit" );
+		const qunit = id( "qunit" );
```
leobalter, line 602: nit: I believe this is not addressing any accessibility issue.

reporter/html.js
```diff
@@ -611,12 +611,14 @@ export function escapeText( s ) {
...
 
+			qunit.setAttribute( "role", "main" );
+
 			// Since QUnit 1.3, these are created automatically if the page
 			// contains id="qunit".
 			qunit.innerHTML =
 			"<h1 id='qunit-header'>" + escapeText( document.title ) + "</h1>" +
-			"<h2 id='qunit-banner'></h2>" +
-			"<div id='qunit-testrunner-toolbar'></div>" +
+			"<div id='qunit-banner'></div>" +
```
Krinkle, line 620: I think this should wait until the QUnit 3.0 cycle. The other two changes are fine though, and should not affect compatibility in any way.

Krinkle (changes requested):

reporter/html.js
smcclure15: True - you are correct ;-)
I try to cautiously fix as I go, just within the function block, if it doesn't sprawl on me.
If it would be more palatable, I could work through an independent change to convert as many `var` declarations as possible to `let`/`const`? That would keep changes more atomic with intent. What do you think?

reporter/html.js
smcclure15: That sounds fair, thanks for the review!

reporter/html.js
Krinkle: I did indeed say on a previous CR it was fine to change them all at once for a given area of code when already changing other variables inside of it.

I do agree though that it stands out here a bit unrelated as nothing else was changed regarding variables.

Doing them all at once, I wouldn't mind either. Maybe to keep the PR lighter to review, do it in two or three medium-size passes, e.g. `/src/cli`, `/src` (rest), `/test` and then misc files like `/runner` etc. And to keep down the need to rebase, I'd recommend doing one at a time, e.g. start the next after the previous one lands (unless you don't mind rebasing once or twice).

Thanks!

Krinkle (approved):

Krinkle (approved):

runner/fixture.js
```diff
@@ -20,6 +20,7 @@ import { window, document } from "../src/globals";
 
 		var fixture = document.getElementById( "qunit-fixture" );
 		if ( fixture ) {
+			fixture.setAttribute( "role", "region" );
```
Krinkle, line 23: What is this one for? I would expect the fixture to be treated as a hidden thing and not a jump target. At any given moment for a developer consuming the test report or watching the progress, this wouldn't be meaningful or change too rapidly, I think?

I'm open to adding it, but want to understand why so that the intention can be preserved and not accidentally removed in the future.

I do worry that it might also be a compat concern, I'll have to do some digging to confirm that. The fixture should be as plain as possible so that test code has full control over what it behaves like. I can see how a frontend library would behave differently if placed inside a certain role area, or that e.g. a frontend library's test suite asserts one of its creations in the fixture to have an ancestor of role=…something which could now false-positive on the fixture itself.

smcclure15: That's a really good point of incompatibility; I could see this holding off until 3.0, if it is in fact something we need.

This was due to #3 on the checklist: "All page content must be contained by landmarks".
I tagged the "main" qunit div as role=main, but the fixture is a sibling of that, not a child, so it needed its own role as well.

I dabbled with the idea of leaving some of this to the test writer - they maintain their HTML runners, and could just as well provide the right roles. The fix here could arguably just a doc update at best?

Krinkle: I would not consider the fixture as part of the presented "page content".

Krinkle (changes requested):
Fixture attribute is likely breaking. Let's continue the conversation on that and land the rest.

smcclure15:
This change has reduced to simply adding a `main` role to the "qunit" div and a `navigation` role to the toolbar to avoid potential incompatibilities.

runner/fixture.js
smcclure15: Reverted!

reporter/html.js
smcclure15: Thanks for the suggestions - I'll prep some scoped changes for a few waves of churn.

Krinkle (approved):

PR #1541, opened 2021-01-12
Build: Use qunit/ during development instead of dist/

Krinkle:
Use the same directory structure for the build artefact during releases and during development.

Since the release file path is used by published npm and bower packages, keep that as-is, and rename the internal development target from `dist/` to `qunit/`.

After this, contributors and maintainers will want to delete any `dist/` directory they might have, to avoid accidental linting or
publishing.

Fixes https://github.com/qunitjs/qunit/issues/1133.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,8 +36,8 @@
         "__codeorigin/**",
         "coverage/**",
         "docs/_site/**",
-        "dist/**",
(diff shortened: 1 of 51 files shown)


Krinkle:
Could use a sanity check on this. Want to make sure I haven't missed anything for end-users or contributors.

PR #1550, opened 2021-01-26
Update internal linting dependencies

bmish:
Had to disable a few new eslint-plugin-node violations.


src/cli/run.js
```diff
@@ -79,7 +79,7 @@ async function run( args, options ) {
 				require( filePath );
 			} catch ( e ) {
 				if ( e.code === "ERR_REQUIRE_ESM" && ( !nodeVint || nodeVint >= 72 ) ) {
-					await import( filePath );
+					await import( filePath ); // eslint-disable-line node/no-unsupported-features/es-syntax
```
Krinkle, line 82: This rule sounds pretty severe, but looks like it is an upstream bug. This syntax is supported since Node 10. The actual functionality is supported (without experimental flag) since Node 12, but that is a run-time check. The syntax is safe to have (unused) on Node 10.

Disabling the rule here seems fine, since we are already covered from a syntax perspective by ESLint's built-in checks, and by the integration tests running on the various Node versions in Travis CI.

Ref https://github.com/mysticatea/eslint-plugin-node/issues/250.

Krinkle:
Confirmed locally that re-generating `package-lock.json` yields the exact same file (no diff).

Krinkle:
Thanks @bmish !

PR #1558, opened 2021-02-26
strict mode compatibility

### LINKED ISSUE #1557: Not strict mode safe
As distributed on NPM, qunit is not strict-mode safe. It relies on global `this`.

The reason this matters is that people are increasingly relying on the browser's native module loader to pull in all dependencies during development (snowpack, vite). These tools can wrap the vast majority of NPM packages up as modules to achieve interoperability, but they can't readily escape the strict mode requirement imposed by modules.

Strict mode has been javascript best practice for twelve years now. All qunit's own source is already authored under strict mode. It's only the published build that relies on non-strict behavior.

This issue could be resolved by publishing an ESM build as discussed in https://github.com/qunitjs/qunit/issues/1551, but it could also be resolved with a smaller change by switching to a different technique for locating `globalThis`.

`globalThis` is now a stable (stage4) ECMAScript feature, and there are [broadly-compatible polyfills](https://github.com/ungap/global-this) for it that take care of cross-engine issues.

ef4:
This fixes #1557 by adopting [globalThis](https://github.com/tc39/proposal-global), and polyfilling it for environments that don't implement it yet via `@ungap/global-this`.

### DIFF
--- a/lib/global-this-polyfill.js
+++ b/lib/global-this-polyfill.js
@@ -1,0 +1,49 @@
+/*
+https://github.com/ungap/global-this/blob/v0.4.4/esm/index.js
+
+Copyright (c) 2020, Andrea Giammarchi, @WebReflection
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE
+OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+PERFORMANCE OF THIS SOFTWARE.
+
+-------
+
+Patches for use in QUnit:
+
+- 2021-02-25: Export as module only, don't change global scope as QUnit must not
+  affect the host context (e.g. people may test their application intentionally
+  with different or no polyfills and we must not affect that).
+
+*/
+
+let foundGlobalThis;
+
+(function (Object) {
+  if (typeof globalThis === "object") {
+    foundGlobalThis = globalThis;
+  } else {
+    this
+      ? get()
+      : (Object.defineProperty(Object.prototype, "_T_", {
+          configurable: true,
+          get: get,
+        }),
+        _T_);
+
+    function get() {
+      foundGlobalThis = this || self;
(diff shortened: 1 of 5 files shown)


rwjblue (changes requested):
In general, we need to avoid poly filling in tests only. This PR as it stands could enable your tests to pass in circumstances that would not work when QUnit itself wasn't loaded.  For example, imagine your application code could access `globalThis` but forgot to bring the polyfill along with it. With the current implementation in this PR your tests would pass still, but if you deployed to environments without a naturally available `globalThis` your application would crash.

Since the code for the polyfill is *sooo* small, it might just be easier to put it alongside our [promise polyfill](https://github.com/qunitjs/qunit/blob/master/lib/promise-polyfill.js) instead of using `@ungap/global-this` (which mutates the environment to define `globalThis`).

I think you could get away with something like:

```js
/*
https://github.com/ungap/global-this/blob/v0.4.4/esm/index.js

Copyright (c) 2020, Andrea Giammarchi, @WebReflection

Permission to use, copy, modify, and/or distribute this software for any
purpose with or without fee is hereby granted, provided that the above
copyright notice and this permission notice appear in all copies.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE
OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
PERFORMANCE OF THIS SOFTWARE.

-------
Patches for use in QUnit:
- 2021-02-25: Export as module only, don't change global scope as QUnit must not
  affect the host context (e.g. people may test their application intentionally
  with different or no polyfills and we must not affect that).
*/

let foundGlobalThis;

(function (Object) {
  typeof globalThis !== 'object' && (
    this ?
      get() :
      (Object.defineProperty(Object.prototype, '_T_', {
        configurable: true,
        get: get
      }), _T_)
  );
  function get() {
    foundGlobalThis = this || self;
    delete Object.prototype._T_;
  }
}(Object));

export default foundGlobalThis;
```

ef4:
This now incorporates Robert's feedback to no longer leak the polyfill.

rwjblue (approved):

rwjblue:
@Krinkle - Any chance you could review? The strict mode breakage was introduced in https://github.com/qunitjs/qunit/commit/2671d3f6c1c2b72a5fb9299bdce6eab70f2876a6 (moving to `import global from 'global') due to the changes made in https://github.com/qunitjs/qunit/commit/484d3967a8b7d1b481eebf5a88b62b466c6469ce (specifically the fake global module added in the `rollup.config.js`).

Krinkle (approved):

Krinkle:
Tested locally in Firefox, and in IE9-11 via BrowserStack.

rwjblue:
Thank you @Krinkle!

22a:
Thanks for this fix! Is there any ETA on when it will be bundled into an npm release?

Krinkle:
@22a Tomorrow, or else in the weekend.

22a:
@Krinkle Wonderful, thank you!

Issue #869, opened 2015-10-05
Improve module environment inheritance with a Object.create like


leobalter:
Ref https://github.com/jquery/qunit/pull/859#discussion_r40965697

Instead of using QUnit's `extend` method to extend a module's env. inheritance, we should improve it with a Object.create's like inheritance.

Suggested by @gibson042: 

``` js
var createObject = (function( Base ) {
    return function( prototype ) {
        Base.prototype = prototype;
        return new Base();
    };
})( Function() );
…
env = createObject( parentModule.testEnvironment );
```

We could also use `Object.create` adding a [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#Polyfill) that would be used only on IE6~IE8 for the supported browsers.

jzaefferer:
Sounds good to me. Would help to see an actual implementation.

jzaefferer:
For the `Object.create` polyfill, we can inline that without the comments, but with a `// Support ...` comment.

leobalter:
I want to investigate using LoDash's `_.create` as we might use `_.isEqual` in #865

jzaefferer:
Good idea. It'll be a while until I can get back to working on the build to include lodash modules, so feel free to take that over.

Maybe we should start by trying to switch to ES6 modules (again), with rollup? Then add more dependencies from there.

Or maybe using webpack? I had good experience with that on other projects. It would certainly deal well with a mix of internal ES6 modules and external modules, in whatever format.

trentmwillis:
Is this still desired? Since we now support IE9+ we could use `Object.create` directly,  but I'm unsure what benefit this gives us over using the current `extend` implementation.

gibson042:
The biggest reason for this issue was ensuring that every test has a distinct environment based on its module's environment and shared with the entire before/beforeEach queue and afterEach/after stack, and that is already fulfilled per https://github.com/qunitjs/qunit/pull/859#r40982356 . There's still value in actual inheritance over shallow copying, but I think it's exclusive to sugar for advanced asynchronous use cases (and "sugar" in the sense that they can still work without it). For example:
```js
QUnit.module("outer", function( outerModule ) {
	// Load an ES2018 module into this environment, asynchronously
	// cf. https://github.com/tc39/proposal-dynamic-import
	this.dynamicModulePromise = import("./dynamicModule.js")
		.then( module => this.dynamicModule = module );

[...]

PR #1559, opened 2021-03-04
Core: Run before and after hooks with module context

### LINKED ISSUE #1328: Share 'before' hook test environment with child module

* **QUnit version**: 2.6.2
* **How are you running QUnit? (e.g., script, testem, Grunt)**: ember-qunit

## What are you trying to do?

```js
  hooks.before(function() {
    this.hello = 'world';
  });

  test('outer', function(assert) {
    assert.equal(this.hello, 'world'); // pass
  });

  module('child', function() {
    test('inner', function(assert) {
      assert.equal(this.hello, 'world'); // fails
    });
  });
});
```

Expecting nested child modules to share test environment with parent module.

## What actually happened?

Nested child module have it's own test environment object.

raycohen:
The before and after hooks run once per module as long as there is at least one
test in the module. Using environment inheritance allows us to use the module
context in those hooks, which allows reading the expected changes to the
context from a before hook inside nested modules.

This is a breaking change and would need to be part of a major version bump.

Fixes #1328.
Ref #869.

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -39,10 +39,10 @@ function createModule( name, testEnvironment, modifiers ) {
 		ignored: modifiers.ignored || false
 	};
 
-	const env = {};
+	let env = {};
 	if ( parentModule ) {
 		parentModule.childModules.push( module );
-		extend( env, parentModule.testEnvironment );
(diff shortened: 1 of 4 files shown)


raycohen:
replaced by #1762

PR #1567, opened 2021-03-18
Allow throws to gracefully fail when mismatched class is used as matcher function

### LINKED ISSUE #1530: assert.throws() must not call mismatched class as matcher function

* **QUnit version**: 2.13.0 release, also the 2.10.1 release
* **What environment are you running QUnit in? (e.g., browser, Node)**: Firefox 84.0 64-bit on Ubuntu 18.04.5 LTS
* **How are you running QUnit? (e.g., script, testem, Grunt)**: Python script to host the CSS/JS/HTML on `localhost`.

## What are you trying to do?

While testing a `assert.throws()`, I noticed that if the expected value (a class/constructor) didn't match the exception, QUnit would "die" rather than print that the error was the wrong type.

```js
QUnit.test("Example", function (assert) {
    class CustomError extends Error {  }

    assert.throws(
        function(){
            throw new RangeError();
        },
        CustomError,
    );
});
```

If you have any relevant configuration information, please include that here: None that I know of.

Should indicate that an exception was thrown, but it didn't match (failed and `instanceof` check).

## What actually happened?

The test run failed immediately.

What I see as output:
```
Died on test #1 @http://localhost:1000/test.js:1:7
: class constructors must be invoked with 'new'@ 3 ms
Source: 	

throws@http://localhost:1000/qunit-2.13.0.js:3998:61
@http://localhost:1000/test.js:4:12
runTest@http://localhost:1000/qunit-2.13.0.js:3044:32
run@http://localhost:1000/qunit-2.13.0.js:3032:15
runTest/<@http://localhost:1000/qunit-2.13.0.js:3255:15
processTaskQueue@http://localhost:1000/qunit-2.13.0.js:2640:26
processTaskQueue/<@http://localhost:1000/qunit-2.13.0.js:2644:28
promise callback*processTaskQueue@http://localhost:1000/qunit-2.13.0.js:2640:34
processTaskQueue/<@http://localhost:1000/qunit-2.13.0.js:2644:28
promise callback*processTaskQueue@http://localhost:1000/qunit-2.13.0.js:2640:34
advanceTaskQueue@http://localhost:1000/qunit-2.13.0.js:2625:20
advance@http://localhost:1000/qunit-2.13.0.js:2611:4
advanceTestQueue@http://localhost:1000/qunit-2.13.0.js:2670:4
advance@http://localhost:1000/qunit-2.13.0.js:2614:6
unblockAndAdvanceQueue@http://localhost:1000/qunit-2.13.0.js:4314:20
promise callback*begin@http://localhost:1000/qunit-2.13.0.js:4343:9
scheduleBegin/<@http://localhost:1000/qunit-2.13.0.js:4305:8
setTimeout handler*scheduleBegin@http://localhost:1000/qunit-2.13.0.js:4304:18
load@http://localhost:1000/qunit-2.13.0.js:4282:10
EventListener.handleEvent*addEvent@http://localhost:1000/qunit-2.13.0.js:5295:11
@http://localhost:1000/qunit-2.13.0.js:6154:14
@http://localhost:1000/qunit-2.13.0.js:6201:4
@http://localhost:1000/qunit-2.13.0.js:7302:2
```

I believe the issue is [this code](https://github.com/qunitjs/qunit/blob/master/src/assert.js#L328):
```js

			// Expected is a constructor, maybe an Error constructor.
			// Note the extra check on its prototype - this is an implicit
			// requirement of "instanceof", else it will throw a TypeError.
			} else if ( expectedType === "function" &&
				expected.prototype !== undefined && actual instanceof expected ) {
				result = true;

			// Expected is an Error object
			} else if ( expectedType === "object" ) {
```
If the code was switched around to do this in the released JS file, it seems to work for the above check.
```js
	        } else if (expectedType === "function" && expected.prototype !== undefined) {
	          result = (actual instanceof expected); // Expected is an Error object
	        } else if (expectedType === "object") {
```
I was going to setup a PR, but I'm not sure how to write tests that verify a QUnit test page reports a failure...

smcclure15:
Resolves #1530.

As mentioned in the issue, this proposal helps a lot, though we can do more in QUnit 3.0 with explicit ES6 semantics. This is a forward-compatible change to allow those enhancements at a later time.

Added unittests, both positive and negative.
Bashed and verified diagnostics in runs via HTML reporter and CLI runner.

(meta: recommend to "squash" changes on merge since this branch has been around for awhile)


src/assert.js
```diff
@@ -340,9 +340,17 @@ class Assert {
...
 
 			// Expected is a validation function which returns true if validation passed
-			} else if ( expectedType === "function" && expected.call( {}, actual ) === true ) {
-				expected = null;
-				result = true;
+			} else if ( expectedType === "function" ) {
+
+				// protect against accidental semantics which could hard error in the test
+				try {
+					result = expected.call( {}, actual ) === true;
```
Krinkle, line 347: result needs to be a boolean I think, so may need to assin in the `catch` es well as otherwise I think it would remain `undefined`.

Krinkle (changes requested):
LGTM, one nit.

src/assert.js
smcclure15: The `result` var is actually initialized to `false`, so the "nothing matched" path is a big fallthrough to just default to false.

Krinkle: Aye, thanks. I missed that.

Krinkle (approved):

smcclure15:
I don't have write access to merge in, even though it's all approved.
Not sure if this is pending any more action on my part... LMK

Krinkle (approved):

PR #1569, opened 2021-03-19
Core: Add test.each.

ventuno:
WIP implementation of https://github.com/qunitjs/qunit/issues/1568.


smcclure15:
I'm really liking this; both with the client-facing API design and it's clean implementation, IMHO. Thanks for your patience in iterating.

I agree the exact naming can be left as a follow up; this looks like a solid foundational piece to build up the other support that was discussed. Next, allow objects with field names? Then support "flat" (or idk, "primitive") arrays for convenience? I feel like those should be very near-term though separate follow ups, just to round out the feature.

ventuno:
Thanks @smcclure15, I just added support of 1D arrays. Let's tackle "objects with field names" as a fast follow up as I would like to discuss a few examples in https://github.com/qunitjs/qunit/issues/1568 first before working on it. @Krinkle FYI.

src/test.js
```diff
@@ -692,6 +696,49 @@ export function test( testName, callback ) {
...
+	newTest.queue();
+}
+
+function normalizeTable( data ) {
+	const isTable = data.every( Array.isArray );
+	if ( isTable ) {
+		return data;
+	}
+	return data.map( datum => [ datum ] );
+}
```
smcclure15, line 736: Do we need some guards to validate that the user-specified `data` is an Array before we start iterating at all? Thinking of edge cases...

```js
QUnit.test.each('raw object', {foo: 2}, assert => { ... })

QUnit.test.each('string', 'hello world', assert => { ... })
```

Otherwise they'd see ".every is not a function" instead of a tailored message, like "Data provided to QUnit.test.each must be an array."

src/test.js
```diff
@@ -706,33 +753,39 @@ extend( test, {
...
 		} );
+	}
+} );
 
-		newTest.queue();
+
+extend( test.each, {
+	skip: function( testName, data ) {
+		runEach( data, ( datum, i ) => {
+			skip( `${i} ${testName}`, datum );
```
smcclure15, line 782: Do we really need to pass `datum` through here? I think leaving it off (and becoming `undefined`) would be just as well, like how `skip` already doesn't need the callback.

src/test.js
```diff
@@ -692,6 +696,49 @@ export function test( testName, callback ) {
...
 
+function skip( testName, data ) {
+	if ( focused || config.currentModule.ignored ) {
+		return;
+	}
+
+	const test = new Test( {
+		testName: testName,
+		skip: true,
+		params: data
```
smcclure15, line 707: Do we really need to pass `data` all the way through? (there's a few spots that maybe could be leaned down by removing them from calls and letting it cast to `undefined`s)

test/each-only.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<title>QUnit Main Test Suite</title>
```
smcclure15, line 5: might be helpful to rename

test/main/each.js
```diff
@@ -0,0 +1,14 @@
...
+		assert.strictEqual( a + b, result );
+	} );
+	QUnit.test.each( "test.each 1D", [ 1, [], "some" ], function( assert, value ) {
+		assert.true( Boolean( value ) );
+	} );
+} );
+QUnit.module( "test.each.skip", function() {
+	QUnit.test( "do run", function( assert ) { assert.true( true ); } );
+	QUnit.test.each.skip( "test.each.skip", [ [ 1, 2, 3 ], [ 1, 1, 2 ] ], function( assert, a, b, result ) {
+		assert.strictEqual( a + b, -result );
```
smcclure15, line 12: Maybe just `throw` to be super explicit?

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,56 @@
...
+### Description
+
+Add a parametrized test to run using `QUnit.test.each()`.
+
+The `assert` argument to the callback contains all of QUnit's [assertion methods](../assert/index.md). Use this argument to call your test assertions.
+
+See also:
+* [`QUnit.test.only()`](./test.only.md)
+* [`QUnit.test.skip()`](./test.skip.md)
+* [`QUnit.test.todo()`](./test.todo.md)
```
smcclure15, line 37: Might want to reference how `QUnit.test` can handle `then`-ables (or decouple that and somehow say that `each` generates multiple `test` calls).

Also would want to call out `each.skip` and `each.only`.

What about `each.todo`? I'm thinking more for consistency than anything else, to avoid a surprising gap in support.

src/test.js
ventuno: Ah we do not. Removed it.

docs/QUnit/test.each.md
ventuno: Called out `skip` and `only` in these docs and added support for `each.todo`.

test/main/each.js
ventuno: Used the more common `assert.true( false );` which is the more common approach I have seen. Let me know if you still prefer me to `throw` instead.

test/each-only.html
ventuno: Updated here and `each.html` as well. The curse of copy-paste :-).

src/test.js
ventuno: Ah not for sure in this case. Removed here as well. Let me know if it looks better now.

src/test.js
ventuno: That makes sense. I'm `throw`ing an error now, let me know if that is not the preferred way to signal a problem with the test. Please also take a look at the test and let me know if it looks like a good way to test this behavior.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,56 @@
...
+### Examples
+
+A practical example, using the assert argument and no globals.
+
+```js
+function square( x ) {
+  return x * x;
+}
+
+QUnit.test( "square()", [ [ 2, 4 ], [ 3, 9 ] ], assert value, expected => {
```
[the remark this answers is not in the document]
ventuno: Fixed. Thanks :-).

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,56 @@
...
+
+Add a parametrized test to run.
+
+| parameter | description |
+|-----------|-------------|
+| `name` (string) | Title of unit being tested |
+| `data` (Array) | Array of arrays of parameters to be passed as input to each test. Or a 1D array of primitives |
+| `callback` (function) | Function to close over assertions |
+
+#### Callback parameters: `callback( assert, ..args )`:
```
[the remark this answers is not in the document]
ventuno: Fixed. Thanks.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+`QUnit.test.each( name, data, callback )`
+
+Add a parametrized test to run.
+
+| parameter | description |
+|-----------|-------------|
+| `name` (string) | Title of unit being tested |
+| `data` (Array) | Array of arrays of parameters to be passed as input to each test. Or a 1D array of primitives |
```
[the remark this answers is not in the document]
ventuno: Fixed. As these docs don't end with a period I used your phrasing but removed the ending period. Let me know if you want me to update that as well.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,56 @@
...
+excerpt: Add a parametrized test to run.
+categories:
+  - main
+  - async
+version_added: "1.0"
+---
+
+`QUnit.test.each( name, data, callback )`
+
+Add a parametrized test to run.
```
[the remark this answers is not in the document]
ventuno: Ah thanks. For some reason `:set spell` did not highlight that mistake.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,64 @@
...
+### Examples
+
+A practical example, using the assert argument and no globals.
+
+```js
+function square( x ) {
+  return x * x;
+}
+
+QUnit.test.each( "square()", [ [ 2, 4 ], [ 3, 9 ] ], ( assert value, expected ) => {
```
smcclure15, line 51: Missing a comma
```js
(assert, value, expected) =>
```

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,64 @@
...
+
+QUnit.test.each( "isEven()", [ 2, 4 ], ( assert, value ) => {
+  assert.true( isEven( value ), `isEven(${value})` );
+});
+
+QUnit.test.each( "isAsyncEven()", [ 2, 4 ], ( assert, value ) => {
+  return isAsyncEven( value ).then( ( assert, value ) => {
+    assert.true( isAsyncEven( value ), `isAsyncEven(${value})` );
+  } );
+});
```
smcclure15, line 63: These are great to callout, but it would be good to include the `isEven` and `isAsyncEven` implementations for completeness.

src/test.js
```diff
@@ -692,47 +696,112 @@ export function test( testName, callback ) {
...
+	}
+	return data.map( datum => [ datum ] );
+}
+
+function runEach( data, eachFn ) {
+	if ( Array.isArray( data ) ) {
+		normalizeTable( data ).forEach( eachFn );
+	} else {
+		throw new Error(
+			`test.each expects array of arrays or an array of primitives is the expected input.
```
smcclure15, line 757: "primitives _as_ the expected input." ?

src/test.js
```diff
@@ -692,47 +696,112 @@ export function test( testName, callback ) {
...
+	}
+}
+
+extend( test, {
+	todo: function( testName, callback ) {
+		todo( testName, undefined, callback );
+	},
+	skip: function( testName ) {
+		skip( testName );
+	},
```
smcclure15, line 769: Even cleaner,
```js
skip: skip
```
or just `skip,`

src/test.js
```diff
@@ -692,47 +696,112 @@ export function test( testName, callback ) {
...
+	only: function( testName, callback ) {
+		only( testName, undefined, callback );
 	},
-	only: function only( testName, callback ) {
+	each: function( testName, data, callback ) {
 		if ( config.currentModule.ignored ) {
 			return;
 		}
```
smcclure15, line 776: what about if `focused`, like the `test` implementation?

Is there anyway to reuse that logic? Could we have a local `addTestWithData` function that both `test` and `each` leverage? Right now we could make `test` really call `each` to reuse logic, but that smells funny, and I think highlights the need for a lower level helper function.

If this is creeping scope into "pretty code for pretty sake", then I'm okay with that; just looking for where to lean down to reduce edge cases that have already been solved before.

docs/QUnit/test.each.md
ventuno: Fixed, thank you.

docs/QUnit/test.each.md
ventuno: Added an implementation for both `isEven` and `isAsyncEven`.

src/test.js
ventuno: Changed to: "... expects _an_ array of arrays or an array of primitives _as_ ...". Thanks for bringing this up.

src/test.js
ventuno: Done, thanks.

src/test.js
ventuno: That makes a lot of sense. Refactored the code into `addTestWithData`.

smcclure15 (approved):
I am in favor of this change; still want to bash it interactively for edge cases, but the design/implementation looks solid.
I defer to @Krinkle as lead for the go/no-go on this.

ventuno:
@Krinkle any chance you might take a look at this?

Krinkle:
Thanks @ventuno and @smcclure15, this is looking good.

My main concerns at this point are about making sure the ergonomics hold up in relation to these two things:
1. Use of `only`, `skip` and `todo`.
2. (In the near future)  use of module context parameter (ref https://github.com/qunitjs/qunit/issues/978), e.g. the "q" object or otherwise.

By adding `each` as a property of `test` I think these become a bit less intuitive, so I'd like us to explore alternatives explicitly as well. This isn't a hard-no, I'd just like to make sure we think about the options a bit more.

It could be a sibling of `test` (e.g. `testEach`, `testProvide`, etc.). We did something like this in the past with [`QUnit.asyncTest`](https://qunitjs.com/upgrade-guide-2.x/). But, we also have since phased that out in favour of Promise support directly in `QUnit.test`, and otherwise via `assert.async()`. This has the benefit of being in the right place in the hierarchy, I think.

Another way would be to try to squish it into the `test` function's signature directly. This might not be so infeasible, and is also something we've done before. There used to be a [middle parameter of `QUnit.test`](https://qunitjs.com/upgrade-guide-2.x/#replace-expected-number-argument-of-qunittest). This was removed in 2.0 in favour of `assert.expect()`. This is sufficiently far in the past that we could very well make use of this again, I think. I don't know whether "less is more" holds up here, though. Some amount of explicitness could be good, to have at least one word or name to identify the feature with, and to discover it prominently in the API.

So the options would be:

1. `QUnit.test.each( 'name', { data }, ( assert, data ) => {`

This feels a bit out of place in that it isn't a mode of "test" like the other three properties. On the other hand, it is the least disruptive to the API, but placing it in the only "test"-related corner of the API that has precedent for expansion.

2. `QUnit.test( 'name', { data }, ( assert, data ) => {`

This is the leanest option and is also least disruptive. It retains the logic that the two main methods are "module" and "test" which together do everything. The two methods both have various aliases of themselves as properties which invoke them in special modes.

The downside I see with this is that it might be a bit too lean for less experienced developers, and leaves any form of name for the feature to documentation and social constructs, which also make it harder to search for help. There's also potential for difficult linting / typing when the callback accepts variadic parameters in some cases but not others. For ESLint, I'm not worried about this as it should be trivial to detect the presence of a middle parameter and allow extra arguments based on that. For TypeScript, it has support for [overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), so that seems fine.

3. `QUnit.each( 'name', { data }, ( assert, data ) => {`

This one feels structurally correct to me, in that it's not a method that defines a test. Rather, it's a new primitive between "module" and "test" that defines tests on your behalf. The downside is that it breaks the single-object model we have currently, which would make nesting difficult (ref https://github.com/qunitjs/qunit/issues/978).

It basically has to be either squished into the "test" method, or be a property of it.

4. `QUnit.testEach( 'name', { data }, ( assert, data ) => {`

Same as "3" but more verbosely named. I was pondering with this, thinking it might sort or look better in the API for discovery and in IDEs. But, would probably be worse long-term, because `testFoo` and `test.foo` look alike, and we would have both in our API. It's the sort of thing that might make a library feel complicated.

------

So, yeah, I think `test.each` and `test` are good both options. After I reminding myself to think of the test function as not just merely having three aliases under it, but actually being "the" primary function object, it feels less bad to put `each` under there. But, I could also see us going with `test` if others convince me it's not a bad choice. What do y'all think? How would it be if we made it part of the `test` signature?

ventuno:
Thanks @Krinkle for taking the time to look into this. Personally I don't have a strong preference between `test` and `test.each`, but if I had to vote for one, I'd go for `test.each`.

My original intention was to actually provide it as a separate `each` function, but then realized the existing `only` and `skip` were also made moved into `test.only` and `test.skip`, which I really like and decided to use that style.

I really like how explicit `test.each` is as it also matches what [other test frameworks](https://jestjs.io/docs/api#testeachtablename-fn-timeout) do which is always a nice to have when users move from one framework to another to find a similar API.

Taking into account #978 I would say that if we go with the `q` approach (e.g.: Option 2 in https://github.com/qunitjs/qunit/issues/978#issuecomment-723498685) then I wouldn't want to do `q.test.each` but either `q.each` or `q.test` as a variadic function. If `hooks` becomes the `test` function (e.g.: Option 3 in https://github.com/qunitjs/qunit/issues/978#issuecomment-723498685) then my preference goes again for `test.each` in place of variadic `test`.

I personally would really like to be able to use `test.each` sooner than later. So assuming we would like to move forward with some "each" capability before #978 is sorted out, I think `test.each` is the more conservative approach as adding the extra parameter to `test` can always be done later, but it's probably harder to do the inverse.

I hope this helps!

Krinkle:
Alright, let's go with `test.each()` then. Some feedback from a first review pass:

* I think `test.only.each` might be little easier to explain and scan for in code and docs, than `test.each.only`. Also with our current documentation structure this means you can choose `test.only` from the list, and then on those pages we would only explain one additional variant (the `each`), instead of three. It also aligns more with Jest which I guess is easy enough to do at this point.
* I'd like to avoid the variadic arguments if it's all the same to you. Users can still opt to destructure the array if they prefer, but this makes the logic easier to reason about, I think. It just means each data set is passed as second parameter and from there the user can treat it as a named primitive, or a destructured array, or destructured object, or leave it as-is. Especially because I plan to support objects here, it seems potentially confusing that an array turns into already expanded arguments, but an object has be destructured. Sometimes it can be too magical, perhaps, or it might grow on me. Curious what you think.

Thinking out loud for a minute, I believe these would be the six variants we'd then support:

<table>
<tr><th>numbered</th><th>named</th></tr>
<tr>
<td>

```js
QUnit.test.each('primitive', [
  'a',
  'x'
], (assert, str) => {
  // …
});
// primitive #0
// primitive #1

QUnit.test.each('object', [
  { from: 'a', to: 'b' },
  { from: 'x', to: 'y' }
], (assert, { from, to }) => {
  // …
});
// object #0
// object #1

QUnit.test.each('array', [
  ['a', 'b'],
  ['x', 'y']
], (assert, [from, to]) => {
  // …
});
// array #0
// array #1
```
</td><td>

```js
QUnit.test.each('named primitive', {
  foo: 'a',
  bar: 'x'
}, (assert, str) => {
  // …
});
// named primitive foo
// named primitive bar

QUnit.test.each('named object', {
  foo: { from: 'a', to: 'b' },
  bar: { from: 'x', to: 'y' }
}, (assert, { from, to }) => {
  // …
});
// named object foo
// named object bar

QUnit.test.each('named array', {
  foo: ['a', 'b'],
  bar: ['x', 'y']
}, (assert, [from, to]) => {
  // …
});
// named array foo
// named array bar
```
</td></tr></table>

ventuno:
Thanks @Krinkle, I like both ideas, unfortunately we cannot use array destructuring in our own tests, but I guess it is fine as we can still mention that in the docs.
One final thought: I do not believe at this point if the 1D array handling is especially useful or intuitive. Users are giving us in input a 1D array but then have to access each input value as `data[ 0 ]`. We should either handle it specifically or just get rid of it. Thoughts?

Krinkle:
> One final thought: I do not believe at this point if the 1D array handling is especially useful or intuitive. Users are giving us in input a 1D array but then have to access each input value as `data[ 0 ]`. We should either handle it specifically or just get rid of it. Thoughts?

I agree it would be counter-intuitive if a single `value` ended up as `[value]` in the data parameter. Rather than handling this case special, within our special handling, perhaps we could do without special handling altogether? Given we'd support single value, array, and object (which covers all possible values) then perhaps don't need any special normalizing for the data value and can always pass it onwards as-is. Would that work?

We could keep a light iteration to throw for when the data is of an unexpected type, but I'm unsure if we can even formalize what would be "unexpected" here. For example, I could see a non-plain object or array subclass as unintentional if the intention was a plain list to be destructured. On the other hand, that same value could also be intentionally passed as a single data value, e.g. the difference between `[ { a: 1 }, { b: 2 } ]` and `[ new Thing(1), new Thing(2) ]`.

Maybe we need to limit our validation logic to the top-level structure in that case (plain array, or plain object) and pass-through values as-is.

ventuno:
Thanks for the quick reply, @Krinkle. Yes, the only validation at the moment is that the input needs to be an array (see test "test.each fails with non-array input").

Looking again at [Jest's documentation for `test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) it is explicitly mentioned that:
> If you pass in a 1D array of primitives, internally it will be mapped to a table i.e. `[1, 2, 3] -> [[1], [2], [3]]`

So maybe it is still better to be able to do:
```JavaScript
test.each('some test', [ 1, 2, 3 ], function ( assert, [ value ] ) {
 ...
}
```
than
```JavaScript
test.each('some test', [ [ 1 ], [ 2 ] , [ 3 ] ], function ( assert, [ value ] ) {
 ...
}
```
So maybe it would be enough to provide an example and a mention that in the documentation to make this clear to users.

Krinkle:
The normalization that Jest does seems counter-intuitive to me. The test function bodies are generally not re-usable, so I'm a bit confused as to how the user benefits from their non-array being turned into an array. Admittedly, I've not thought about this very long and if you think it's sensible, I'd like to learn more about your thinking behind that. Maybe this is common in other frameworks and that by itself makes it intuitive? (Let me know!)

For myself, I'd intuitively expect each data in the dataset to come through as-is:

```js
test.each('eg', [ 1, 2, 3 ], function ( assert, data ) { // data = 1
test.each('eg', [ /foo/, /bar/ ], function ( assert, data ) { // data = RegExp(/foo/)
test.each('eg', [ [1, 2], [2, 4] ], function ( assert, data ) { // data = [1, 2]
test.each('eg', [ [1, 2], [2, 4] ], function ( assert, [num, double] ) { // num = 1, double = 2
test.each('eg', [ null, undefined, false ], function ( assert, data ) { // data = null
test.each('eg', [ new Cat(42), new Cat(80) ], function ( assert, cat ) { // cat.softness = 42
test.each('eg', [ { num: 1, double: 2 } ], function ( assert, data ) { // data.num = 1
test.each('eg', [ { num: 1, double: 2 } ], function ( assert, { num, double } ) { // num = 1, double = 2
```

ventuno:
I totally agree that it is our specific use-case which makes this less intuitive, because Jest simply uses variadic arguments.

I'm not sure why they need to specify how the list is transformed in the documentation, this feels like an implementation detail.
Maybe they just want to make sure it is absolutely clear the data is simply turned into what would be the required input without a special handling of a 1D array of primitives (e.g.: it is just syntactic sugar).

To reuse the example currently in the docs. I find a lot simpler to write:
```
QUnit.test.each( "isEven()", [ 2, 4 ], ( assert, [ value ] ) => {
  assert.true( isEven( value ), `isEven(${value})` );
});
```
for 1-argument functions. Than:
```
QUnit.test.each( "isEven()", [ [ 2 ], [ 4 ] ], ( assert, [ value ] ) => {
  assert.true( isEven( value ), `isEven(${value})` );
});
```

Unless we would like to support:
```
QUnit.test.each( "isEven()", [ 2, 4 ], ( assert, value ) => {
  assert.true( isEven( value ), `isEven(${value})` );
});
```
which I think is probably the most confusing of all options as it breaks the pattern even more.

Unless this comment changes anything for you, let us keep it simple for now and leave support 1D arrays until a later time. As usual, it is easy to add new features than remove existing ones. What do you think?

ventuno:
Silly me I just realized how pointless is any special handling with the non-variadic approach. Let me update this.

smcclure15:
> Unless this comment changes anything for you, let us keep it simple for now and leave support 1D arrays until a later time. As usual, it is easy to add new features than remove existing ones. What do you think?

Totally agree. I too am still interested in that support, but that negotiation shouldn't block the initial (and conservative) offering. If we error when users pass in input where each element isn't an array, then we're set up for forward compatibility to simply add a new feature without changing any valid behavior and surprising anyone.

ventuno:
@smcclure15 this is actually even simpler than that. As we do not spread the content of the input array any longer, but simply send it as a parameter there is no more need for a special handling for 1D arrays. They are simply passed as one parameter and no destructuring is needed for the "special case".

smcclure15:
I missed that important subtlety!
This is looking really slick, and makes for clean code (and especially understanding) for both test authoring and our implementation and support.

I'm in favor of that direction for both top-level arrays (now with no need to even validate any content types) and objects for convenient iterating and naming.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,74 @@
...
+QUnit.test.each( "square()", [ [ 2, 4 ], [ 3, 9 ] ], ( assert, [ value, expected ] ) => {
+  assert.equal( square( value ), expected, `square(${value})` );
+});
+
+QUnit.test.each( "isEven()", [ 2, 4 ], ( assert, value ) => {
+  assert.true( isEven( value ), `isEven(${value})` );
+});
+
+QUnit.test.each( "isAsyncEven()", [ 2, 4 ], ( assert, value ) => {
+  return isAsyncEven( value ).then( ( assert, value ) => {
```
Krinkle, line 70: The Promise example doesn't look quite right (assert param, second call below).

Krinkle (changes requested):

src/test.js
```diff
@@ -674,63 +674,120 @@ function checkPollution() {
...
-			callback: callback
+	skip: function( testName, data ) {
+		runEach( data, ( _, i ) => {
+			skip( `${i} ${testName}` );
 		} );
+	}
+} );
 
-		newTest.queue();
+extend( test.only, {
```
Krinkle, line 786: I think we'll have just the one property here for now, so might do without `extend()` and assign `test.only.each` instead, perhaps?

src/test.js
```diff
@@ -674,63 +674,120 @@ function checkPollution() {
...
+	each: function( testName, data, callback ) {
```
Krinkle, line 787: LGTM. Should we do the same for `todo` and `skip` (e.g. todo.each, skip.each, so the extra variant is always first?).

Krinkle (changes requested):
I think the index number might look better if appended rather than prepended. I think that produces a test report and CLI output that feels more aligned/organised. This will become more pronounced when we support key, for named dataset objects. What do you think?

My inspiration (or bias) is from PHPUnit:

CSSMin-phpunit-coverage

* [Link to report](https://doc.wikimedia.org/cover/mediawiki-libs-Minify/CSSMin.php.html#263)
* [Link to source code](https://github.com/wikimedia/minify/blob/2.2.2/tests/CSSMinTest.php#L387-L418)

This is quite verbose, but something like it might be usefut. That way, when we attribute something to `Foo bar square valid`, it'll be more obvious to look for "square valid" rather than "Foo bar", in terms of test names in the source code.

In terms of formatting some kind of character to indicate that it is an offset might help as well. E.g. `#0`, `#1` for lists, and perhaps (later) quoting the dataset key like `"Foo bar"`. Alternatively we could pick some sort of separator character like `: ` or `> `.

docs/QUnit/test.each.md
ventuno: Ah `true`. Thanks.

src/test.js
ventuno: Makes sense, done.

src/test.js
ventuno: That makes it a lot cleaner. Done.

ventuno:
@Krinkle: updated the name to append the index prefixed by `#` (see screenshot). I personally like Java's `DataProvider`s style that use `[]`. In this case it would be particularly fitting because `[]` would fit both array indexes and property names when the input is an object.
Screen Shot 2021-05-09 at 6 08 26 PM

For example:
```
test.each('my test', [1, 2], function(assert, v) {
  assert.true(v > 0);
});
```
Current output with array indexes:
```
my test #0
my test #1
```

Suggested output:
```
my test [0]
my test [1]
```

Suggested output with property names:
```
test.each('my test', { foo: 1, bar: 2 }, function(assert, v) {
  assert.true(v > 0);
});
...
my test [foo]
my test [bar]
```

Krinkle:
@ventuno Sounds good, let's go with `[key]` then. I've not seen that before so it does feel a little odd, but it's as good as anything I think.

Actually, using `#` is troublesome for us since we're moving toward embracing TAP more (both for cross-browser testing, and for CLI) where `#` is generally used for adding additional metadata to a test result line, so using `#` could mess that up.

ventuno:
Thanks @Krinkle, just updated it. See attached screenshot.
Screen Shot 2021-05-14 at 4 02 01 PM

Krinkle:
Thanks!

I might do a follow-up to add some spacing in the default theme between `[1]` and `(1)`. And perhaps in QUnit 3 we can come up with something further integrated that might be a bigger visual difference. But, as a new feature adding to what we have, I think this makes sense.

ventuno:
@Krinkle that sounds good to me. Or we can also address this as part of my next PR to support objects as inputs. I also wanted to get your comments on https://github.com/qunitjs/qunit/issues/1568#issuecomment-835627337 so that I know in which direction to start.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,74 @@
+---
+layout: default
+title: QUnit.test.each()
+excerpt: Add a parameterized test to run.
+categories:
+  - main
+  - async
+version_added: "1.0"
```
Krinkle, line 8: Set this to `"unreleased"` for now.

docs/QUnit/test.each.md
```diff
@@ -0,0 +1,74 @@
+---
+layout: default
+title: QUnit.test.each()
+excerpt: Add a parameterized test to run.
+categories:
+  - main
+  - async
```
Krinkle, line 7: I think we should omit it from the `async` category. That's something we do only for `QUnit.test()` for legacy reasons as `asyncTest` used to be there. The other variants aren't under there.

ventuno: Makes sense. Removed.

docs/QUnit/test.each.md
ventuno: Done. Is this going to be updated automatically with a new release?

Krinkle:
I'll land this tomorrow after one final closer look. From glancing it over now, I think maybe the added tests aren't actually being run by either test-on-node, nor the grunt/browser test. I may've missed it, but it seems like it's not wired up. If that's indeed the cast, I suggest the main each test be added to the main list of tests (via test/index.html and the list of test-on-node tests), and then the special one that has to be its own file (each-only) kept as-is but make sure it is in the list of tests that the grunt/browser task iterates over, if it isn't already included by other means.

ventuno:
Thanks @Krinkle, take your time. I [added `test/each.html` and `test/each-only.html` to `Gruntfile.js`](https://github.com/qunitjs/qunit/pull/1569/files#diff-8e7e5f941c67b0cf481a9d650295119d4922b96117e7adc069ba30a81a78c65fL104) and when I run `npm test` I see both tests in the output (see attached screenshot). I may be missing something, just let me know and I'll make the necessary updates.

Screen Shot 2021-05-15 at 6 25 45 PM

docs/QUnit/test.each.md
Krinkle: Not at the moment. I try to remember them, or grep for it, during a release and fill them in as part of the release prep commit. While our release cadence is increasing through past and this year, I don't expect features to accumulate. So, for now, my view of the [Is It Worth The Time?](https://xkcd.com/1205/) table says that it's not worth automating.

Having said that, it seems like a simple thing to automate and so I wouldn't be worried about it complicating our maintenance work. If you feel compelled to give it a try (perhaps as a way to familiarize with the release process), it would likely take the form of an additional step in [build/prep-release.js](https://github.com/qunitjs/qunit/blob/2.15.0/build/prep-release.js).

Krinkle (approved):

docs/QUnit/test.each.md
ventuno: Thanks. Let me look into that after https://github.com/qunitjs/qunit/pull/1614.

src/test.js
```diff
@@ -674,66 +674,122 @@ function checkPollution() {
 	}
 }
 
+function addTestWithData( data ) {
```
Krinkle, line 677: I like this abstraction a lot, did you mean to make more use of it? I went ahead and did so as part of https://github.com/qunitjs/qunit/pull/1620.

ventuno: This actually came from @smcclure15's [recommendation](https://github.com/qunitjs/qunit/pull/1569#discussion_r608754046). Glad it's useful!

PR #1573, opened 2021-03-31
Trim whitespace of filter input

nlfurniss:
Often when pasting in a filter string, a user may copy extra whitespace, causing QUnit to throw an error about no tests matching.

After looking through the tests, I'm not sure the best place to put a unit test. If the maintainers are interested in adding this functionality, I'll figure out the best place to add the test.

### DIFF
--- a/src/html-reporter/html.js
+++ b/src/html-reporter/html.js
@@ -51,6 +51,15 @@ export function escapeText( s ) {
 		unfilteredUrl = setUrl( { filter: undefined, module: undefined,
 			moduleId: undefined, testId: undefined } );
 
+
+	function trim( string ) {
+		if ( typeof string.trim === "function" ) {
+			return string.trim();
+		} else {
+			return string.replace( /^\s+|\s+$/g, "" );
+		}
+	}
+
 	function addEvent( elem, type, fn ) {
 		elem.addEventListener( type, fn, false );
 	}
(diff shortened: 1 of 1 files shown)


Krinkle:
@nlfurniss Thanks for the patch, this looks good to me. Just two things:

1. We support projectst that in turn support older browsers, so we'll need to use a local trim function for now rather than ES5. See https://github.com/qunitjs/qunit/blob/f3d7f93f0ce23541559bace7d5d3af5a3f894ccc/src/html-reporter/html.js#L98 for example. Feel free to move that into a local function for re-use.
2. Sign at <https://cla.js.foundation/qunitjs/qunit?pullRequest=1573> to indicate that you are the author of this code and are okay with releaseing it under our open source license.

Krinkle:
> When i go to that page it says `You have signed the CLA for qunitjs/qunit` at the top 🤷

It seems your Git config author is not associated with your GitHub account. That might be the issue. The system is associated by email address I believe, so you may need to add the address used for your commits (`git show`) to your GH profile. (It can be a hidden/secondary address, doesn't need to be public I think.)

nlfurniss:
> > When i go to that page it says `You have signed the CLA for qunitjs/qunit` at the top 🤷
> 
> It seems your Git config author is not associated with your GitHub account. That might be the issue. The system is associated by email address I believe, so you may need to add the address used for your commits (`git show`) to your GH profile. (It can be a hidden/secondary address, doesn't need to be public I think.)

Very embarrassing, I spelled my email wrong 😳

Krinkle (approved):

Krinkle:
No worries, glad that was all. Thanks again!

I'll cut a patch release within the next few day.

Issue #1407, opened 2019-09-30
js-reporter does not recognize filtered test points when using QUnit.config.filter


muthu90ec:
## Tell us about your runtime:

* **QUnit version**: 2.9.1
* **What environment are you running QUnit in? (e.g., browser, Node)**: node
* **How are you running QUnit? (e.g., script, testem, Grunt)**: qunit cli

## What are you trying to do?
[js-reporter](https://github.com/js-reporters/js-reporters#readme) errors out when QUnit.config.filter is used. QUnit.config.filter works fine, but the printing portion errors out when using js-reporter.

trentmwillis:
Can you provide a reproduction or, at least, a sample console output of what error is occurring? That would make it easier for us to debug and fix.

muthu90ec:
> Can you provide a reproduction or, at least, a sample console output of what error is occurring? That would make it easier for us to debug and fix.
> 
> Related issue from js-reporters project: [js-reporters/js-reporters#113](https://github.com/js-reporters/js-reporters/issues/113). Unsure where the fix will be needed, so leaving both issues open for now.

Here is the test tExample.js:

```
let bpPoint = [': test try bp']
const regexEndAppender = val => "^" + val + "$";
let mapedP = bpPoint.map(regexEndAppender)
QUnit.config.filter = `!/${mapedP.join("|")}/`
var add = require(require('path').resolve('addFunc.js'));
QUnit.test("test try bp", function (assert) {
    assert.ok(add(1,2) === 3, "dummy 1 of 2");
    assert.ok(false, "first test");
});
QUnit.test("test 2", function (assert) {
    assert.ok(add(1,2) === 3, "dummy 1 of 2");
    assert.ok(true, "second test");
});
```
using a simple custom js-reporter as customRep.js:
```
var JsReporters = require('js-reporters');
var runner = JsReporters.autoRegister();
var path = require('path');

runner.on('testEnd', function(test) {
    //Do some stuff
});

runner.on('runStart', function(test) {
   //Do some stuff
});


runner.on('runEnd', function(test) {
    //Do some stuff
});
```

Command used to run:
```
qunit -require customRep.js tExample.js

```

Error produced:

```
TAP version 13
ok 1 test 2
1..1
# pass 1
# skip 0
# todo 0
# fail 0
not ok 2 global failure
  ---
  message: "Cannot read property 'slice' of undefined"
  severity: failed
  actual: {}
  expected: undefined
[...]

PR #1577, opened 2021-04-04
CLI: Upgrade `js-reporters` to 2.0.0

Krinkle:
Fixes https://github.com/qunitjs/qunit/issues/1340.
Fixes https://github.com/qunitjs/qunit/issues/1406.
Fixes https://github.com/qunitjs/qunit/issues/1555.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3913,9 +3913,9 @@
       }
     },
     "js-reporters": {
-      "version": "1.2.3",
-      "resolved": "https://registry.npmjs.org/js-reporters/-/js-reporters-1.2.3.tgz",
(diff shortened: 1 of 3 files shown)


Krinkle:
@smcclure15 I believe the issue you looked at in https://github.com/qunitjs/qunit/issues/1407 is also fixed by this. However I'm not able to reproduce the issue the same way as before. Would appreciate a second look by you as well to make sure it is indeed addressed.

smcclure15:
> @smcclure15 I believe the issue you looked at in #1407 is also fixed by this... Would appreciate a second look by you as well to make sure it is indeed addressed.

Confirmed!
I transplanted js-reporters 2.0.0 alongside our current QUnit 2.11.2 version as well as latest 2.14.1 and the issue looks to be resolved in either case.

Issue #1576, opened 2021-04-04
Error when a hook is invoked inside the wrong module


raycohen:
When a user forgets to provide the `hooks` param in a module callback, they may unintentionally install a hook on a different module than is visually indicated, eg:

```
module('outer', function (hooks) {
  test('one', ...);

  module('inner a', function () {
    hooks.beforeEach(...); // will run before all three tests even though visually nested in "inner a"
    test('two', ...);
  });

  module('inner b', function () {
    test('three', ...);
  });
});
```

There is a [lint rule](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-hooks-from-ancestor-modules.md) for this, but I think it would make sense for QUnit to throw an error when this occurs. Is there any reason a user would want to do this intentionally?

Krinkle:
Yeah, I agree we could flag this automatically. Basically if hooks are called on during the top-level `executeNow()` phase of a module, we would throw an error if it is not for the current inner-most module. This would mean it would also throw if you for some reason intentionally named them differently and accessed it directly:

```js
module('outer', hooks => {
  test('one', …);

  module('inner a', gizmos => {
    gizmos.beforeEach(…);
    hooks.beforeEach(…); // will throw Error
    test('two', …);
  });
});
```

I'd welcome this change in a PR, to land for QUnit 3. Preferably to first as a warning to land in QUnit 2.x, to aid in migration (or fixing of mistakes) prior to the upgrade.

PR #1578, opened 2021-04-05
Core: Throw an error when hooks are added outside of their module

raycohen:
Ref: #1576

For QUnit version 3.0


test/main/modules.js
```diff
@@ -419,4 +419,35 @@ QUnit.module( "QUnit.module", function() {
...
+				this.beforeEachError = e;
+			}
+
+			this.innerHookRan = false;
+			innerHooks.beforeEach( function() {
+				this.innerHookRan = true;
+			} );
+
+			QUnit.test( "calling parent module's beforeEach errors", function( assert ) {
+				assert.strictEqual( this.beforeEachError.message,
```
Krinkle, line 440: I believe this test will fail in a less-than-ideal way if the error is not thrown. It would abort from property `message` on undefined `this.beforeEachError`, instead of on a failed assertion.

Perhaps an approach could be to have the `catch` statement above assign the result of `e.message`. That way over here you have a consistently accesible value that is either undefined or string.

Krinkle (changes requested):
LGTM. Naturally will land a bit later as we're not yet in the homestretch to QUnit 3.0.

But, we could land an alternate version in QUnit 2.x now, with a warning. That way, people can smoothly update to the last 2.x release, and there take their time to address any warnings whilst builds remain passing, and then move to 3.0 without no changes or other surprises.

test/main/modules.js
raycohen: Updated tests, now running the new tests without the change to src/module.js will fail with:

```
>> Testing test/main/modules.js
>>
>> test: calling parent module's beforeEach errors
>> module: Nested module with different hooks variable name
>> message: Correct error thrown by beforeEach
>>   at Object.<anonymous> (/Users/rcohen/code/oss/forks/qunit/test/main/modules.js:442:12)
>>
>> test: calling parent module's beforeEach errors
>> module: Nested module with different hooks variable name
>> message: outer hook beforeEach should not run
>>   at Object.<anonymous> (/Users/rcohen/code/oss/forks/qunit/test/main/modules.js:445:17)
>>
>> test: hooks error when invoked during test execution
>> module: Improperly invoked hooks
>> message: Correct error thrown by beforeEach
>>   at Object.<anonymous> (/Users/rcohen/code/oss/forks/qunit/test/main/modules.js:451:11)
>> 35 tests in 19ms, 30 passed, 2 failed, 3 skipped.
```

Krinkle (approved):

PR #1579, opened 2021-04-06
Build: sort packages

smcclure15:
I'm not sure what caused this, but lately when I do `npm install`, my `package.json` becomes alphabetized.
That's not a bad thing (and I actually love all of the auto-fix steps), but it was extremely confusing while I was making other (completely unrelated) changes to see this side-effect.
Committing this to reduce churn/confusion in local development.

Side note - should I be using `npm ci` instead? Is my workflow out-of-model?


Krinkle:
Yeah, `npm install` has a tendency to use newer or diferent versions of packages than what the previous contributer (and CI) would be using and also involves more online connectivity.

The change to `package.json` is interesting. I've not seen `npm install` do that unless adding or updating top-level dependencies. Might be something from a newer version of npm. Which node/npm versions are you on currently?

smcclure15:
I happen to be on node 14.15.0 and npm 7.5.2.

Krinkle:
@smcclure15 I'm currently using an (older) virtual dev environment with Node 10 LTS and npm 6. I guess npm didn't sort things as often back then. I'll keep the older lock format version for now, but the change is fine indeed since npm 6 doesn't change it back at least.

PR #1580, opened 2021-04-07
Build: fix npm run dev

ventuno:
I noticed that `npm run dev` failed since the `build` task got removed. The original `build` `grunt` task relied on `grunt-rollup` which has been removed.

I am not sure whether this is the best fix or not, but based [on the content of the commit message when `grunt-rollup` was removed](https://github.com/qunitjs/qunit/commit/484d3967a8b7d1b481eebf5a88b62b466c6469ce#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519):
> To keep the rollup config simpler and to allow all upstream
Rollup documentation to apply directly, I've removed the
grunt-rollup indirection and call it rollup directly instead.

it is still preferable to invoke `rollup` directly.

cc @Krinkle, @smcclure15 this may just be me not aware of a better way to run tests in dev mode, so the better fix here may be cleaning up existing `npm` `scripts` and perhaps update the docs.


Krinkle:
Thanks for reminding me. I left this unfinished indeed.

Use `npm test` during development, which builds and then runs tests. Or use`npm run build` to only build (and then e.g. test manually in the browser), or use `npm run test:cli` to only run tests.

There is not a "dev mode" version of QUnit. There is only one build. 🙂  The `npm run dev` entrypoint was mainly to use grunt-watch, but I found that cause issues too many times and have removed them as such, except I left part of it behind. I don't usually use watch mode during development, but one could e.g. keep `rollup` and `qunit` both running in their own native watch modes if you prefer to work that way. I think that works, and if people like that, we could probably add `npm run ` scripts aliases to make that easier to do. E.g. `npm run build:watch` and `npm run test:cli:watch` or some such.

ventuno:
@Krinkle as this seems to require a more structured conversation I created https://github.com/qunitjs/qunit/issues/1581. Do you mind continuing our conversation there?

scripts/watch.js
```diff
@@ -0,0 +1,63 @@
...
+loadConfigFile( path.resolve( process.cwd(), "rollup.config.js" ) ).then(
+	async( { options, warnings } ) => {
+
+		console.log( `We currently have ${warnings.count} warnings` );
+		warnings.flush();
+
+		for ( const optionsObj of options ) {
+			const bundle = await rollup.rollup( optionsObj );
+			await Promise.all( optionsObj.output.map( bundle.write ) );
+		}
```
ventuno, line 42: I got this entire piece of code straight out of the [`rollup` docs](https://rollupjs.org/guide/en/#programmatically-loading-a-config-file).
I tried with and without this `for` and it seems to work just the same. Perhaps one of the reviewers with more experience on `rollup` can comment?

scripts/watch.js
```diff
@@ -0,0 +1,74 @@
...
+	try {
+		const requestedPath = path.format( {
+			root: process.cwd(),
+
+			// path.normalize will remove any "..".
+			// Probably useless as at least Chrome resolves any ".."
+			// _before_ sending the request.
+			// Probably still a good idea to avoid accessing anything
+			// _outside_ of process.cwd() anyways.
+			base: path.normalize( req.url )
```
Krinkle, line 18: At first glance, this did not appear to have the intended effect:

```
> path.normalize('../bar/quuz');
'../bar/quuz'

> path.format({ root: '/tmp/', base: '../bar/quuz' });
'/tmp/../bar/quuz'
```

But I understand now, that it works so long as the normalize input has a leading slash:
```
> path.normalize('/../../bar/quuz');
'/bar/quuz'
```

Krinkle: As such, it seems `path.format()` and its `root` option appear not add benefits or restrictions compared to the simpler `path.join(cwd, normalized)`. Is that right, or is there a different reason for using `path.format()`?

scripts/watch.js
Krinkle: I think this loop is for if you have multiple bundle inputs/outputs, but we currently have just the one. LGTM.

Krinkle:
This patch LGTM. Unless I missed something, this is essentially the equivalent of running `python3 -m http.server 4000` and `rollup -w` side by side.

That's fine by me as a starting point, but I thought you wanted it to run tests as well. In any event, I'm happy to land this as-is. I'd recommend moving the file to `build/` though and naming the command `build:dev` for now. I might do that as part of landing in a couple days. We'll likely do another release next weekend.

Krinkle (approved):

scripts/watch.js
ventuno: Updated to use `path.join`, indeed there was no use for `path.format`, my bad.

scripts/watch.js
ventuno: Thanks for looking into this. I'll leave this as-is unless you have another preference.

ventuno:
@Krinkle, thanks for looking into this. Yes, there's indeed no difference between running the two commands side by side, but I think there's a lot of benefit in running the two commands together. I could have replaced running the small server with `python3 -m http.server 4000`, but I prefer to avoid making assumptions on a local setup if I can help it.

I may look into running tests as well, but as far as this goes it covers ~90% of my needs when working with QUnit. I would be curious to know if you have any thoughts about having livereload as well?

package.json
```diff
@@ -85,6 +85,7 @@
     "test:lint": "eslint --cache .",
     "test:main": "npm run build && grunt test",
     "test:cli": "npm run build && bin/qunit.js test/cli/*.js",
+    "build:dev": "node scripts/watch.js",
```
Krinkle, line 88: This and eslintrc still refer to the old filename. Good to go otherwise!

Krinkle (changes requested):

package.json
ventuno: Ooops.

Krinkle (approved):

PR #1586, opened 2021-04-10
Core: Warn when hooks from a different module are invoked

raycohen:
Ref: #1576

Updated with test coverage via cli test.


Krinkle:
Perhaps a CLI test could work:

```
$ cat test/cli/fixtures/single.js
QUnit.module( "Single", function() {
	QUnit.test( "has a test", function( assert ) {
		assert.push( true, 1, 1, "message here" );
	} );
} );

$ bin/qunit.js test/cli/fixtures/single.js 
TAP version 13
assert.push is deprecated and will be removed in QUnit 3.0. Please use assert.pushResult instead (https://api.qunitjs.com/assert/pushResult).
ok 1 Single > has a test
1..1
# pass 1
# skip 0
# todo 0
# fail 0
```

(That does remind me, this probably isn't valid TAP output...)

Krinkle:
> (That does remind me, this probably isn't valid TAP output...)

Taking a closer look, I _think_ we're fine here actually. On the CLI, tap parsers consume stdout. In the browser, tap parser consume `console.log`. Our warnings go use `console.warn()` which means we're definitely fine in-browser, and on CLI this is channeled by Node.js to stderr rather than stdout, so we should be fine there as well.

```
$ bin/qunit.js test/cli/fixtures/single.js >/dev/null
assert.push is deprecated and will be removed in QUnit 3.0. Please use assert.pushResult instead (https://api.qunitjs.com/assert/pushResult).
```
```
$ bin/qunit.js test/cli/fixtures/single.js 2>/dev/null
TAP version 13
ok 1 Single > has a test
1..1
# pass 1
# skip 0
# todo 0
# fail 0
```

Anyway, so much for that sidebar. Carry on!

Krinkle (approved):

PR #1587, opened 2021-04-10
HTML Reporter: Increase contrast and use richer colors overall

Krinkle:
Tweak our colors away from the semi-greyish colors we had before,
toward more dark/saturated or bright colors.

Also reduce the overall number of distinct colors, reusing the same
ones more often.

* Change most uses of black (`#000000`) to `#0D3349`, which is the very dark blue we already used for the `#qunit-header` background.

* Change text in `#qunit .pass` from `#528CE0` to `#2F68DA`, this is more saturated and appears darker. I also moved its hue slightly from cyan-blue, more toward blue-violet.

* Change the text color for `#qunit-testresult` from `#3B81AF` to the same dark blue as used for module names, `#366097`. The color previously used here was a medium blue that we didn't use for any other text.

* Change the Rerun link from the **near-invisible** `#C2CCD1` to instead inherit the text color of `#qunit .pass` or `#qunit .fail`, e.g. `#2F68DA`. Also, to make it still stand out as special, add an underline. It retains its liberal padding for easy clickability.

Clean up:

* Remove unused color value for `#qunit-header`. Instead set it to that of `#qunit-header a` (it only contains a link), and let the latter inherit this color.
* Replace redundant `#qunit-tests .fail .test-name` and and `#qunit-tests .fail .module-name ` styles.

| Before | After
|--|--
| 01-before | 02-after

Ref https://github.com/qunitjs/qunit/issues/1427.

### DIFF
--- a/src/qunit.css
+++ b/src/qunit.css
@@ -68,7 +68,7 @@
 #qunit-header {
 	padding: 0.5em 0 0.5em 1em;
 
-	color: #8699A4;
+	color: #C2CCD1;
 	background-color: #0D3349;
 
 	font-size: 1.5em;
@@ -80,7 +80,7 @@
 
 #qunit-header a {
 	text-decoration: none;
-	color: #C2CCD1;
+	color: inherit;
 }
 
 #qunit-header a:hover,
@@ -179,7 +179,7 @@
 	border: 1px solid #D3D3D3;
 	border-top: none;
 	border-radius: 0 0 .25em .25em;
-	color: #000;
+	color: #0D3349;
 	background-color: #F5F5F5;
 	z-index: 99;
 }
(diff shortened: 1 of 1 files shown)


smcclure15:
New CSS looks reasonable, and I like the look of the newly styled report!
Does this check off a few more violations from what Axe caught regarding contrast? Either way, a worthwhile improvement.

smcclure15 (approved):

Krinkle:
@smcclure15 Yeah, it gets us closer to Axe compliance.

I carved this commit out from an experimental branch I have in my fork with some ideas for a bigger redesign:

https://github.com/Krinkle/qunit/commits/qunit3-theme
https://twitter.com/TimoTijhof/status/1381829404725219328

PR #1590, opened 2021-04-14
Test: increase code coverage

smcclure15:
Overall line coverage increased from 89.06% to 91.65%.

Specifically on `diff.js`: 88.82 -> 93.88
Specifically on `test.js`: 91.07 -> 98.8

Most of the changes are about adding tests to exercise missed logic.

The edge case of a test for `QUnit.diff` (with the a,b,c,d,x,y sort of mess) was somewhat incidental; I was coming up with different inputs to debug and see how to hit some missed code, and suddenly it hit a LOT of missed code when I had it just right. I trimmed down as much as I could, though I admit the specific path of logic is still not so straight-forward to me, but I feel good that this is now exercised and validated against regression.

I ended up tweaking the `cli/helpers/execute.js` transformation logic, when I added tests to exercise the following expected error message:
"Expected at least one assertion, but none were run - call expect(0) to accept zero assertions."

The "at" and the parenthesis made this look like part of a stack, which was incorrectly transformed and mangled. Adding just another space-character set it apart.

Inside `src/test.js`, I actually removed a few `return` statements, to instead fallthrough and resume the test run (and process the failure). This was previously unhit logic, and exercising it resulted in a hang, instead of a failure. New tests validate this.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -529,11 +529,11 @@ Test.prototype = {
 			if ( objectType( then ) === "function" ) {
 				const resume = internalStop( test );
 				if ( config.notrycatch ) {
-					then.call( promise, function() { resume(); } );
+					then.call( promise, resume );
 				} else {
 					then.call(
 						promise,
-						function() { resume(); },
+						resume,
 						function( error ) {
 							const message = "Promise rejected " +
 								( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
--- a/src/test.js
+++ b/src/test.js
@@ -806,7 +806,6 @@ function internalStart( test ) {
 			"Invalid value on test.semaphore",
 			sourceFromStacktrace( 2 )
 		);
-		return;
 	}
 
 	// Don't start until equal number of stop-calls
@@ -822,7 +821,6 @@ function internalStart( test ) {
 			"Tried to restart test while already started (test's semaphore was 0 already)",
 			sourceFromStacktrace( 2 )
 		);
-		return;
 	}
 
 	// Add a slight delay to allow more assertions etc.
--- a/src/test.js
+++ b/src/test.js
@@ -528,24 +528,24 @@ Test.prototype = {
 			const then = promise.then;
 			if ( objectType( then ) === "function" ) {
 				const resume = internalStop( test );
+				const resolve = function() { resume(); };
 				if ( config.notrycatch ) {
-					then.call( promise, resume );
+					then.call( promise, resolve );
 				} else {
-					then.call( promise, resume,
-						function( error ) {
-							const message = "Promise rejected " +
-								( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
-								" \"" + test.testName + "\": " +
-								( ( error && error.message ) || error );
-							test.pushFailure( message, extractStacktrace( error, 0 ) );
+					const reject = function( error ) {
+						const message = "Promise rejected " +
+							( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
+							" \"" + test.testName + "\": " +
+							( ( error && error.message ) || error );
+						test.pushFailure( message, extractStacktrace( error, 0 ) );
 
-							// Else next test will carry the responsibility
-							saveGlobal();
+						// Else next test will carry the responsibility
+						saveGlobal();
 
-							// Unblock
-							internalRecover( test );
-						}
-					);
+						// Unblock
+						internalRecover( test );
+					};
+					then.call( promise, resolve, reject );
 				}
 			}
 		}
@@ -741,8 +741,6 @@ export function resetTestTimeout( timeoutDuration ) {
 }
 
 // Put a hold on processing and return a function that will release it.
-// This is used when the test returns a Promise, and the returned function
-// from this (to resume) is passed as its "resolve" handler.
 export function internalStop( test ) {
 	let released = false;
 	test.semaphore += 1;
--- a/src/test.js
+++ b/src/test.js
@@ -531,9 +531,7 @@ Test.prototype = {
 				if ( config.notrycatch ) {
 					then.call( promise, resume );
 				} else {
-					then.call(
-						promise,
-						resume,
+					then.call( promise, resume,
 						function( error ) {
 							const message = "Promise rejected " +
 								( !phase ? "during" : phase.replace( /Each$/, "" ) ) +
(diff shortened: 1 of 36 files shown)


src/test.js
```diff
@@ -529,11 +529,11 @@ Test.prototype = {
 			if ( objectType( then ) === "function" ) {
 				const resume = internalStop( test );
 				if ( config.notrycatch ) {
-					then.call( promise, function() { resume(); } );
+					then.call( promise, resume );
```
Krinkle, line 532: LGTM, but worth documenting I think (perhaps at internalStop's return statement) that we're going to potentially pass that callback unrelated foreign arguments and that it must thus never accept any arguments of its own so long as that's the case.

smcclure15:
I added a commit with some more comments.
I'm not entirely sure that captures your concerns...? I wasn't following why it would matter what args were passed to the resume handler, given that they're just ignored.

Krinkle:
> I wasn't following why it would matter what args were passed to the resume handler, given that they're just ignored.

Yeah, they're ignored now. But if I understand correctly, the ultimate caller to this chain is the end-user's code, not ours. Which means we don't control those args indeed. So, if we were to use `internalStop` in the future to accept some kind of parameter in its `resume` function, and we'd pass this in one or two places, that might cause some unexpected downstream behaviour when the end-user's arguments suddenly end up in our `resume` function. This was previously accounted for with the extra function wrapper that just called `resume();` plainly.

I think when removing this, it is now much less obvious what is going on. E.g. I could search the code base for calls to `resume` and `internalStop` and not realize that it is actually already getting called with arguments here.

Anyway, that's what I was thinking originally. I'll take a closer look later this week. I may've been overthinking a little 🙂

smcclure15:
OH! right right right. I 100% agree, and I'll touch this up shortly to repair.
Thank you for that explanation and patience around this.

test/cli/fixtures/hard-error-in-test.js
```diff
@@ -0,0 +1,5 @@
+QUnit.test( "contains a hard error", assert => {
+	assert.async(); // and ensure it can recover
```
Krinkle, line 2: I didn't quite get this line. I assume it was needed to cover a certain branch, but it's not clear to me what it covers from an end-user perspective. Are hard errors from would-be async tests special? If so, perhaps make that part of the file/test name.

test/cli/main.js
```diff
@@ -280,6 +314,94 @@ QUnit.module( "CLI Main", () => {
...
+				await execute( "qunit noglobals/remove-global.js" );
+			} catch ( e ) {
+				assert.pushResult( {
+					result: e.stdout.indexOf( "message: Deleted global variable(s): dummyGlobal" ) > -1,
+					actual: e.stdout + "\n" + e.stderr
+				} );
+			}
+		} );
+
+		QUnit.test( "forgive global variable", async assert => {
```
Krinkle, line 376: Took me a minute to realize why this one should pass since it isn't configured or CLI-option'ed to behave differnet, but it's that `qunit-` internals are allowed to be global indeed. Might be worth adding to the test name, e.g. "forgive qunit DOM globals" or some such.

Krinkle:
> Inside `src/test.js`, I actually removed a few `return` statements, to instead fallthrough and resume the test run (and process the failure). This was previously unhit logic, and exercising it resulted in a hang, instead of a failure. New tests validate this.

Nice catch. I didn't get a hang when I tried it, but it did indeed fail harder than is neccecary. The error message provide gets lost and with this change it actually makes it through to the end-user.

Before:
```
nobody@:/qunit$ bin/qunit.js tmp.js 
TAP version 13
Error: Process exited before tests finished running
nobody@:/qunit [exit 1] $
```

After:
```
nobody@:/qunit$ bin/qunit.js tmp.js 
TAP version 13
not ok 1 Single > semaphore is set to NaN
  ---
  message: Invalid value on test.semaphore
  severity: failed
  actual  : null
  expected: undefined
  stack: |
        at resume (/qunit/qunit/qunit.js:2788:7)
  ...
1..1
# pass 0
# skip 0
# todo 0
# fail 1
nobody@:/qunit [exit 1] $
```

test/cli/main.js
```diff
@@ -322,4 +444,55 @@ QUnit.module( "CLI Main", () => {
...
+			try {
+				const result = await execute( command );
+				assert.pushResult( {
+					result: true,
+					actual: result.stdout
+				} );
+			} catch ( e ) {
+				assert.pushResult( {
+
+					// This isn't the *exact* the error message we want to see - it's been transformed
```
Krinkle, line 475: one too many "the"s here, I think.

test/cli/main.js
```diff
@@ -322,4 +444,55 @@ QUnit.module( "CLI Main", () => {
...
+					// by the execution to standardize on formatting. Should be improved.
```
Krinkle, line 476: Is this still the case, or does your normalize fix make it match the real one? It does not look transformed to me.

Krinkle (changes requested):
I've given it a full review. Looks great. Left a few (probably last) questions before landing, otherwise good to go!

test/cli/main.js
smcclure15: It was better, but still wrangling it. I'm left with a stack like this:
```
      stack: |
            at Object.<anonymous> (/qunit/test/cli/fixtures/assert-expect/no-assertions.js:1:7)
            at internal
            at run (/qunit/src/cli/run.js:79:5)
```
and that `run.js:79:5` is going to be fragile. That's not getting the "internal" escape from the regex since it starts with a forward-slash:

https://github.com/qunitjs/qunit/blob/main/test/cli/helpers/execute.js#L25
```js
.replace( / {2}at .+\([^/)][^)]*\)/g, "  at internal" )
```

I loosened that requirement, but that broke other verifications, so I left it alone here. I added some more straightforward comments; what I had was admittedly a bit cryptic, even to myself one week later :-/

smcclure15:
I believe I repaired one flaky test which should be producing the `TypeError: Cannot read property 'length' of undefined` error. I think sometimes the process was too fast (or is it slow?) for that error to not be hit in the right spot - I delayed the callback a little to make it more robust.
Following that, I see other "watch"-related failures, but I want to say that those are independently flaky, and independent of my changes...

Krinkle:
@smcclure15 Yeah, the watch test has been flaky on macOS for a while, including on Travis CI. I've not been able to reproduce it locally, in either Node 14 or Node 10. I've marked it optional in the PR merge requirement and filed https://github.com/qunitjs/qunit/issues/1601 to follow-up.

PR #1598, opened 2021-04-21
add window.QUnit.reset()

izelnakri:
Adds window.QUnit.reset() public API to make qunit state reset-able for future test suite restarts without page refreshes.

Relevant discussion: https://github.com/izelnakri/qunitx/pull/2

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -58,6 +58,28 @@ extend( QUnit, {
 	skip: test.skip,
 	only: test.only,
 
+	restart() {
+		ProcessingQueue.finished = false;
+		globalStartCalled = false;
+		runStarted = false;
+
+		config.queue.length = 0;
+		config.modules.length = 0;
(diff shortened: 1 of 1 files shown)


rwjblue:
> Adds window.QUnit.reset()

Hmm, isn't this adding a function named `restart`?

izelnakri:
I thought reset is a better name because Im not calling this.start() inside the window.QUnit.reset().

Ive intentionally kept it this was because test modules need to be loaded/reevaluated before QUnit.start() and after QUnit.reset()

Krinkle (changes requested):
Thanks so much. I'm setting a negative score here for now to hold this back from my review queue while we continue at https://github.com/izelnakri/qunitx/pull/2#issuecomment-828085110 (or elsewhere/here if you prefer).

PR #1595, opened 2021-04-16
Produce clearer, contextual failures when assert.async callbacks are called incorrectly

smcclure15:
When the `assert.async` callback is invoked outside of its originating tests, we can issue a better message showing what test it came from. This is true for when it is called in another test (and we leverage `pushFailure`) or after all tests have run (and we `throw new Error`).

This was done in tandem with removing the global `onError` handler condition where we were _outside_ of tests, instead of making a new "fake" or "placeholder" sort of test on the fly to report those errors, which caused more issues down the line since that was unexpected as far as "hook lifecycles" are concerned.

A few prior issues contain some key use cases, which highlight the user-facing differences...

## From #1377
Repro:
```js
QUnit.test( "demo", assert => {
	assert.timeout( 10 );
	const done = assert.async();
	assert.true( true );
	setTimeout( done, 20 );
} );
```

#### CLI
```
TAP version 13
not ok 1 demo
  ---
  message: "Test took longer than 10ms; test timed out."
  severity: failed
  actual  : null
  expected: undefined
  stack:     at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
  ...
1..1
# pass 0
# skip 0
# todo 0
# fail 1
/SNIP/qunit.js:2352
        if (module.hooks[handler].length) {
                                  ^

TypeError: Cannot read property 'length' of undefined
    at processHooks (/SNIP/qunit.js:2352:35)
    at Test.hooks (/SNIP/qunit.js:2361:9)
    at runTest (/SNIP/qunit.js:2498:43)
    at advanceTestQueue (/SNIP/qunit.js:1913:20)
    at Object.advance (/SNIP/qunit.js:1863:7)
    at Test.queue (/SNIP/qunit.js:2516:25)
    at test (/SNIP/qunit.js:2743:13)
    at Object.onError (/SNIP/qunit.js:3434:7)
    at process.<anonymous> (/SNIP/qunit/src/cli/run.js:107:9)
    at process.emit (events.js:315:20)
```

#### Browser
Same test results, but with errors in console:
```
qunit.js:3005 Uncaught Error: assert.async callback called after test finished.
    at done (qunit.js:3005)
qunit.js:2352 Uncaught TypeError: Cannot read property 'length' of undefined
    at processHooks (qunit.js:2352)
    at Test.hooks (qunit.js:2361)
    at runTest (qunit.js:2498)
    at advanceTestQueue (qunit.js:1913)
[...]

PR #1599, opened 2021-04-22
Indicate which test drooling assert.async callbacks came from

smcclure15:
This is a more tightly scoped version of #1595, tackling only the `assert.async` callback error handling. The remainder (handling the `onError` and `global failure` scenarios) grew larger than expected, so I'd like to slice this part off for iterative improvements, and circle back on the more generic issues.

The source change is straightforward.
Whenever a "done" callback is fired outside of the originating test, we currently see this:
```
assert.async callback called after test finished.
```

We can tease that apart for 2 distinct use cases.
If we're in another test, we can use `pushFailure` to show:
```
`assert.async` callback from test 'foobar' was called during this test.
```

Or if we're outside of the tests, we can `throw`:
```
`assert.async` callback from test 'foobar' called after tests finished.
```

I added a handful of tests - some to exercise these new paths, and some for related workflows by simply validating current behavior to avoid any unexpected changes, even if the current behavior is less than ideal.

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -88,14 +88,14 @@ class Assert {
 		return function done() {
 
 			if ( config.current === undefined ) {
-				throw new Error( "`assert.async` callback from test '" +
-					test.testName + "' called after tests finished." );
+				throw new Error( "`assert.async` callback from test \"" +
+					test.testName + "\" called after tests finished." );
 			}
 
 			if ( config.current !== test ) {
 				config.current.pushFailure(
-					"`assert.async` callback from test '" +
-					test.testName + "' was called during this test." );
+					"`assert.async` callback from test \"" +
+					test.testName + "\" was called during this test." );
 				return;
 			}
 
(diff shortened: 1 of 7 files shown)


src/assert.js
```diff
@@ -86,8 +86,17 @@ class Assert {
 		const resume = internalStop( test );
 
 		return function done() {
+
+			if ( config.current === undefined ) {
+				throw new Error( "`assert.async` callback from test '" +
```
Krinkle, line 91: Very slight preference for an escaped double quote here to match the formatting of errors from `assert.rejects`.

I might explore doing a larger consistency clean up over these kinds of messages. One of the things I quite liked from another patch recently is placing the tes name on its own line separate from the error. That seems to help both in brevity (problem first, then context), as well as less likely to become difficult to read when test names contain quotes and other symbols, e.g. `assert.async() callback after test finished.\nTest: foo bar's "baz"`. But, that's for another one.

I'll swap the quote here when landing the PR, no amend needed just for that.

PR #1611, opened 2021-05-07
Filter queued modules based on `ignored` bit to handle nested-modules issue

### LINKED ISSUE #1610: QUnit.module.only including surprising tests
Using latest QUnit 2.15.0 via CLI.

Repro:
```js
QUnit.module.only('Module A', () => {
    QUnit.test('Test A', assert => {
        assert.ok(1);
    });
});

QUnit.module('Module B', () => {
    QUnit.test('Test B', assert => {
        assert.ok(0);
    });
});

QUnit.module('Module C', () => {
    QUnit.test('Test C', assert => {
        assert.ok(0);
    });
    QUnit.module('Module D', () => {
        QUnit.test('Test D', assert => {
            assert.ok(0);
        });
    });
});
```

I expect that only `Test A` is executed, but in fact `Test D` is also run (demonstrated by the failing verification):
```
TAP version 13
ok 1 Module A > Test A
not ok 2 Module C > Module D > Test D
  ---
  message: failed, expected argument to be truthy, was: 0
  severity: failed
  actual  : 0
  expected: true
  stack: |
        at Object.<anonymous> (/SNIP/demo.js:22:11)
  ...
1..2
# pass 1
# skip 0
# todo 0
# fail 1
```

I verified that this does not reproduce in release 2.13.0, introduced in 2.14.0, and could very well be related to #1527.

smcclure15:
Resolves #1610.

The `isParentModuleInQueue` routine wasn't looking at the `ignored` bit on the modules; yes, the parent module is technically in the queue, but it's been filtered and we need to honor that so more modules aren't added into the queue if a prior module used the `only` affordance.

Qualification: Added a small portion into an existing unittest that exercises this fix, and also manually bashed in CLI and browser regarding other "nested" and "flat" scenarios with `only` workflows.

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -10,7 +10,9 @@ import { globalSuite } from "./core";
 const moduleStack = [];
 
 function isParentModuleInQueue() {
-	const modulesInQueue = config.modules.map( module => module.moduleId );
+	const modulesInQueue = config.modules
+		.filter( module => !module.ignored )
+		.map( module => module.moduleId );
 	return moduleStack.some( module => modulesInQueue.includes( module.moduleId ) );
 }
 
--- a/test/cli/fixtures/only/module.js
+++ b/test/cli/fixtures/only/module.js
@@ -45,6 +45,14 @@ QUnit.module.only( "module D", function() {
 	} );
 } );
 
+QUnit.module( "module D1", function() {
+	QUnit.module( "module D2", function() {
+		QUnit.test( "test D3", function( assert ) {
+			assert.true( false, "this test should not run" );
+		} );
+	} );
+} );
+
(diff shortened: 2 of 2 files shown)


Krinkle:
@smcclure15 Nice catch. I've given it a quick glance only in comparison to https://github.com/qunitjs/qunit/pull/1527/files and LGTM, but it's not obvious to me what made it work before. I'll do a proper review and release tomorrow or Sunday.

My guess is that previously the early return in `module()` was preventing the ignored parent from entering the queue, but since #1527 we're letting `processModule()` be called and so that's why `isParentModuleInQueue` needed to be changed. IS that right?

PR #1614, opened 2021-05-16
Support objects as input to test.each

ventuno:
Follow up on https://github.com/qunitjs/qunit/issues/1568#issuecomment-835627337 to add support for objects as input to `test.each`.

Screen Shot 2021-05-16 at 7 09 04 PM

@Krinkle, @smcclure15 let me know what your thoughts are on this change.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -746,10 +746,15 @@ function makeEachTestName( testName, argument ) {
 function runEach( data, eachFn ) {
 	if ( Array.isArray( data ) ) {
 		data.forEach( eachFn );
+	} else if ( typeof data === "object" && data !== null ) {
+		const keys = Object.keys( data );
+		keys.forEach( ( key ) => {
+			eachFn( data[ key ], key );
+		} );
 	} else {
 		throw new Error(
-			`test.each expects an array of arrays or an array of primitives as the expected input.
-${typeof data} was found instead.`
+			`test.each expects an array of arrays, an array of primitives, or an object 
+ as the expected input. ${typeof data} was found instead.`
 		);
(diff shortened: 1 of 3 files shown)


src/test.js
```diff
@@ -746,10 +746,15 @@ function makeEachTestName( testName, argument ) {
...
+	} else if ( typeof data === "object" && data !== null ) {
+		const keys = Object.keys( data );
+		keys.forEach( ( key ) => {
+			eachFn( data[ key ], key );
+		} );
 	} else {
 		throw new Error(
-			`test.each expects an array of arrays or an array of primitives as the expected input.
-${typeof data} was found instead.`
+			`test.each expects an array of arrays, an array of primitives, or an object
```
Krinkle, line 756: I think the array/primitive distinction might be best omitted here now that we don't use variadic args. Something like "... expect an array or object as input."

Krinkle (changes requested):
LGTM 👍. I've fiddled with the docs in b3521d052ba9d990d748a610a53645f621a4318f, so I'm afriad this needs a rebase (perhaps make the "object as data" case its own example?), but otherwise good to go!

src/test.js
ventuno: After rebasing with your changes, rewrote this as "Array or object of data values passed to each test case" let me know if that works for you.

ventuno:
@Krinkle, thanks do you mind taking another look?

src/test.js
Krinkle: Sounds good to me! I don't see the change applied, however.

PR #1616, opened 2021-05-24
Core: Error when a module callback has a promise as a return value

raycohen:
Ref https://github.com/qunitjs/qunit/issues/1600.

For the QUnit version 3.0 release.

I'd have moved the tests for this into `test/main/modules.js` but I am not able to use async/await syntax nor promises in that file.


Krinkle:
Continuing at https://github.com/qunitjs/qunit/pull/1761.

PR #1617, opened 2021-05-25
Core: Add new `failOnZeroTests` configuration option

brendenpalmer:
This PR adds an optional `failOnZeroTests` configuration that defaults to `true`.

Some context:

For apps using `ember exam` throwing errors when no tests are run can cause issues, since there are cases when running a small number of tests in parallel with multiple browsers where some browsers may not get any tests; however, the test-suite as a whole does have tests. In this case, we prefer to only error when there are no tests at all, not in the case where an individual browsers gets no tests.

### DIFF
--- a/test/cli/main.js
+++ b/test/cli/main.js
@@ -108,9 +108,12 @@ QUnit.module( "CLI Main", () => {
 	} );
 
 	QUnit.test( "exit code is 0 when no tests are run and failOnZeroTests is `false`", async assert => {
-		assert.expect( 0 );
 		const command = "qunit assert-expect/no-tests.js";
-		await execute( command );
+		const execution = await execute( command );
+
+		assert.equal( execution.code, 0 );
+		assert.equal( execution.stderr, "" );
+		assert.equal( execution.stdout, expectedOutput[ command ] );
 	} );
 
 	QUnit.test( "exit code is 1 when no tests exit before done", async assert => {
--- a/test/cli/main.js
+++ b/test/cli/main.js
@@ -107,6 +107,12 @@ QUnit.module( "CLI Main", () => {
 		}
 	} );
 
+	QUnit.test( "exit code is 0 when no tests are run and failOnZeroTests is `false`", async assert => {
+		assert.expect( 0 );
+		const command = "qunit assert-expect/no-tests.js";
+		await execute( command );
+	} );
+
 	QUnit.test( "exit code is 1 when no tests exit before done", async assert => {
 		const command = "qunit hanging-test";
 		try {
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -14,6 +14,10 @@ const config = {
 	// Block until document ready
 	blocking: true,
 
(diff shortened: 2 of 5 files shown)


step2yeung (approved):
Thanks Brenden! This change looks good

test/cli/main.js
```diff
@@ -107,6 +107,12 @@ QUnit.module( "CLI Main", () => {
 		}
 	} );
 
+	QUnit.test( "exit code is 0 when no tests are run and failOnZeroTests is `false`", async assert => {
+		assert.expect( 0 );
+		const command = "qunit assert-expect/no-tests.js";
```
Krinkle, line 112: I think we should also validate the TAP output as for other tests since this is a new use case for the formatter and I can see how that could cause a bug somewhere.

Krinkle (changes requested):
Thx, looks good. One nitpick for the test, but otherwise good to go. I'll make sure this is part of the next release (next Sunday at the latest I hope).

test/cli/main.js
brendenpalmer: @Krinkle there already is an existing test for `No tests were run`: https://github.com/qunitjs/qunit/blob/main/test/cli/main.js#L99. Do you think this is sufficient?

Krinkle: This would be similar, but produce a success instead of a failure. Observing what an empty success looks like, I think, would be interesting and useful to assert. E.g. dealing with zero in these places correctly.

Compare to https://github.com/qunitjs/qunit/blob/2.15.0/test/cli/fixtures/expected/tap-outputs.js#L148-L168

brendenpalmer: Ah, fair enough. Yeah, makes sense, let me update

Krinkle (approved):

Krinkle:
Since this is your first contribution, the CLA permissino and license needs to be confirmed (for QUnit, that is the MIT license).

Sign over at: <https://cla.js.foundation/qunitjs/qunit?pullRequest=1617>

brendenpalmer:
> TAP output looks good; can we confirm the verbiage in the HTML Report looks reasonable? Any special branching required in that env?

@smcclure15 I checked the HTML report in browser (running tests through `ember exam`) and it looked reasonable; the text specifically was:

```
0 tests completed in 4 milliseconds, with 0 failed, 0 skipped, and 0 todo.
0 assertions of 0 passed, 0 failed.
```

rwjblue (approved):

brendenpalmer:
@Krinkle just wanted to follow-up here; when do you think we'll release the next version of `qunit`?

Krinkle:
@brendenpalmer I ran through my checklist last week and noticed the feature was missing docs, and forgot to report back. Sorry about that. I've written it now at https://github.com/qunitjs/qunit/pull/1618.

Have a look if you can. I'll trigger the release pipeline tomorrow or Sunday and push out the latest changes along with it.

PR #1621, opened 2021-06-05
Build: Add version release number in unreleased docs

ventuno:
Following up on [this conversation](https://github.com/qunitjs/qunit/pull/1569#discussion_r633010577): replace "unreleased" with the version number in docs when releasing a new version.

```
git diff                                                          
diff --git a/docs/QUnit/test.each.md b/docs/QUnit/test.each.md                            
index 463537b..52e9ce0 100644                                                                                                                                                        
--- a/docs/QUnit/test.each.md                                                                                                                                                        
+++ b/docs/QUnit/test.each.md                                                             
@@ -4,7 +4,7 @@ title: QUnit.test.each()                                                  
 excerpt: Add tests using a data provider.                                                                                                                                           
 categories:                                                                              
   - main                                                                                 
-version_added: "unreleased"                 
+version_added: "10.0.0"                                                                                                                                                              ---                                         
                                                                                          
 `QUnit.test.each( name, dataset, callback )`<br>
```

### DIFF
--- a/build/prep-release.js
+++ b/build/prep-release.js
@@ -12,10 +12,30 @@ const util = require( "util" );
 const cp = require( "child_process" );
 const gitAuthors = require( "grunt-git-authors" );
 
+function parseLineResults( output = "" ) {
+	return output.trim().split( "\n" );
+}
+
+function versionAddedString( version ) {
+	return `version_added: "${version}"`;
+}
+
 const Repo = {
 	async prep( version ) {
 		if ( typeof version !== "string" || !/^\d+\.\d+\.\d+$/.test( version ) ) {
 			throw new Error( "Invalid or missing version argument" );
+		}
+		{
+			const UNRELEASED_VERSION = versionAddedString( "unreleased" );
+			parseLineResults( cp.execSync(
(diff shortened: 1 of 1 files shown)


Krinkle:
Perfect. Thank you!

Krinkle:
@ventuno Looks like it didn't go exactly as planned. The link at <https://api.qunitjs.com/QUnit/test.each/> is now a 404 Error, looking for `2.16.0.0` instead of `2.16.0`. We don't add new features in patch releases, and as such our doc pages are able to use a version number without trailing `.0`.

Could you take another look?

I think it would be tricky to handle this in the build script since from its perspective there is no right answer (stripping the patch number unconditionally means it would link to the wrong version if something did get released in a patch version, stripping the patch number only if it is `0` would mean the link is broken e.g. to `1.1.1.0`).

Perhaps we can embrace the way you did it here and update the markdown files to use the fulll version instead. That seems simpler in the long run. If you go that route, remember to also update `_includes/version.html` to not append `.0` anymore.

ventuno:
@Krinkle https://github.com/qunitjs/qunit/pull/1625 should fix the issue mentioned above.

PR #1626, opened 2021-06-24
feat(html-reporter): add "Rerun all failed tests"

### LINKED ISSUE #1474: [Feature] Rerun all failed tests
I often find myself wanting just rerun all failed tests, no matter what module they're from. The use case is changing a central piece in a large application, that causes test failures in a lot of places. While fixing these, I would like to focus on the failed tests only and avoid rerunning the full test suite.

I currently evaluate this snippet in the DevTools console to generate such a link:

```js
location.origin + '/tests?testId=' + [
  ...document.querySelectorAll('.fail[id^="qunit-test-output"]')
].map(n => n.id.slice('qunit-test-output-'.length)).join('&testId=')
```

I think adding a proper button for this would be immensely useful.

buschtoens:
Closes #1474.

## New Feature: "Rerun all failed tests" Link

Appends a "Rerun all failed tests" link to the progress report in the `qunit-testresult-display` header, that uses `setUrl({ ... })` to generate a URL that includes the IDs of all failed tests as a list of `testId` query parameters.

### Screenshots

<figure>
  <figcaption>During test execution</figcaption>
  Screenshot 2021-09-05 at 17 43 40
</figure>

<figure>
  <figcaption>After completion</figcaption>
  Screenshot 2021-09-05 at 17 41 13
</figure>

### How?

To achieve this, the `stats.failedTests` property was changed from a `number` to a `string[]` array that keeps track of all failed test IDs.

#### Considered Alternatives

- Keep `stats.failedTests` a `number` and add an extra `failedTestIds: string[]` array.
  I disliked this because of the double book-keeping, which could lead to `stats.failedTests` and `failedTestIds` getting out of sync.
- Lazily query the IDs of all failed tests from the DOM on demand.
  Based on the snippet I included in my original issue #1474, I could have also generated the "Rerun all failed tests" link on demand, like so:
  ```js
  const failedTestIds = [
    ...document.querySelectorAll('.fail[id^="qunit-test-output"]')
  ].map(n => n.id.slice('qunit-test-output-'.length));
  const href = setUrl({ testId: failedTestIds });
  ```
  However, that seemed quite hacky and would require more complex code to intercept interactions with the `<a>` tag to set the `href` attribute in time, where the current `stats.failedTests` solution is very straightforward.

#### Common Utility Functions

Furthermore the registered hooks for `QUnit.done(...)` and `QUnit.testStart(...)` were refactored to use common utility functions:

##### `getProgressHtml(runtime: number, stats: Stats, total: number): string`

- When called from `Qunit.testStart(...)`, the count of completed tests is lower than `total` and the return value looks like this:
  > 209 / 978 tests completed in 398101 milliseconds, with 2 failed, 6 skipped, and 0 todo.
- When called from `Qunit.done(...)`, the count of completed tests is equal to `total` and the return value looks like this:
  > 978 tests completed in 703624 milliseconds, with 2 failed, 6 skipped, and 0 todo.

##### `getAssertionsSummaryHtml(total: number, passed: number, failed: number): string`

Called from `Qunit.done(...)` to print the summary of assertions. Example:
```html
<span class="passed">116</span> assertions of
<span class="total">116</span> passed,
<span class="failed">0</span> failed.
```

##### `getRerunFailedHtml(failedTests: string[]): string`
Called from both `Qunit.done(...)` and `Qunit.testStart(...)` to generate a link to rerun all (so far) failed tests. Example:
```html
<a href="http://localhost:1337/tests?testId=883cd3ae&testId=b2f8d939">
  Rerun all 2 failed tests
</a>
```
Returns an empty string (`""`), if `failedTests` is empty (`[]`).

#### `<br>` Insertion

I also refactored how `<br>` tags are inserted between these fragments. Instead of including the `<br>` in the return value of e.g. `getProgressHtml(...)` and in the fragment arrays, they are inserted by `.filter(Boolean).join("<br">)`, when the fargments are concatenated into the final HTML string, which has the added advantage of conveniently filtering out optional fragments first.

#### Avoid `innerHTML`

I would have also liked to change the inner workings to construct real DOM directly instead of using `element.innerHTML = "..."`, but I did not want to change more than necessary, as the above already are quite significant refactors.

### DIFF
--- a/src/html-reporter/html.js
+++ b/src/html-reporter/html.js
@@ -8,7 +8,7 @@ import fuzzysort from "fuzzysort";
 
 const stats = {
 	passedTests: 0,
-	failedTests: 0,
+	failedTests: [],
 	skippedTests: 0,
 	todoTests: 0
 };
@@ -678,6 +678,52 @@ export function escapeText( s ) {
 		appendInterface();
 	} );
 
+	function getProgressHtml( runtime, stats, total ) {
+		var completed = stats.passedTests +
+			stats.skippedTests +
+			stats.todoTests +
+			stats.failedTests.length;
+
+		return [
+			completed < total ? completed + " / " : "",
+			total,
+			" tests completed in ",
+			runtime,
+			" milliseconds, with ",
+			stats.failedTests.length,
+			" failed, ",
+			stats.skippedTests,
+			" skipped, and ",
+			stats.todoTests,
+			" todo."
+		].join( "" );
+	}
+
+	function getAssertionsSummaryHtml( total, passed, failed ) {
+		return [
+			"<span class='passed'>",
+			passed,
+			"</span> assertions of <span class='total'>",
+			total,
+			"</span> passed, <span class='failed'>",
+			failed,
+			"</span> failed."
+		].join( "" );
+	}
+
+	function getRerunFailedHtml( failedTests ) {
+		if ( failedTests.length === 0 ) {return "";}
+
+		var href = setUrl( { testId: failedTests } );
+		return [
+			"<a href='" + href + "'>",
+			failedTests.length === 1 ?
+				"Rerun 1 failed test" :
+				"Rerun all " + failedTests.length + " failed tests",
+			"</a>"
+		].join( "" );
+	}
+
 	QUnit.done( function( details ) {
 		var banner = id( "qunit-banner" ),
 			tests = id( "qunit-tests" ),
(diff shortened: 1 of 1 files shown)


buschtoens:
I wasn't really sure where to add tests for this. I think https://github.com/qunitjs/qunit/blob/main/test/reporter-html/reporter-html.js is the best candidate, but it doesn't really have broad test coverage to begin with. If you think, it's necessary, I can try adding some test cases for this change in there.

Krinkle:
@buschtoens Thanks, this looks great. I agree with your decision tree and overall approach.

I think a test would be useful here, but perhaps better as a standalone HTML file to make it less subject to change. The test suite would e.g. have 2 fake failing tests with a known test ID, and a third test asserting that the link is generated as expected. It then needs the "usual" trickery with `done()` (see events.js) to replace the (expected) failures with a success.

Krinkle:
Merged in https://github.com/qunitjs/qunit/commit/f5333b27cf4af45b57e147219a596b0beb44df3d.

PR #1627, opened 2021-06-27
Build: Improve error handling in `build:dev` script

Krinkle:
When Rollup finds an error (e.g. syntax error typo in a source file), it provides an Error object with added properties. Some of these are quite large (e.g. watchedFiles), or otherwise less useful. Format the error object into a string, similar to what Rollup normally does.

When Rollup finds an error during the first run, the error is thrown from the manual Rollup API call - before the watcher exists, and thus it causes three problems:

1. the error is reported by Node.js as an unhandled rejection,
2. the watcher won't get created,
3. the process won't exit (given active http server).

Address the first issue by attaching a `catch()` handler.

Address the third issue by explicitly exiting the process if we encounter a generic script error before the watcher is created.

The second issue seemed unavoidable. But, I noticed that there were actually two initial runs. One from the manual Rollup API call, which has no output when it starts or finishes (unless failed), which was confusing to me. And then immediately after that, another "initial" run from the watcher. It would seem that our own initial manual run is not needed then. Alternatively, if we do need it, we could address the "silent" problem by printing "BUNDLE START/END" around that too.

With these changes, the watcher is now created and remains active even if there is already an error to begin with. I kept the `catch()` handler in case there are other issues, such as an issue with the `watch.js` or `rollup.config.js` files. It can be tested by placing someting like `undef()` on the first line of the `then()` callback.

Ref https://github.com/qunitjs/qunit/pull/1580.

### DIFF
--- a/build/watch.js
+++ b/build/watch.js
@@ -3,6 +3,7 @@ const http = require( "http" );
 const path = require( "path" );
 
 const rollup = require( "rollup" );
+const kleur = require( "kleur" );
 const loadConfigFile = require( "rollup/dist/loadConfigFile" );
 
 const server = http.createServer( ( req, res ) => {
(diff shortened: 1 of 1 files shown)


ventuno (approved):
LGTM, thanks for fixing this.

PR #1635, opened 2021-07-05
Assert: Improve rejects/throws validation handling

### LINKED ISSUE #1630: assert.rejects can error if a fat-arrow function is used

* **QUnit version**: 2.16.0
* **Which environment are you using? (e.g., browser, Node)**: Node
* **How are you running QUnit? (e.g., QUnit CLI, Grunt, Karma, manually in browser)**: CLI

When the reject "reason" is a string, both `function` handlers and "fat-arrow" functions work well (both of these pass):
```js
QUnit.test('String reason, function handler', assert => {
    let p = Promise.reject('reason');
    assert.rejects(p, function(err) {
        return err.toString().includes('reason');
    });
});
QUnit.test('String reason, arrow handler', assert => {
    let p = Promise.reject('reason');
    assert.rejects(p, err => {
        return err.toString().includes('reason');
    });
});
```
```
TAP version 13
ok 1 String reason, function handler
ok 2 String reason, arrow handler
1..2
# pass 2
# skip 0
# todo 0
# fail 0
```

However, when the "reason" is an Error object, the function handlers work well, but the fat-arrow's do not:
```js
QUnit.test('Error reason, function handler', assert => {
    let p = Promise.reject(new Error('reason'));
    assert.rejects(p, function(err) {
        return err.toString().includes('reason');
    });
});
QUnit.test('Error reason, arrow handler, sync', assert => {
    let p = Promise.reject(new Error('reason'));
    assert.rejects(p, err => {
        return err.toString().includes('reason');
    });
});
QUnit.test('Error reason, arrow handler, async', async assert => {
    let p = Promise.reject(new Error('reason'));
    await assert.rejects(p, err => {
        return err.toString().includes('reason');
    });
});
```

If I run that as written, I see
```
TAP version 13
ok 1 Error reason, function handler
Error: Process exited before tests finished running
Last test to run (Error reason, arrow handler, sync) has an async hold. Ensure all assert.async() callbacks are invoked and Promises resolve. You should also set a standard timeout via QUnit.config.testTimeout.
```
So it barfed on the second test. If I `skip` the second one, I get:
```
TAP version 13
ok 1 Error reason, function handler
ok 2 # SKIP Error reason, arrow handler, sync
not ok 3 Error reason, arrow handler, async
  ---
  message: "Promise rejected during \"Error reason, arrow handler, async\": Function has non-object prototype 'undefined' in instanceof check"
  severity: failed
  actual  : null
  expected: undefined
  stack: |
    TypeError: Function has non-object prototype 'undefined' in instanceof check
        at Function.[Symbol.hasInstance] (<anonymous>)
  ...
1..3
# pass 1
# skip 1
# todo 0
# fail 1
```

I'm not sure if that sync/async behavior is related, or is a separate issue.

The error message is reminiscent of #1389, and likely a similar fix should resolve it.

smcclure15:
Resolves #1630 and more...

I noticed that `rejects` was also subject to #1567, which was identified and fixed only for `throws`; the solution was the same. Comparing the logic side-by-side, I also noticed that we could improve the "invalid expected type" case, where we used to have:
```js
QUnit.test( "demo, function( assert ) {
    assert.throws(
        () => { throw new Error( "foo" ) },
        2,
        "demo"
    );
} );
```
```
TAP version 13
not ok 1 demo
  ---
  message: "demo"
  severity: failed
  actual  : Error: foo
  expected: 2
  stack: |
        at Object.<anonymous> (/SNIP/demo.js:7:9)
  ...
1..1
# pass 0
# skip 0
# todo 0
# fail 1
```

but now that reads
```
  message: "invalid expected value provided to `assert.throws` callback in \"demo\": number."
  severity: failed
  actual  : Error: foo
  expected: 2
  stack: |
        at Object.<anonymous> (/SNIP/demo.js:7:9)
```

just as `rejects` handled it.

Once all the logic/behavior was more aligned, it was fairly trivial to consolidate the logic into one reusable validation-helper. This should protect us from finding/fixing one and not the other.


Gruntfile.js
```diff
@@ -136,6 +136,10 @@ module.exports = function( grunt ) {
 				"test/cli/fixtures/only/module.js",
 				"test/cli/fixtures/only/module-flat.js",
 
+				"test/es2018/async-functions.js",
+				"test/es2018/rejects.js",
+				"test/es2018/throws.js"
+
 				// FIXME: These tests use an ugly hack that re-opens
```
Krinkle, line 143: (Unrelated to this patch; This is probalby not true anymore as of https://github.com/qunitjs/qunit/pull/1629. I'll make a note at https://github.com/qunitjs/qunit/issues/1511 to revisit this.)

src/assert.js
```diff
@@ -429,61 +390,80 @@ class Assert {
...
+		} catch ( e ) {
+
+			// assign the "expected" to a nice error string to communicate the local failure to the user
+			expected = errorString( e );
+		}
+
+	// Expected is some other invalid type
+	} else {
+		result = false;
+		message = "invalid expected value provided to `assert." + assertionMethod + "` " +
```
Krinkle, line 457: Is it possible to reach this deterministically without the `actual` value (before the callback is invoked)? If so, we may want to `throw` hard here instead of pushing a new/alternate assertion, as we do with other validation errors (such as when a string is passed, earlier on in the `assert.throws()` code).

Krinkle (changes requested):
I think we may want to throw rather than push a failure for this one.

test/es2018/rejects.js
```diff
@@ -0,0 +1,80 @@
+
```
Krinkle, line 1: strip newline :)

src/assert.js
smcclure15: There's one branch checking both:
```js
	} else if ( expectedType === "function" &&
		expected.prototype !== undefined && actual instanceof expected ) {
		result = true;
```
That gives a different "fallthrough" approach than if we just check if it's a function and then calculate the result in the body, because this one comes later:
```js
// Expected is a validation function which returns true if validation passed
} else if ( expectedType === "function" ) {
```

But... that means we can still check if it's a function and know that it wouldn't make it to the "invalid" case. So yes, we can validate the expected before doing anything with the actual!

I can make that change independent of what it does for that case: fail or error. Are you suggesting it should now error? The change would affect code like
```js
assert.throws(fcn, 2)
```
that would fail anyway. Erroring would be really clean and probably better if it's going _ahead_ of the invocation/resolution.

That was verbose, but thinking aloud; just want to make sure I have that all right. I'm onboard with such a change, even in this PR (creeping scope a bit, but it's all cohesive).

Krinkle: Yeah, I think it'd be good to have the validation error be deterministic and thrown early as an exception rather than assertion failure.

smcclure15:
Sounds great, thanks! I'll touch this up, though it might take a week with my schedule.

src/assert.js
```diff
@@ -295,20 +295,9 @@ class Assert {
 		let actual,
 			result = false;
 
-		const currentTest = ( this instanceof Assert && this.test ) || config.current;
+		[ expected, message ] = validateExpectedExceptionArgs( expected, message, "throws/raises" );
```
Krinkle, line 298: It's a bit unrelated, but feel free to omit `raises` from the error message here. We support it as a legacy alias and that doesn't have to go anywhere, but I think keeping it here adds more confusion than it removes. People that had to use it, presumably know that it is an alias and can work it out :)

src/assert.js
```diff
@@ -479,13 +370,98 @@ class Assert {
...
+}
+
+function validateException( actual, expected, message ) {
+	let result = false;
+	const expectedType = objectType( expected );
+
+	// These branches should be exhaustive, based on validation done in validateExpectedException
+
+	// We don't want to validate
+	if ( expected === undefined || expected === null ) {
```
Krinkle, line 422: I haven't confirmed it, but it seems from reading the removed code, that previously a `null` expected value would reach the `else` branch that reports "invalid expected value provided". If so, did you mean to change that?

I didn't actually realize that null (non-object and unboxable primitive) was valid in a throw statement, but it looks like it actually is possible to throw `null` (the same way strings and numbers can be thrown).

I'm neutral on whether we need to support that, but I think the current state might be a bit strange, to have `null` behave the same as no argument, e.g. when would one use `assert.throws(fn, null)` instead of `assert.throws(fn)`?

Krinkle (changes requested):
Left a question regarding null. I might be missing something, let me know what you meant there!

src/assert.js
smcclure15: The old `throws` logic actually used a `!expected` check
([assert.js#L325](https://github.com/qunitjs/qunit/blob/main/src/assert.js#L325)), so undefined, null, false, 0, etc could all make it through here to be "valid" with no matching at all, just that it threw.

The old `rejects` logic used a more explicit `expected === undefined` check ([assert.js#L435](https://github.com/qunitjs/qunit/blob/main/src/assert.js#L435)), so the null/false/0 sort of values would be considered invalid and produce failures. That made for some inconsistencies between the two.

To target the "nullish" values, I went with `expected === undefined || expected === null`. So I've restricted the null/false/0 values for both (which now throw, not fail), but now a `null` expected value for `rejects` has changed from an invalid type to a valid/ignored one. I could go one further and restrict it to _only_ `undefined`, so that a `null` value would error as well for both.

What do you think?

Krinkle: I think any value other than `undefined` should either be considered invalid, or be used as comparison target in some way. Using `null` in this way would be unexpected, in my opinion. I did not realize that `assert.throws()` was implicitly allowing null and other falsely values as if they are undefined. We probably shouldn't change that in a minor release.

Basically, `undefined` is here to represent the case of no `expected` parameter being passed and asserting only that something was thrown. A stricter version of this code would have used `arguments.length` to detect that, instead of comparing to undefined.

Two ideas:
1. In this commit, change both to use `!expected` (assuming empty string remains invalid and handled earlier). That should not cause any previously passing code to start failing, and will temporarily lax `rejects()` to start tolerating any other garbage passed to it. Assumign that people's existing tests are passing, this should not change any behaviour, but will add a temporary blind splot. Then in 3.0 we could change it to  `=== undefined` (or `arguments.length`) and consider unhandled types as invalid.
2. Alternatively, perhaps vary this section of the code by `method` such that the behaviour remains the same for both. The throws branch accepts falsey as alias for undefined, the rejects branch considers non-undefined falsey values as invalid. The throws branch can then be deprecated in a follow-up commit.

Krinkle (changes requested):

Krinkle:
I'll try to summarise what we did and to confirm intentions:

* Validation and matching are now the same between `assert.throws` and `assert.rejects`, this (temporarily) loosens `assert.rejects` to tolerate and silently ignore all falsey `expected` value. We'll deprecate it for both in a follow-up patch.
* Non-function values passed as `block` to `assert.throws()` now result in a simulated assertion failure with descriptive error message, instead of previously where we failed via `block.call()` throwing a TypeError which the user presumably would not have been (wanting to be) matching.

Is that right?

smcclure15:
> I'll try to summarise what we did and to confirm intentions:
> [...]
> Is that right?

Yes; that's a good summary of these sprawling changes, thanks!

PR #1641, opened 2021-07-29
Build: Add copying of CSS to "build:dev" task

Krinkle:
Use node-watch to listen for src/qunit.css, and copy it over with the same preprocess() applied as Rollup and Grunt.

Adjust log messages to differentiate between `[rollup]` and `[css]`.

Remove seemingly redundant `process.on("exit")` listener for cleaning up the watchers since this is afaik only when terminating the process manually which happens just fine (and quietly) with and without this.

Closes https://github.com/qunitjs/qunit/issues/1581.

### DIFF
--- a/build/watch.js
+++ b/build/watch.js
@@ -4,7 +4,10 @@ const path = require( "path" );
 
 const rollup = require( "rollup" );
 const kleur = require( "kleur" );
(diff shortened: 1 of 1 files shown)


ventuno (approved):
LGTM.

PR #1642, opened 2021-07-29
Assert: Improve detection of bad calls to `assert.async()` callbacks

Krinkle:
### Background

When creating two async pauses in a test, it was possible for a test to pass by invoking one of them twice, and the other not at all.

Easy scenario (though perhaps not realistic):

> Use `assert.async()` twice, assigned as done1 and done2 in the same `QUnit.test()` case, and then simulate the failure scenario such that you wrongly call done1 two times, and forget to call done2.

Complex scenario across `QUnit.test()` and "afterEach" hooks, since these previously shared a single semaphore:

> Use `assert.async()` once in a simple test, and schedule the resume call in the future, but then fail with an uncaught error. The uncaught error is found and `Test.run()` would internally kill the pause by resetting the semaphore to zero (this make sense since we shouldn't wait for the release once the test is known to have failed).
> After this reset, we proceed to the "afterEach" hook. Suppose this hook is also async, and during its execution, the originally scheduled resume call happens. This would effectively end up releasing the afterEach's async pause despite not being finished yet, and then we proceed to the next test. That test would then fail when the afterEach's own release call happens, failing as "release during a different test".

This is the scenario of https://github.com/qunitjs/qunit/issues/1432.

Fix this and numerous other edge cases by making the returned callbacks from `assert.async()` strict about which locks they release.

Each lock now adds a unique token to a map, and invoking the release function decrements/removes this token from the map.

### Notes

* es6-map.js assigns the fallback in all browsers.
  This is a bug, to be fixed later.

* The `isNaN(semaphore)` logic was originally added in 2015 by ea3e350f.
  At the time, the internal resume function was public, and NaN could emerge through `QUnit.start("bla")` as result of `semaphore += "bla"`. This has not been possible for a while. During PR #1590, I did not trace the origin of this code, and thus did not realize that it was already obsolete (the semaphore itself is not publicly supported).

* The "during different test" error is now almost impossible to trigger since we now kill pending locks during test failures and
  tolerate all late calls equally. This meant the `drooling-done.js` test case now fails in a more limited way.

  I added a new test case for coverage, that reproduces it still, but it's a lot more obscure – it requires the original test to pass and then also have an unexpected call during a different test.

* I considered using the phrase "async lock" in the public-facing error messages, but found this perhaps too internal/technical when coming from the perspective of `var done = assert.async();`.

  In order to keep the code shared between handling of async-await, Promise, and assert.async, but remain friendly and understandable, I went for the phrase "async pause".

Fixes https://github.com/qunitjs/qunit/issues/1432.

### DIFF
--- a/src/cli/run.js
+++ b/src/cli/run.js
@@ -109,7 +109,7 @@ async function run( args, options ) {
 			console.error( "Error: Process exited before tests finished running" );
 
 			const currentTest = QUnit.config.current;
-			if ( currentTest && currentTest.asyncPauses.size ) {
+			if ( currentTest && currentTest.asyncPauses.size > 0 ) {
 				const name = currentTest.testName;
 				console.error( "Last test to run (" + name + ") has an async hold. " +
 					"Ensure all assert.async() callbacks are invoked and Promises resolve. " +
--- a/src/cli/run.js
+++ b/src/cli/run.js
@@ -109,7 +109,7 @@ async function run( args, options ) {
 			console.error( "Error: Process exited before tests finished running" );
 
 			const currentTest = QUnit.config.current;
-			if ( currentTest && currentTest.semaphore ) {
+			if ( currentTest && currentTest.asyncPauses.size ) {
 				const name = currentTest.testName;
 				console.error( "Last test to run (" + name + ") has an async hold. " +
 					"Ensure all assert.async() callbacks are invoked and Promises resolve. " +
--- a/src/cli/run.js
+++ b/src/cli/run.js
@@ -109,7 +109,7 @@ async function run( args, options ) {
 			console.error( "Error: Process exited before tests finished running" );
 
 			const currentTest = QUnit.config.current;
-			if ( currentTest && currentTest.asyncPauses.size > 0 ) {
+			if ( currentTest && currentTest.pauses.size > 0 ) {
 				const name = currentTest.testName;
 				console.error( "Last test to run (" + name + ") has an async hold. " +
 					"Ensure all assert.async() callbacks are invoked and Promises resolve. " +
--- a/src/html-reporter/es6-map.js
+++ b/src/html-reporter/es6-map.js
@@ -1,14 +1,35 @@
-// Support IE 9-10, PhantomJS: Fallback for fuzzysort.js used by ./html.js
+// Support IE 9-10, Safari 7, PhantomJS: Partial Map fallback.
+// Used by html.js (via fuzzysort.js), and test.js.
+//
+// FIXME: This check is broken. This file is embedded in the qunit.js closure,
+// thus the Map var is hoisted in that scope, and starts undefined (not a function).
 var Map = typeof Map === "function" ? Map : function StringMap() {
 	var store = Object.create( null );
+	var hasOwn = Object.prototype.hasOwnProperty;
 	this.get = function( strKey ) {
 		return store[ strKey ];
 	};
 	this.set = function( strKey, val ) {
+		if ( !hasOwn.call( store, strKey ) ) {
+			this.size++;
+		}
 		store[ strKey ] = val;
 		return this;
 	};
+	this.delete = function( strKey ) {
+		if ( hasOwn.call( store, strKey ) ) {
+			delete store[ strKey ];
+			this.size--;
+		}
+	};
+	this.forEach = function( callback ) {
+		for ( var strKey in store ) {
+			callback( store[ strKey ], strKey );
+		}
+	};
 	this.clear = function() {
 		store = Object.create( null );
+		this.size = 0;
 	};
+	this.size = 0;
 };
--- a/src/test.js
+++ b/src/test.js
@@ -832,8 +832,8 @@ export function resetTestTimeout( timeoutDuration ) {
 // * The test fails due to an uncaught exception.
 //
 //   In this case, Test.run() will call internalRecover() which empties the clears all
-//   async pauses and sets `killed = true`. The killed flag means we silently ignore
-//   any late calls to the resume() callback, as we will have moved on to a different
+//   async pauses and sets the cancelled flag, which means we silently ignore any
+//   late calls to the resume() callback, as we will have moved on to a different
 //   test by then, and we don't want to cause an extra "release during a different test"
 //   errors that the developer isn't really responsible for. This can happen when a test
 //   correctly schedules a call to release(), but also causes an uncaught error. The
@@ -871,13 +871,13 @@ export function internalStop( test, requiredCalls = 1 ) {
 
 	const pauseId = `#${test.asyncNextPauseId++}`;
 	const pause = {
-		killed: false,
+		cancelled: false,
 		remaining: requiredCalls
 	};
 	test.asyncPauses.set( pauseId, pause );
 
 	function release() {
-		if ( pause.killed ) {
+		if ( pause.cancelled ) {
 			return;
 		}
 		if ( config.current === undefined ) {
@@ -920,7 +920,7 @@ export function internalStop( test, requiredCalls = 1 ) {
 						sourceFromStacktrace( 2 )
 					);
 
-					pause.killed = true;
+					pause.cancelled = true;
 					test.asyncPauses.delete( pauseId );
 					internalStart( test );
 				};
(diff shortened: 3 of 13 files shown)


src/test.js
```diff
@@ -810,22 +811,106 @@ export function resetTestTimeout( timeoutDuration ) {
...
+//   This is close to impossible because this error only happens if the original test
+//   succesfully finished first (since other failure scenarios kill pauses and ignore
+//   late calls). It can happen if a test ended exactly as expected, but has some
+//   external or shared state continuing to hold a reference to the release callback,
+//   and either the same test scheduled another call to it in the future, or a later test
+//   causes it to be called through some shared state.
+//
+// * Pause release() is called too often, during the same test.
+//
+//   This is simply throws an error, after which uncaught error handling picks it up
```
smcclure15, line 867:
```suggestion
//   This simply throws an error, after which uncaught error handling picks it up
```

src/test.js
```diff
@@ -810,22 +811,106 @@ export function resetTestTimeout( timeoutDuration ) {
...
+//   and processing resumes.
+export function internalStop( test, requiredCalls = 1 ) {
 	config.blocking = true;
 
-	// Set a recovery timeout, if so configured.
+	const pauseId = `#${test.asyncNextPauseId++}`;
```
smcclure15, line 872: If we're using Maps, can't we just use the integer directly as a key, instead of this stringed version? Or does that cause some memory bloat for ES5 environments?

OH. I see ahead that those are used in the messaging. This makes sense to simply store this ID once to reuse in several places later.

src/test.js
```diff
@@ -810,22 +811,106 @@ export function resetTestTimeout( timeoutDuration ) {
...
+	const pause = {
+		killed: false,
```
smcclure15, line 874: Alternatively? `cancelled`
Coder's choice... not sure which communicates intent more clearly.

src/cli/run.js
```diff
@@ -109,7 +109,7 @@ async function run( args, options ) {
 			console.error( "Error: Process exited before tests finished running" );
 
 			const currentTest = QUnit.config.current;
-			if ( currentTest && currentTest.semaphore ) {
+			if ( currentTest && currentTest.asyncPauses.size ) {
```
smcclure15, line 112:
```suggestion
			if ( currentTest && currentTest.asyncPauses.size > 0 ) {
```
for consistency with your other checks on this

smcclure15 (approved):
I know my domain has some test utilities that sniff at that `semaphore` bit to get really picky about choosing to throw or continue in different scenarios. Some areas use QUnit.log and upon a failure, if they are in a pause situation, they want to jump to the next test so it's not left in a bad state. All of which I anticipate this handles much more gracefully after all.

That `semaphore` logic is completely implementation detail so that's not fair to hold up this PR, nor trigger a major version, just noting that dependency for us and others that may have found their way to relying on that.

src/test.js
Krinkle: Yeah, this is a bit of a remnant. At one point I had a single map shared for all tests and even tried to cram the test name in here, too, for easier debugging, but ended up pushing it back toward the current local approach.

Looking now, I think this could even be a simple a simple array (or a Set in ES6) since we just have zero or more objects (rarely more than 1) that we push into the list and then remove (or indexOf-splice) them later. It's not like it'd be a performance issue for the small scale we have here.

TODO: Use numbers directly, move `#` to the message, same as the rest of the formatting (`[async #${id}]`).

src/html-reporter/es6-map.js
```diff
@@ -1,14 +1,35 @@
-// Support IE 9-10, PhantomJS: Fallback for fuzzysort.js used by ./html.js
+// Support IE 9-10, Safari 7, PhantomJS: Partial Map fallback.
+// Used by html.js (via fuzzysort.js), and test.js.
+//
+// FIXME: This check is broken. This file is embedded in the qunit.js closure,
+// thus the Map var is hoisted in that scope, and starts undefined (not a function).
 var Map = typeof Map === "function" ? Map : function StringMap() {
```
gibson042, line 6: Should we just fix it now?
```suggestion
if ( typeof Map !== "function" ) {
	var Map = function StringMap() {
```

src/test.js
```diff
@@ -25,12 +25,13 @@ import TestReport from "./reports/test";
 export default function Test( settings ) {
 	this.expected = null;
 	this.assertions = [];
-	this.semaphore = 0;
 	this.module = config.currentModule;
 	this.steps = [];
 	this.timeout = undefined;
 	this.data = undefined;
 	this.withData = false;
+	this.asyncNextPauseId = 1;
```
gibson042, line 33: Nit: I think "asyncNextPauseId" would read more naturally as "nextAsyncPauseId" or even "asyncPausesCreated" (defaulting to 0 rather than 1).

src/test.js
```diff
@@ -810,22 +811,106 @@ export function resetTestTimeout( timeoutDuration ) {
...
 	if ( setTimeout ) {
 		let timeoutDuration;
-
 		if ( typeof test.timeout === "number" ) {
 			timeoutDuration = test.timeout;
 		} else if ( typeof config.testTimeout === "number" ) {
 			timeoutDuration = config.testTimeout;
 		}
 
+		// Set a recovery timeout, if so configured.
```
gibson042, line 913:
```suggestion
	// Set a recovery timeout, if so configured.
	if ( setTimeout ) {
		let timeoutDuration;
		if ( typeof test.timeout === "number" ) {
			timeoutDuration = test.timeout;
		} else if ( typeof config.testTimeout === "number" ) {
			timeoutDuration = config.testTimeout;
		}

```

src/test.js
```diff
@@ -834,8 +919,10 @@ export function internalStop( test ) {
 						`Test took longer than ${timeout}ms; test timed out.`,
 						sourceFromStacktrace( 2 )
 					);
-					released = true;
-					internalRecover( test );
+
+					pause.killed = true;
+					test.asyncPauses.delete( pauseId );
```
gibson042, line 924: Should these lines appear before invoking `pushFailure`? I'm wondering if there's any chance of `pushFailure` prompting code that can observe the not-yet-abandoned async pause.

gibson042 (commented):
Nice!

src/html-reporter/es6-map.js
Krinkle: That won't work, it'll still be hosted just the same, right? It's a pretty tough situation. I set it up such that it is included by rollup as `intro` (within the file closure, so as to not leak and be seen by end-user source code), and thus the variable will be seen by fuzzysort.js and other code we use naturally as if it was a global.

Without pulling in `globalThis` into here, I'm not sure how to do this because as soon as you declare anything as `Map` you inherently also deprive your ability to see the outer scope's value for that same reference.

src/test.js
Krinkle: Yep. I'll also make it call `test.pushFailure()` directly here.

src/html-reporter/es6-map.js
gibson042: Oh, right. What's needed is something like the following, which can't be fixed at this level.
```js
(function(…, Map, …) {
	if ( typeof Map !== "function" ) {
		Map = function StringMap() {…};
	}
})(…, typeof Map === "function" && Map, …)
```

src/test.js
Krinkle: I didn't quite get what you meant here. I had moved the inline comment to be closer to where "if configure" happens. Are you suggesting to leave it as-is? Or is there another difference?

gibson042: I hadn't noticed it matched the preexisting pattern because of the new code, but yes. Having the inline comment just before the check for `setTimeout` inherently explains why that check is there.

src/test.js
Krinkle: I had mentally namespaced these two properties under an "async" prefix, so it was meant more as "async: nextPauseId". I've dropped the prefix for now as it wasn't adding much.

src/test.js
Krinkle: I thought of the comment as explaining that `config.timeoutHandler` happens conditionally based on whether a timeout is configured (via one of the two means). The env check for setTimeout seems unrelated, but I don't disagree that reading it up front before mentally entering the overall block is also useful. If/when the block ends up doing multiple things, then I might push to move it back, but anyway, we're not there now. I'll leave it where it was.

PR #1643, opened 2021-08-01
HTML Reporter: Internally use `runStart` and `runEnd` events

Krinkle:
Switch from `QUnit.begin()` to `QUnit.done()` to the `runStart` and `runEnd` events, as well as the data these provide, such as the test counts and runtime.

As part of making the interface less cluttered, I propose we remove the runtime and status-specific counts from the real-time progress text, as these were rapidly changing and are hard to read.

These changes, together, would allow removal of a number of things:

* Remove internal dependency on Test class (in favour of runStart data).
* Remove local counting of test by status (in favour of runEnd data).
* Remove local time measuring.

I've moved the progress line from being the third line in the testresult-display element, to be the first line. This way, the number of completed tests will be in the same place as where it is displayed when the run has ended.

I've also removed the line break between "Running" and the name of the currently running tests. This way the testresult display will generally have the same visual height during and after the test run. (Previously, there was jump ump after the test run, since there are only two lines in the final state, but with the linebreak we had three lines when tests are running.)

Preview of test suite with this PR applied:

| Running... | Done
|--|--
| after-running | after-done

Ref https://github.com/qunitjs/qunit/issues/1486.

### DIFF
--- a/src/html-reporter/html.js
+++ b/src/html-reporter/html.js
@@ -1,17 +1,13 @@
 import QUnit from "../core";
 import Logger from "../logger";
-import Test from "../test";
-import { now, extend, errorString } from "../core/utilities";
+import { extend, errorString } from "../core/utilities";
 import { window, document, navigator, console } from "../globals";
 import "./urlparams";
 import fuzzysort from "fuzzysort";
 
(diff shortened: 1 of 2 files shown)


gibson042 (approved):
LGTM, but I wonder if some kind of clock should remain since we _do_ know for how long the suite has been running.

PR #1644, opened 2021-08-02
Core: Fix bad module nesting when module closure throws global error

Krinkle:
Fixes https://github.com/qunitjs/qunit/issues/1478.

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -61,7 +61,7 @@ function processModule( name, options, executeNow, modifiers = {} ) {
 		options = undefined;
 	}
 
-	let module = createModule( name, options, modifiers );
+	const module = createModule( name, options, modifiers );
 
 	// Move any hooks to a 'hooks' object
 	const testEnvironment = module.testEnvironment;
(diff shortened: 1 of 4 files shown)


src/module.js
```diff
@@ -79,24 +79,30 @@ function processModule( name, options, executeNow, modifiers = {} ) {
...
+					"This will become an error in QUnit 3.0." );
+			}
+		} finally {
+
+			// If the module closure threw an uncaught error during the load phase,
+			// we let this bubble up to global error handlers. But, not until after
+			// we teardown internal state to ensure correct module nesting.
+			// Ref https://github.com/qunitjs/qunit/issues/1478.
+			moduleStack.pop();
+			config.currentModule = module.parentModule || prevModule;
```
Krinkle, line 102: Same as before, we only restore `currentModule` if the module in question is a scoped module. If it is a linear module, then we assign current and leave it as-is. I hope the new structure makes this more obvious. In the old code, this was done a bit strangely.

PR #1646, opened 2021-08-09
Tests: Sync various test suites and add enforce with structural test

Krinkle:
Details in separate commit messages.

Ref #1511.


test/headless.html
```diff
@@ -2,27 +2,33 @@
...
+				for ( var i = 0; i < logs.length; i++ ) {
+					createCallback( logs[ i ] );
+				}
 			}
 		}() );
 	</script>
 </head>
 <body>
+	<!-- There is no div#qunit element, which means there should
+	     not be any default UI UI on this page.
```
smcclure15, line 30:
```suggestion
	     not be any default UI on this page.
```

smcclure15 (approved):
Looks great - was certainly overdue for a re-org.
Thanks for the succinct and separable commits!

test/headless.html
Krinkle: Ack, applied.

PR #1650, opened 2021-09-02
Core: ensure skipped modules' hooks don't leak

### LINKED ISSUE #1649: Skipped child module hooks are leaked

* **QUnit version**: 2.14, 2.16
* **Which environment are you using? (e.g., browser, Node)**: both browser & Node
* **How are you running QUnit? (e.g., QUnit CLI, Grunt, Karma, manually in browser)**: Ember tests & CLI

## What are you trying to do?

Execute a test run that filters out some child modules that have hooks.

```js
import { module, test } from 'qunit';

class Foo {}

  let foo1;

  hooks.beforeEach(function() {
    foo1 = new Foo();
  });

  test('foo1 is ok', function(assert) {
    assert.ok(foo1);
  });

  module('child', function(hooks) {
    let foo2;

    hooks.beforeEach(function() {
      foo2 = new Foo();
    });

    test('foo2 is ok', function(assert) {
      assert.ok(foo2);
    });
  });
});
```

Run with `filter=!child` so the child module is not executed.

`foo1` and `foo2` are garbage collected after the test run because of the code that deletes module hooks out of the global config.

## What actually happened?

They are not because the child module's tests never execute, so the child module never "completes" and its hooks are never deleted out of the global config.

bendemboski:
Make sure that even if a module is entirely skipped, its hooks are deleted out of the global config so that they don't cause data they reference to leak.

Fixes #1649

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -398,8 +398,20 @@ Test.prototype = {
 		function logSuiteEnd( module ) {
 
 			// Reset `module.hooks` to ensure that anything referenced in these hooks
-			// has been released to be garbage collected.
-			module.hooks = {};
+			// has been released to be garbage collected. Descendant modules that were
+			// entirely skipped, e.g. due to filtering, will never have this method
+			// called for them, but might have hooks with references pinning data in
+			// memory (even if the hooks weren't actually executed), so we reset the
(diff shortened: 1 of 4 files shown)


Krinkle (approved):
This is perfect. Thank you!

PR #1656, opened 2021-09-13
Tests: Add integration test with grunt-contrib-qunit

Krinkle:
_Depends on https://github.com/qunitjs/qunit/pull/1655._

-------

Including an end-to-end failure test for https://github.com/qunitjs/qunit/pull/1655.


Krinkle:
After I've patched grunt-contrib-qunit to report/fail on uncaught errors, I'll add an integration test for that as well.

PR #1658, opened 2021-09-17
Core: correct flat module.only usage by marking currentModule as ignored

### LINKED ISSUE #1645: QUnit.module.only including sibling tests when used in the middle of a set

* **QUnit version**: `2.16.0` (CLI and Browser)
* **Which environment are you using? (e.g., browser, Node)**: Tested in both
* **How are you running QUnit? (e.g., QUnit CLI, Grunt, Karma, manually in browser)**: QUnit CLI, and Manually in browser

## What are you trying to do?

```js
QUnit.module('Module A', () => {
    QUnit.test('Test A', assert => {
        assert.ok(0);
    });
    QUnit.module.only('Module B', () => {
        QUnit.test('Test B', assert => {
            assert.ok(1);
        });
    });
    QUnit.test('Test C', assert => {
        assert.ok(0);
    });
});
```

Expected only `Module A > Module B > Test B` to run.

## What actually happened?

Both `Module A > Module B > Test B` and `Module A > Test C` ran evident by the failing assertion.

```
TAP version 13
ok 1 Module A > Module B > Test B
not ok 2 Module A > Test C
  ---
  message: failed, expected argument to be truthy, was: 0
  severity: failed
  actual  : 0
  expected: true
  stack: |
        at Object.<anonymous> (/Users/dweaver/source/foo/foo-test.js:11:16)
  ...
1..3
# pass 2
# skip 0
# todo 0
# fail 1
```

smcclure15:
Resolves #1645.

The first time we process a `module.only` call, we chop the list of tests that have accumulated so far, and then mark everything after that as an early-return to not even process it. But there's the edge-case where we're currently within a nested module, which might add more tests as "siblings" to that module.only usage. By forcefully marking the `currentTest` as `ignored`, we are sure to accurately filter other trailing cases.

Added a unittest by squeezing in a new test case to a prior test for `module.only`. That is admittedly growing in complexity, but I think still serves as a great integration-level test to really stress the combinations.

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -136,11 +136,13 @@ module.only = function( ...args ) {
 	if ( !focused ) {
 		config.modules.length = 0;
 		config.queue.length = 0;
+
+		// aggressively ignore any direct tests in "outer" nested modules
+		config.currentModule.ignored = true;
 	}
 
-	processModule( ...args );
-
 	focused = true;
+	processModule( ...args );
 };
 
 module.skip = function( name, options, executeNow ) {
--- a/test/cli/fixtures/only/module.js
+++ b/test/cli/fixtures/only/module.js
@@ -26,6 +26,10 @@ QUnit.module( "module B", function() {
 		} );
 	} );
 
+	QUnit.test( "test with the only-module as its peer", function( assert ) {
+		assert.true( false, "this test should not run" );
+	} );
+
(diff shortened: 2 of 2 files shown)


Krinkle:
He, I was just working on the same patch. Same src fix, different test. I'll submit mine after this.

Krinkle (approved):

PR #1659, opened 2021-09-18
Tests: Add "only/module-then-test.js" test fixture

Krinkle:
Ref https://github.com/qunitjs/qunit/issues/1645.
Ref https://github.com/qunitjs/qunit/pull/1658.


smcclure15 (approved):
What can I say - good timing!

PR #1668, opened 2022-01-14
assert.propContains API Implementation & Tests

izelnakri:
I think this is a good addition to the existing QUnit public API. It is rather too common that I need to partially check object values. Also includes `notPropContains`.

Example:

```js
assert.propContains(new User({ id: 1, firstName: 'Izel' }), { firstName: 'Izel' });
assert.notPropContains(new User({ id: 1, firstName: 'Izel' }), { firstName: 'Timo' });
```

### DIFF
--- a/src/core/utilities.js
+++ b/src/core/utilities.js
@@ -67,21 +67,52 @@ export function inArray( elem, array ) {
 }
 
 /**
- * Makes a clone of an object using only Array or Object as base,
- * and copies over the own enumerable properties.
+ * Recursively clone an object into a plain array or object, taking only the
+ * own enumerable properties.
  *
- * @param {Object} obj
- * @return {Object} New object with only the own properties (recursively).
+ * @param {any} obj
+ * @param {bool} [allowArray=true]
+ * @return {Object|Array}
  */
-export function objectValues( obj ) {
-	const vals = is( "array", obj ) ? [] : {};
+export function objectValues( obj, allowArray = true ) {
+	const vals = ( allowArray && is( "array", obj ) ) ? [] : {};
 	for ( const key in obj ) {
 		if ( hasOwn.call( obj, key ) ) {
 			const val = obj[ key ];
-			vals[ key ] = val === Object( val ) ? objectValues( val ) : val;
+			vals[ key ] = val === Object( val ) ? objectValues( val, allowArray ) : val;
 		}
 	}
 	return vals;
+}
+
+/**
+ * Recursively clone an object into a plain object, taking only the
+ * subset of own enumerable properties that exist a given model.
+ *
+ * @param {any} obj
+ * @param {any} model
+ * @return {Object}
+ */
+export function objectValuesSubset( obj, model ) {
+
+	// Return primitive values unchanged to avoid false positives or confusing
+	// results from assert.propContains().
+	// E.g. an actual null or false wrongly equaling an empty object,
+	// or an actual string being reported as object not matching a partial object.
+	if ( obj !== Object( obj ) ) {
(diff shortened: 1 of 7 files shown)


Krinkle:
Thanks @izelnakri, I support inclusion of a method like this. In terms of semantics, I believe this would be most useful as sibling to propEqual rather than deepEqual. In other words, own properties only (ignore constructor) and non-recursive, which I believe is how you've implemented it as well.

I'd propose a slightly different name that incorporates the term "prop" for consistency with `propEqual`. I note that in other frameworks, a partial assertion like this is often named with "contain" or "include" as the verb. Perhaps `assert.propContain()` would work here.

Krinkle (approved):
LGTM. I'll probably land/release next weekend. Would you be interested in adding a doc page as well?

izelnakri:
I'm going for a holiday tomorrow, but I'll see if I can do it during my trip :)

Krinkle:
@izelnakri I've landed this now to simplify unrelated code changes I'm doing today.

I made a few judgement calls that I think will align with your intent, but I'd love to hear from you prior to releasing this, I'm  open to changing or reconsidering any aspect of it! Thanks again for the PR.

src/core/utilities.js
```diff
@@ -67,23 +67,54 @@ export function inArray( elem, array ) {
...
+	if ( obj !== Object( obj ) ) {
+		return obj;
+	}
+
+	// Unlike objectValues(), subset arrays to a plain objects as well.
+	// This enables subsetting [20, 30] with {1: 30}.
+	const subset = {};
+
+	for ( const key in model ) {
+		if ( hasOwn.call( model, key ) && hasOwn.call( obj, key ) ) {
```
izelnakri, line 111: Considering we already iterate on model keys with `const key in model`, can this line become the one below?

```
if (hasOwn.call( obj, key ) ) {
 // ...
}
```

> // This enables subsetting [20, 30] with {1: 30}.

I prefer if we don't have this feature actually :) I think this could be a feature that we don't ship, if we dont need to, considering:
- single values would be checked with [index] reference and equal/notEqual or deepEqual.
- subset of an array can be changed from beginning or end with `Array.prototype.slice`. For majority of the users, I suspect they would like to always check a complex/nested object in separate assertion in my view, thats probably 99% of the use cases, wdyt?
- I think this is less readable/maintainable for users: `assert.propContains(pageModelsArray, { 0: izel, 3: timo });` and might be making the test runtime more complex(with extra internal hasOwn checks) and slower(due to more if checks for smt that could be done in multiple assertions faster). 

If we dont support array subsetting, could this line also be removed? If we cant remove it, could it be reduced to single hasOwnProperty check when we don't support array subsetting?

izelnakri, line 111: @Krinkle 

> // This enables subsetting [20, 30] with {1: 30}.

I prefer if we don't have this feature actually :) I think this could be a feature that we don't ship, if we dont need to, considering:
- single values would be checked with [index] reference and equal/notEqual or deepEqual.
- subset of an array can be changed from beginning or end with `Array.prototype.slice`. For majority of the users, I suspect they would like to always check a complex/nested object in separate assertion in my view, thats probably 99% of the use cases, wdyt?
- I think this is less readable/maintainable for users: `assert.propContains(pageModelsArray, { 0: izel, 3: timo });` and might be making the test runtime more complex(with extra internal hasOwn checks) and slower(due to more if checks for smt that could be done in multiple assertions faster). 

If we dont support array subsetting, could this line also be removed? If we cant remove it, could it be reduced to single hasOwnProperty check when we don't support array subsetting?

PR #1673, opened 2022-02-14
Core: Add `QUnit.hooks` to globally add beforeEach and afterEach

Krinkle:
Ref https://github.com/qunitjs/qunit/issues/1475.

-------

Re-created PR due to a GitHub bug not letting me re-open https://github.com/qunitjs/qunit/pull/1670.

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -93,6 +93,10 @@ const config = {
 		}
 	},
 
+	// Exposed to make resets easier
+	// Ref https://github.com/qunitjs/qunit/pull/1598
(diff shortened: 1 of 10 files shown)


src/core/config.js
```diff
@@ -93,6 +93,10 @@ const config = {
 		}
 	},
 
+	// Exposed to make resets easier
+	// Ref https://github.com/qunitjs/qunit/pull/1598
+	globalHooks: {},
```
Krinkle, line 98: \cc @izelnakri FYI :)

PR #1696, opened 2022-07-05
Remove QUnit.extend()

izelnakri:
During my code removals I already saw that this method is deprecated and planned to be removed in v3. This PR could serve as a removal commit(s), otherwise feel free to close the PR ;)


docs/extension/QUnit.extend.md
```diff
@@ -1,49 +0,0 @@
----
-layout: page-api
-title: QUnit.extend()
-excerpt: Copy the properties from one object into a target object.
-groups:
-  - extension
-  - deprecated
```
Krinkle, line 7: I've prepared a `removed` group in the main branch and live at <https://api.qunitjs.com/removed/> where this can go. I'd prefer we keep the docs live to serve for past versions as well as people tend to use older versions for a long time, and we've been fairly good at making that stable and easy to do 🙂

izelnakri: :+1:  done

docs/removed.md
```diff
@@ -1,5 +1,46 @@
 ---
 layout: group
+title: QUnit.extend()
```
Krinkle, line 3: This is the index file for the group, similar to `deprecated.md` and `extension/index.md`. It will list the pages that use that group.

I think removing `extend` from the extension group is fine, but perhaps keep it in that directory for now for its URL continuity. We could possibly move it to a new directory as well for removed things if we want the URL to change (and keep a redirect), but this file is not meant to be an entry by itself :)

izelnakri: Should I create a `removed` folder and move the file there? I removed the docs change commit for now ;)

PR #1698, opened 2022-07-06
Core: Optimize strictEquality checks

izelnakri:
Removes extra if checks, variable mutations and .valueOf() calls from values that dont need it.

After seeing the test results, it seems to be this is an intentional "feature" rather than a bug. What would be the consequences if we remove this feature for v3? Are there really a lot of libraries/apps that depend on this feature?

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -9,19 +9,11 @@ export default (function () {
   let pairs = [];
 
   function useStrictEquality (a, b) {
-    // This only gets called if a and b are not strict equal, and is used to compare on
-    // the primitive values inside object wrappers. For example:
-    // `var i = 1;`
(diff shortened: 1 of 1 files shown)


Krinkle:
I implemented something similar to what we described here in f578596345f8ab5a019bf1e593555ed6bf62a05a. I approached it a bit differently after I realized we could split things `typeof` vs `objectType` thus removing the ambiguity. The cases where we know the value is a primitive string, it now always uses strict equality. The cases where one or both are a boxed primitive (e.g. String object), are now handled by the "object" callback, and I moved the `valueOf` handling to there to maintain support for it.

Boldly closing as such. Don't hestitate to comment if I missed something. More PRs always welcome as well!

PR #1699, opened 2022-07-12
Core: Remove redundant variable allocations for some assertions

izelnakri:


src/assert.js
```diff
@@ -281,16 +277,12 @@ class Assert {
...
-      currentTest.assert.pushResult({
+      return currentTest.assert.pushResult({
         result: false,
         actual: block,
-        message
+        message: 'The value provided to `assert.throws` in ' +
+          '"' + currentTest.testName + '" was not a function.'
       });
-
-      return;
```
Krinkle, line 293: nit-pick: I'd prefer to keep this `return;` so to clarifu that throws is void. With the combined return, it reads to me that something is returned, or perhaps even that `throws()` and `pushResult` share a return responsibility.

Krinkle (approved):
LGTM, but would prefer a small change.

src/assert.js
izelnakri: :+1:  done

PR #1700, opened 2022-08-16
Core: QUnit.equiv inner refactors [Step 1]

izelnakri:
This makes QUnit.equiv code more readable/maintainable and probably faster due to 2 changes:

- inner private functions moved to ES Module cache instead of being registered on initial IIFE call.
- removal of len variable allocation for array checks.

We can also utilize JS `Set` and remove/inline `isContainer` function altogether here to do faster checks, I left this out from this PR because linter gives a warning telling me `compat/compat: Set is not supported in iOS Safari 7.0-7.1, IE 9`:
https://github.com/qunitjs/qunit/blob/main/src/equiv.js#L63

Also JIT compiler might be able to do further optimizations based on other small changes but unlikely. 

I suggest reviewing per commit ;)

### DIFF
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -253,8 +253,7 @@ const callbacks = {
       bProperties.push(i);
     }
 
-    // Ensures identical properties name
-    return typeEquiv(aProperties.sort(), bProperties.sort());
+    return callbacks.array(aProperties.sort(), bProperties.sort());
   }
 };
 
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -1,4 +1,4 @@
-import { objectType } from './core/utilities';
+import { objectType, slice } from './core/utilities';
 
 // Value pairs queued for comparison. Used for breadth-first processing order, recursion
 // detection and avoiding repeated comparison (see below for details).
@@ -298,7 +298,7 @@ function innerEquiv (a, b) {
   }
 
   // ...across all consecutive argument pairs
-  return arguments.length === 2 || innerEquiv.apply(this, [].slice.call(arguments, 1));
+  return arguments.length === 2 || innerEquiv.apply(this, slice.call(arguments, 1));
 }
 
 export default function equiv (...args) {
--- a/src/equiv.js
+++ b/src/equiv.js
@@ -70,9 +70,7 @@ function breadthFirstCompareChild (a, b) {
   if (!isContainer(a)) {
     return typeEquiv(a, b);
   }
-  if (pairs.every(function (pair) {
-    return pair.a !== a || pair.b !== b;
-  })) {
+  if (pairs.every((pair) => pair.a !== a || pair.b !== b)) {
     // Not yet started comparing this pair
     pairs.push({ a: a, b: b });
   }
@@ -88,11 +86,11 @@ const callbacks = {
   symbol: useStrictEquality,
   date: useStrictEquality,
 
-  nan: function () {
+  nan () {
     return true;
   },
 
-  regexp: function (a, b) {
+  regexp (a, b) {
     return a.source === b.source &&
 
       // Include flags in the comparison
@@ -100,11 +98,11 @@ const callbacks = {
   },
 
   // abort (identical references / instance methods were skipped earlier)
-  function: function () {
+  function () {
     return false;
   },
 
-  array: function (a, b) {
+  array (a, b) {
     const len = a.length;
     if (len !== b.length) {
       // Safe and faster
@@ -125,7 +123,7 @@ const callbacks = {
   // repetitions are not counted, so these are equivalent:
   // a = new Set( [ {}, [], [] ] );
   // b = new Set( [ {}, {}, [] ] );
-  set: function (a, b) {
+  set (a, b) {
     if (a.size !== b.size) {
       // This optimization has certain quirks because of the lack of
       // repetition counting. For instance, adding the same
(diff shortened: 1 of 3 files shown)


src/equiv.js
```diff
@@ -103,13 +101,12 @@ const callbacks = {
   },
 
   array (a, b) {
-    const len = a.length;
```
Krinkle: I believe this was meant to account for cases where length is a (custom) getter rather than a plain property or (native) array accessor. Having said that, for custom objects we shouldn't be using this function so probably fine? Just leaving here as FYI, this is not a problem.

Krinkle:
> inner private functions moved to ES Module cache instead of being registered on initial IIFE call.

Can you elaborate what you mean by this? I understand ES Module cache as being about files executed via `import` statement. I'm not aware of a cache relating to local (not exported) functions declared in a JS file.

I do note as reminder that these are (currently) inlined and compiled to ES5, which may or may not affect the outcome. Perhaps you're measuring this at the distribution layer, e.g. importing the qunit.js artefact and noticing an improvement there by letting it compile these upfront instead of lazily. To my knowledge, JS engines nowadays defer compilation of declared functions until they first run, but that they have special awareness of the IIFE paradigm and eagerly compile those during the top-level file compilation e.g. not re-entering the compiler when the top-level executes the `()` part.

If I understand correctly, this is intended to improve startup time for the script, not runtime for diff operations, correct?

> We can also utilize JS `Set` [but the] linter gives a warning

Ack. If the values are strings only, you can use a StringSet shim that uses native Set when available. I added a StringMap for this as well in src/globals.js. I have a [3-line StringSet implementation](https://github.com/wikimedia/mediawiki/blob/1.38.2/resources/src/startup/mediawiki.loader.js#L22-L32) in MediaWiki that you're welcome to add here.

src/equiv.js
izelnakri: Yes I think this shouldnt be a problem due to a typeOf here: https://github.com/qunitjs/qunit/blob/main/src/equiv.js#L282

src/equiv.js
```diff
@@ -298,7 +298,7 @@ function innerEquiv (a, b) {
   }
 
   // ...across all consecutive argument pairs
-  return arguments.length === 2 || innerEquiv.apply(this, [].slice.call(arguments, 1));
+  return arguments.length === 2 || innerEquiv.apply(this, Array.prototype.slice.call(arguments, 1));
```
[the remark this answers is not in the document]
izelnakri: Good idea implementing it now!

izelnakri:
> Can you elaborate what you mean by this? I understand ES Module cache as being about files executed via import statement. I'm not aware of a cache relating to local (not exported) functions declared in a JS file.

Node.js stores/caches module exports as an object to avoid extra filesystem calls. In node.js commonjs imports this is under `require.cache($path).exports`, in ES modules node.js returns the `Module` object from its internal cache and that object has the exports inside the object: https://nodejs.org/api/esm.html#commonjs-namespaces . 

Here is how for example I circumvent the module cache by using a querystring on an imported file that has its contents changed, only way I could avoid node.js not caching the `import` when the same file changes: https://github.com/izelnakri/qunitx/blob/main/lib/commands/run/tests-in-node.js#L20

There is probably no performance benefit as the cost of calling additional function here is negligible but I think its still better to have the functions more flat, and always return the module object directly instead of doing the computation/mutation on an import.

> Ack. If the values are strings only, you can use a StringSet shim that uses native Set when available. I added a StringMap for this as well in src/globals.js. I have a [3-line StringSet implementation](https://github.com/wikimedia/mediawiki/blob/1.38.2/resources/src/startup/mediawiki.loader.js#L22-L32) in MediaWiki that you're welcome to add here.

:+1: I'll add this to `src/globals.js' just like `StringMap` there in additional commit then!

src/globals.js
```diff
@@ -112,3 +112,25 @@ export const StringMap = typeof g.Map === 'function' &&
       });
     }
   };
+
+export const StringSet = g.Set || function (input) {
```
Krinkle, line 116: When landing, I'll change this one to a `typeof` ternary to match `g.Map`. I tend to not bother adding a type checks for most polyfills (e.g. Array or String prototype methods) as I agree that presence is sufficient. If it's set to something else, that's asking for trouble!

But, global constructors in particular are prone to conflicting with unrelated values. E.g. `<h2 id="Set">Set</h2>` in an HTML fixture will result in `g.Set` coming through the `window` object via WindowProxy as HTMLElement reference. It's one of those legacy features on the web you are likely to run into sooner or later even if you're not (intentionally) looking for it.

<https://html.spec.whatwg.org/#named-access-on-the-window-object>

Krinkle:
I kicked off a browserstack-full run in CI, and noticed that IE11 failed due to an infinite loop.

Screenshot 2022-09-12 at 01 53 37

I narrowed this down to the fact that native `window.Set` is available and used, but did not yet support the `Set(iterable)` constructor parameter. This is similar to `Map` where I left a comment about that. Basically, IE11 will return an empty set when given `new Set(['foo']))` which means all values are seen as non-containers and thus go directly to recursion without checking `pairs` - hence infinite loop. One way is to use `add()` instead. Another is to limit use of the native version to only when it is new enough to support the iterable. I did the later in a follow-up to the draft squash branch. Will check again this week and then land if passing.

PR #1704, opened 2022-09-21
Test: Add benchmark for `QUnit.equiv()`

Krinkle:
Ref https://github.com/qunitjs/qunit/issues/1327.

\cc @izelnakri


izelnakri:
This is amazing! I've skimmed it through, I'm about to read it in-depth soon!

izelnakri:
Deno recently introduced a benchmarking API, should we eventually build a QUnit.bench() API that mimicks the deno's API and makes it usable in browser UI(as a failing test red label if a threshold is given or a different colored label) *in the future*? What do you think? https://deno.land/manual@v1.26.0/tools/benchmarker

izelnakri:
I just tested this on my machine and works well. This is LGTM, as a first step I think. Eventually we can do the following as the next steps:

- On `index-node.js` finalization, build an CI artifact file so we can easily check/download the ops/sec of each build and see if theres a performance difference.
- Put a threshold and make the benchmark fail if ops/sec is lower than the threshold for each group(probably no need to build directly as CI machine resources are dynamically provisioned 3rd party thus could make this CI task flaky.
- Make `npm run benchmark` runs a benchmark comparison of `QUnit.equiv` versus other deep comparison libraries.

izelnakri (approved):

PR #1708, opened 2022-11-09
Fix memory leaks in due to timeout/timeoutHandler

SergeAstapov:
disclaimer: I'm not sure if `Test.finish` is the best place to put such logic, however it seemed most logic place for me.

The problem description:
In Ember.js we have `ember-qunit` package that assigns reference to the currently running application instance to `this.owner`, where `this` is the reference to `testEnvironment` property in `Test` class.

During test execution (here https://github.com/qunitjs/qunit/blob/2.19.3/src/test.js#L752), we persist references to the `timeoutHandler` in the shared `config` object.

After the test execution, `config.timeoutHandler` keeps reference to the last test that was executed (as it's closure created in https://github.com/qunitjs/qunit/blob/2.19.3/src/test.js#L738).

As a result, after tests execution there is a reference to the application instance persisted in heap, which makes it harder to find memory leaks in the application code.

I'm not sure if there is a need/way to write test for this, so please advice.

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -341,6 +341,13 @@ Test.prototype = {
   finish: function () {
     config.current = this;
 
+    // Release the timeout and timeout callback references
+    // to be garbage collected.
+    if (setTimeout) {
+      clearTimeout(this.timeout);
+      config.timeoutHandler = null;
+    }
+
     // Release the test callback to ensure that anything referenced has been
     // released to be garbage collected.
     this.callback = undefined;


SergeAstapov:
@Krinkle do you mind to re-approve CI? appears `setTimeout`/`clearTimeout` is not available in SpiderMonkey CI job (a reminder that is part of DOM API rather than part of JS)

Krinkle:
@SergeAstapov Thanks for this. I thought this might've been a regression from a recent refactor we did (74190290c2a3e56ecca2f4d533f385870ddc8aa3) where this became a method on the Test object, but looks like it would've leaked before tht as well, s it was previously passed as argument, going back to at least 2016 with 33e242b9479ed3b3e3d2893ba9de680da7887f36.

We do have memory leak tests, but this doesn't show up there today as this doesn't appear to be a case of leaking in the typical sense of accumulating memory, and rather the last value not releasing soon enough for your use case. Or... possibly at all come to think of it. Afaik we don't clear it even by `QUnit.done()`, so it'd hang around until shutdown. In any event, LGTM. Will land this weekend after verifying this myself and documenting for future reference anything I think we might forget about this to make sure it doesn't regress.

Krinkle (approved):

SergeAstapov:
Hi @Krinkle! Do you think this is ready to be merged?

Cjewett:
@Krinkle @SergeAstapov Any movement here? We're running into some rather serious memory leaks on our end and after attempting to create a simple application as a control group I noticed every test run seems to be leaking when using a simple Ember.js application. I believe this change would reduce a lot of noise for us.

Krinkle:
@SergeAstapov Thanks for sticking with this one.

> I'm not sure if there is a need/way to write test for this, so please advice.

Generally, yes, I would say we do want to test this to avoid regressions during internal refactoring. We have an existing test for this already, which covers detecting leaks of the `test.callback` closure. I've amended your patch to include a copy of that test focussed on the `Test` object itself, rather than the test closure.

SergeAstapov:
@Krinkle thank you!

PR #1710, opened 2023-02-10
Prevent modules from leaking testEnvironment to parent modules

chriskrycho:
Today, modules' `testEnvironment` is leaked to any parent module. This makes it impossible for callers to rely on garbage collection of items attached to the module context, which in turn makes it impossible to test against memory leaks from the code under test. Accordingly, this PR makes two changes:

- [x] Add failing tests which assert that the `this` of an inner module is not reference-equal to the test of its parent module.
- [ ] Update module environment handling not to leak, matching the behavior of tests.


chriskrycho:
This was a mistaken understanding of the intent here; sorry for the noise!

Krinkle:
@chriskrycho No problem! Two related thoughts come to mind:

QUnit also supports the, sometimes perceived as more modern, way of sharing values between test cases and between different parent modules, through lexical scope rather than a "this" object. The "nested"  module scope approach (<https://api.qunitjs.com/QUnit/module/>) is very similar in terms inherence, so it's not exactly attractive for being different or better. But, in my experience JS developers tend to be less surprised by how values are shared through lexical scopes, and e.g. less surprised by a moduel that has  `let foo;` in the parent scope, and then `beforeEach()` setting `foo = …`, compared to only a beforeEach that assigns`this.foo = …;` and having to guess whether and how that might is inherited or copied etc. I believe for Ember some properties are set automatically through automation, so that might be a reason to stick with the "this" object. No pressure of any kind, both are perfectly fine, but FYI.

There's a proposal on the roadmap to apply more inheritence to the "this" object during `before()` callbacks QUnit 3. If you have ny thoughts on that, I'd love to hear it, over at https://github.com/qunitjs/qunit/issues/1328.

PR #1722, opened 2023-07-25
Core: add type checking for assert.async

### LINKED ISSUE #1721: Improve assert.async Function to Handle Type Checking
* **QUnit version**: 2.19.4
* **Which environment are you using?**: browser
* **How are you running QUnit? **: Karma

## What are you trying to do?
Currently, using a string instead of a number as an argument for assert.async in QUnit causes the test page to hang indefinitely without providing diagnostic messages. To enhance this function and improve diagnostics, we propose adding type checking.

```js
QUnit.test("Test type checking for async", function(assert) {
    assert.async("the test should fail");
});
```

While running the code above, the test should fail and deliver proper diagnostics message.

## What actually happened?

The test hang indefinitely.

zixinyin:
Added type checking for assert.async, so for an invalid input like string, the page would not hang indefinitely.

Fixes #1721

### DIFF
--- a/src/assert.js
+++ b/src/assert.js
@@ -77,7 +77,7 @@ class Assert {
     return this.test.internalStop(requiredCalls);
   }
 
-  // goExports test.push() to the user API
+  // Exports test.push() to the user API
   // Alias of pushResult.
   push (result, actual, expected, message, negative) {
     Logger.warn('assert.push is deprecated and will be removed in QUnit 3.0.' +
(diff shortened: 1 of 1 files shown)


Krinkle:
> The [CLA page](https://cla.js.foundation/qunitjs/qunit) is not working for me initially so I cannot sign up that. I just signed the EasyCLA above but not sure how to proceed. Please help guide this.

I'm sorry about that. Where did you find the link to the (old) CLA service?

PR #1728, opened 2023-10-14
Better ESM support

NullVoxPopuli:
Resolves: https://github.com/qunitjs/qunit/issues/1724

(however, I can't run tests locally, because the browser launcher is puppeteer, and some ancient browser that I don't have installed :sweat_smile: 
```
>> There was an error with headless chrome
Fatal error: Could not find browser revision 818858. Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.
```
)

I had started with an esm.html file to test with, but `script.type=module` doesn't allow the file protocol when loading modules, so we'd have to use an http server, but then the relative file paths for the link / importmap / script.src wouldn't work the same.

### DIFF
--- a/src/export.js
+++ b/src/export.js
@@ -21,7 +21,7 @@ export default function exportQUnit (QUnit) {
 
     // For consistency with CommonJS environments' exports
     module.exports.QUnit = QUnit;
-    // For allowing a CJS-compatible ESM imports
+    // For allowing ESM imports
     module.exports.todo = QUnit.todo;
     module.exports.skip = QUnit.skip;
     module.exports.test = QUnit.test;
--- a/src/export.js
+++ b/src/export.js
@@ -21,6 +21,11 @@ export default function exportQUnit (QUnit) {
 
     // For consistency with CommonJS environments' exports
     module.exports.QUnit = QUnit;
+    // For allowing a CJS-compatible ESM imports
+    module.exports.todo = QUnit.todo;
+    module.exports.skip = QUnit.skip;
+    module.exports.test = QUnit.test;
+    module.exports.module = QUnit.module;
 
     exportedModule = true;
   }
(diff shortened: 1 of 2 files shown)


src/export.js
```diff
@@ -21,6 +21,11 @@ export default function exportQUnit (QUnit) {
```
Krinkle, line 21: Do you know what it is about `module.exports = QUnit;` that doesn't work for ESM compat layers? It seems odd to assign `module.exports = QUnit;` but then (seemingly redundantly) also assign `module.exports.test = QUnit.test;` to itself. If the transformation layers are based on static analysis, that would explain it, but I thought they were runtime based?

NullVoxPopuli: maybe. as I dug in to this, I think it'd be better to have actual separate builds -- and we're already in a good spot since it seems QUnit is written in ESM -- so we'd just emit a new bundle and wire up `package.json#exports` so that the import/require/etc are routed correctly.

Krinkle:
@NullVoxPopuli Can you confirm that the QUnit 3.0.0-alpha.4 release solves this?

E.g. <https://code.jquery.com/qunit/qunit-3.0.0-alpha.4.module.js> or <https://unpkg.com/qunit@3.0.0-alpha.4/qunit/esm/qunit.module.js>

NullVoxPopuli:
It does?

Krinkle:
> It does?

You can `import { module, test } from 'qunit';` with the new ESM build.

At least in the few ways we test it at https://github.com/qunitjs/qunit/blob/3.0.0-alpha.4/demos/bundlers/test/import-named.js, but if you find cases where it doesn't work, that'd be good to know!

NullVoxPopuli:
You're too fast! I used the wrong punctuation

Krinkle:
Ah... 😅 Thanks for checking!

PR #1732, opened 2023-10-25
Drop AMD

NullVoxPopuli:
Drops AMD emitting, https://github.com/qunitjs/qunit/issues/1729

AMD is not a format that survived the module-format wars (we have only ESM and CJS as authoring formats now, and anyone using AMD complies either ESM or CJS to AMD)


Krinkle:
Can you target this to `:next` instead?

NullVoxPopuli:
oops, done!

PR #1738, opened 2024-02-07
Core: Fix compatibility with sinon.useFakeTimers in IE

timmywil:
The Promise polyfill needed to save the unmodified setImmediate function in IE to use for Promise._immediateFn

### DIFF
--- a/lib/promise-polyfill.js
+++ b/lib/promise-polyfill.js
@@ -373,16 +373,18 @@ Promise.race = function(arr) {
 };
 
 // Use polyfill for setImmediate for performance gains
-Promise._immediateFn =
+// @ts-ignore
+if (typeof setImmediate === 'function') {
   // @ts-ignore
-  (typeof setImmediate === 'function' &&
-    function(fn) {
-      // @ts-ignore
-      setImmediate(fn);
-    }) ||
-  function(fn) {
+  var setImmediateFunc = setImmediate;
+  Promise._immediateFn = function(fn) {
+    setImmediateFunc(fn);
+  };
+} else {
+  Promise._immediateFn = function(fn) {
     setTimeoutFunc(fn, 0);
   };
+}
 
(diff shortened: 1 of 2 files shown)


test/main/promise.js
```diff
@@ -247,4 +265,25 @@ QUnit.module('Support for Promise', function () {
...
+      this.setTimeout = globalNS.setTimeout;
+      globalNS.setTimeout = function () {};
+      this.setImmediate = globalNS.setImmediate;
+      globalNS.setImmediate = function () {};
+      // Adds 1 assertion
+      return createMockPromise(assert);
+    },
+    afterEach: function (assert) {
+      globalNS.setTimeout = this.setTimeout;
+      globalNS.setImmediate = this.setImmediate;
```
Krinkle, line 280: This leaves behind an undefined global property in environments where the property did not exist. Can we avoid that?

lib/promise-polyfill.js
```diff
@@ -373,16 +373,18 @@ Promise.race = function(arr) {
 };
 
 // Use polyfill for setImmediate for performance gains
-Promise._immediateFn =
+// @ts-ignore
```
Krinkle, line 376: Please update the `Patches for use in QUnit:` header to document this local patch, to make sure we remember it in the future :)

Krinkle (changes requested):
Two inline issues commented.

Krinkle (approved):

Krinkle:
Looks like GitHub got itself in some kind of broken state, not agreeing on which commit's status to look at for the merge button, thus blocking me from merging due to waiting for a status that would never come. Closing and re-opening fixed this.

timmywil:
Thanks! If you could do a release, I can update the jQuery branch.

Dearley78 (commented):
lib/promise-polyfill.js

PR #1740, opened 2024-02-09
Core: Refactor internal ProcessingQueue into class

Krinkle:
This is motivated by the last remaining build warning from Rollup:

> src/qunit.js → qunit/qunit.js...
> (!) Circular dependency
> src/test.js -> src/core/processing-queue.js -> src/test.js
> created qunit/qunit.js in 2s

ProcessingQueue needs access to the `test` function defined in test.js.

Fix by turning the module into a class, that we create a singleton of in core.js (with access to test.js) and then use that singleton where we previously used ProcessingQueue statically.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -31,6 +31,8 @@ const QUnit = {};
 // rather than partly in config.js and partly here.
 config.currentModule.suiteReport = runSuite;
 
+config.pq = new ProcessingQueue(test);
+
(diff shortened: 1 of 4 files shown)


Krinkle:
@izelnakri FYI: You'll want to update qunitx to where it resets the internal `ProcessingQueue.finished` to reset `config.pq.finished` instead!

PR #1758, opened 2024-05-29
Build: Remove publication to Bower for future releases

Krinkle:
Ref https://github.com/qunitjs/qunit/issues/1677.


mgol (approved):
LGTM. I wonder if we'll ever do the same for jQuery.

PR #1760, opened 2024-06-06
HTML Reporter: Update testresult display in QUnit 3.0

Krinkle:
QUnit 2.21.0:
> 254 tests completed in 912 milliseconds, with 0 failed, 7 skipped, and 4 todo.
> 819 assertions of 823 passed, 4 failed.

New:
> 254 tests completed in 0.9 seconds, with 0 failed, 7 skipped, and 4 todo.

For details, refer to the two sepaerate commit messages.

### DIFF
--- a/src/html-reporter/html.js
+++ b/src/html-reporter/html.js
@@ -721,6 +721,15 @@ const stats = {
     ].join('');
   }
 
+  function msToSec (milliseconds) {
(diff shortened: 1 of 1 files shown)


mgol (approved):
LGTM, especially that we don't provide this detailed info during the run but only at the end and it sounds a bit abstract.

PR #1761, opened 2024-06-08
Core: Error when a module callback has a promise as a return value

raycohen:
Ref https://github.com/qunitjs/qunit/issues/1600.

For the QUnit version 3.0 release.

I'd have moved the tests for this into `test/main/modules.js` but I am not able to use async/await syntax nor promises in that file.

Recreation of #1616


Krinkle:
@raycohen I suggest adding it there with a thenable as return instead.

Like so:

```js
QUnit.module('example', function () {
  QUnit.test('test', function (assert) {
    assert.true(true);
  });

  return { then: function () {} };
});
```

By the way, upon first reading the test, I was hoping you'd make it a failing test, e.g. `assert.true(false)` on the expectation that the test doesn't actually run since we throw right away. However... despite the seemingly "fast" failure through this uncaught error, the tests would still get defined and run regardless (with any async ones going to an unrelated module in all likelihood). This seems unavoidable without more changes, perhaps best outside the scope of this PR. But it did make me think if there's a way we could detect an AsyncFunction without calling it. It looks like there is?

* https://262.ecma-international.org/12.0/#sec-runtime-semantics-namedevaluation
* https://262.ecma-international.org/12.0/#sec-runtime-semantics-instantiateasyncfunctionexpression `AsyncFunction.prototype`

I haven't yet looked at what, if any, established idiom there is for this, but, if nothing else, one could probably use `Object.prototype.toString(fn) === "[object AsyncFunction]"` and skip the call in those cases to avoid starting an untracked asynchronous call stack. What do you think about that? The downside would be that we'd only do that for true/native async functions and not those that are compiled to (or written as) regular functions that merely return promises. Given this is error handling, maybe the inconsistency isn't too bad, for the benefit of less noise/other issues being produced.

test/cli/fixtures/async-module-warning-promise.js
```diff
@@ -1,6 +1,24 @@
-QUnit.module('module manually returning a promise', function () {
-  QUnit.test('has a test', function (assert) {
-    assert.true(true);
+var errorFromModule;
+
+try {
```
Krinkle, line 3: This updated version of the fixture is great for test/main/modules.js (which is where you wanted it originally). If we can make that one work with a then-able function, then I suggest restoring this like it was before as example of how it would fail "for real", and taking adventage of how TAP has a built-in ability to assert failures in-place. This has the benefit of both show-casing and verifying how it fails for real without trying to catch it at the same time.

raycohen:
I added an additional `{ then: function () {} }` fixture, renamed the other two fixtures to "error" instead of "warning", and added the early bail when the async function can be detected.

I'm not sure what's happening with the linter on the Linux ci builds saying Logger is unused.

Krinkle:
> I'm not sure what's happening with the linter on the Linux ci builds saying Logger is unused.

You may have to rebase this locally first. The main branch has seen a few commits since, leaving the call to `Logger.warn()` that your patch removes, as the only one remaining. Thus it is unused from the state of the repo that CI sees (which is, simulate a merge of the PR into latest main, then run the CI jobs).

test/main/modules.js
```diff
@@ -454,4 +454,38 @@ QUnit.module('QUnit.module', function () {
...
+    QUnit.test('module with thenable callback function errored', function (assert) {
+      assert.true(errorFromThenableCallbackModule instanceof Error);
+      assert.strictEqual(errorFromThenableCallbackModule.message, 'Returning a promise from a module callback is not supported. Instead, use hooks for async behavior.');
+    });
+
+    var errorFromPromiseCallbackModule;
+    try {
+      QUnit.module('with promise callback', function () {
+        // eslint-disable-next-line compat/compat
+        return Promise.resolve();
```
raycohen, line 477: `Promise` isn't available with IE11 which triggered this lint. Also with the es5 setup I can't use `async` functions in this file.

Krinkle: We can use the then-able function return here I think? The added fixture is fine for extra coverage, but I meant it originally for here.

raycohen: 👍 I'll remove this one and keep the thenable version

raycohen:
>  Given this is error handling, maybe the inconsistency isn't too bad, for the benefit of less noise/other issues being produced.

I agree the Async function deserves special treatment since we can detect it. I'm guessing it's the most common one the most users will try doing as well,

PR #1762, opened 2024-06-09
Core: Run before and after hooks with module context

### LINKED ISSUE #1328: Share 'before' hook test environment with child module

raycohen:
The before and after hooks run once per module as long as there is at least one test in the module. Using environment inheritance allows us to use the module context in those hooks, which allows reading the expected changes to the context from a before hook inside nested modules.

This is a breaking change from 2.x behavior. Previously the `after` hook ran with the last test's context instead of the module context, which allowed it to access things set on `this` from inside the test. That will no longer be accessible, and some of the tests were relying on that behavior, so I had to refactor them to get that information from variables in scope for both the test and after hook instead.

This PR replaces #1559

Fixes #1328.
Ref #869.

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -21,9 +21,9 @@ function createModule (name, testEnvironment, modifiers) {
   const skip = (parentModule !== null && parentModule.skip) || modifiers.skip;
   const todo = (parentModule !== null && parentModule.todo) || modifiers.todo;
 
-  const env = {};
+  let env = {};
   if (parentModule) {
-    extend(env, parentModule.testEnvironment);
+    env = Object.create(parentModule.testEnvironment || {});
   }
   extend(env, testEnvironment);
 
--- a/src/test.js
+++ b/src/test.js
@@ -158,8 +158,6 @@ Test.prototype = {
     return moduleStartChain.then(() => {
       config.current = this;
 
-      this.testEnvironment = extend({}, module.testEnvironment);
-
       this.started = performance.now();
       emit('testStart', this.testReport.start(true));
       return runLoggingCallbacks('testStart', {
@@ -251,17 +249,19 @@ Test.prototype = {
 
   queueHook (hook, hookName, hookOwner) {
     const callHook = () => {
-      const promise = hook.call(this.testEnvironment, this.assert);
+      let promise;
+      if (hookName === 'before' || hookName === 'after') {
+        // before and after hooks are called with the owning module's testEnvironment
+        promise = hook.call(hookOwner.testEnvironment, this.assert);
+      } else {
+        promise = hook.call(this.testEnvironment, this.assert);
+      }
       this.resolvePromise(promise, hookName);
     };
 
     const runHook = () => {
-      if (hookName === 'before') {
-        if (hookOwner.testsRun !== 0) {
-          return;
-        }
-
-        this.preserveEnvironment = true;
+      if (hookName === 'before' && hookOwner.testsRun !== 0) {
+        return;
       }
 
       // The 'after' hook should only execute when there are not tests left and
@@ -483,13 +483,6 @@ Test.prototype = {
     }
   },
 
-  preserveTestEnvironment: function () {
-    if (this.preserveEnvironment) {
-      this.module.testEnvironment = this.testEnvironment;
-      this.testEnvironment = extend({}, this.module.testEnvironment);
-    }
-  },
-
   queue () {
     const test = this;
 
@@ -507,7 +500,7 @@ Test.prototype = {
         ...test.hooks('before'),
 
         function () {
-          test.preserveTestEnvironment();
+          test.testEnvironment = extend({}, test.module.testEnvironment, false, true);
         },
 
         ...test.hooks('beforeEach'),
--- a/test/main/modules.js
+++ b/test/main/modules.js
@@ -16,42 +16,36 @@ QUnit.module('QUnit.module', function () {
       // parent > child > one
       // parent > child > two
       'parent-before: (empty)',
-      'child-before: beforeP=1',
-      'parent-beforeEach: beforeP=1 beforeC=1',
-      'child-beforeEach: beforeP=1 beforeC=1 beforeEach=P',
-      'child-test: beforeP=1 beforeC=1 beforeEach=PC',
-      'child-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=1',
-      'parent-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=1 afterEach=C',
-      'parent-beforeEach: beforeP=1 beforeC=1',
-      'child-beforeEach: beforeP=1 beforeC=1 beforeEach=P',
-      'child-test: beforeP=1 beforeC=1 beforeEach=PC',
-      'child-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=2',
-      'parent-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=2 afterEach=C',
-      'child-after: beforeP=1 beforeC=1 beforeEach=PC tester=2 afterEach=CP',
-      'parent-after: beforeP=1 beforeC=1 beforeEach=PC tester=2 afterEach=CP afterC=1'
+      'child-before: %beforeP=1',
+      'parent-beforeEach: beforeC=1 beforeP=1',
+      'child-beforeEach: beforeC=1 beforeP=1 beforeEach=P',
+      'child-test: beforeC=1 beforeP=1 beforeEach=PC',
+      'child-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=1',
+      'parent-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=1 afterEach=C',
+      'parent-beforeEach: beforeC=1 beforeP=1',
+      'child-beforeEach: beforeC=1 beforeP=1 beforeEach=P',
+      'child-test: beforeC=1 beforeP=1 beforeEach=PC',
+      'child-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=2',
+      'parent-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=2 afterEach=C',
+      'child-after: beforeC=1 %beforeP=1',
+      'parent-after: beforeP=1'
     ],
-    // FIXME: https://github.com/qunitjs/qunit/issues/1328
-    // - parent test missing own state if there is a child module before the test.
-    // - last test state presists into after()
     'parent with trailing test': [
       // parent > child > one
       // parent > two
       'parent-before: (empty)',
-      'child-before: beforeP=1',
-      'parent-beforeEach: beforeP=1 beforeC=1',
-      'child-beforeEach: beforeP=1 beforeC=1 beforeEach=P',
-      'child-test: beforeP=1 beforeC=1 beforeEach=PC',
-      'child-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=1',
-      'parent-afterEach: beforeP=1 beforeC=1 beforeEach=PC tester=1 afterEach=C',
-      'child-after: beforeP=1 beforeC=1 beforeEach=PC tester=1 afterEach=CP',
-      'parent-beforeEach: (empty)',
-      'parent-test: beforeEach=P',
-      'parent-afterEach: beforeEach=P tester=2',
-      'parent-after: beforeEach=P tester=2 afterEach=P'
+      'child-before: %beforeP=1',
+      'parent-beforeEach: beforeC=1 beforeP=1',
+      'child-beforeEach: beforeC=1 beforeP=1 beforeEach=P',
+      'child-test: beforeC=1 beforeP=1 beforeEach=PC',
+      'child-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=1',
+      'parent-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=1 afterEach=C',
+      'child-after: beforeC=1 %beforeP=1',
+      'parent-beforeEach: beforeP=1',
+      'parent-test: beforeP=1 beforeEach=P',
+      'parent-afterEach: beforeP=1 beforeEach=P tester=2',
+      'parent-after: beforeP=1'
     ],
-
-    // FIXME: https://github.com/qunitjs/qunit/issues/1328
-    // child is missing parent state if there is an initial test before the child module.
     'parent with initial test': [
       // parent > one
       // parent > child > two
@@ -59,14 +53,14 @@ QUnit.module('QUnit.module', function () {
       'parent-beforeEach: beforeP=1',
       'parent-test: beforeP=1 beforeEach=P',
       'parent-afterEach: beforeP=1 beforeEach=P tester=1',
-      'child-before: (empty)',
-      'parent-beforeEach: beforeC=1',
-      'child-beforeEach: beforeC=1 beforeEach=P',
-      'child-test: beforeC=1 beforeEach=PC',
-      'child-afterEach: beforeC=1 beforeEach=PC tester=2',
-      'parent-afterEach: beforeC=1 beforeEach=PC tester=2 afterEach=C',
-      'child-after: beforeC=1 beforeEach=PC tester=2 afterEach=CP',
-      'parent-after: beforeC=1 beforeEach=PC tester=2 afterEach=CP afterC=1'
+      'child-before: %beforeP=1',
+      'parent-beforeEach: beforeC=1 beforeP=1',
+      'child-beforeEach: beforeC=1 beforeP=1 beforeEach=P',
+      'child-test: beforeC=1 beforeP=1 beforeEach=PC',
+      'child-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=2',
+      'parent-afterEach: beforeC=1 beforeP=1 beforeEach=PC tester=2 afterEach=C',
+      'child-after: beforeC=1 %beforeP=1',
+      'parent-after: beforeP=1'
     ],
 
     // Confirm each step waits for the previous before restoring/saving testEnvironment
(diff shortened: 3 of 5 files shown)


test/main/modules.js
```diff
@@ -309,8 +311,14 @@ QUnit.module('QUnit.module', function () {
...
       this.inner = true;
 
+      hooks.before(function () {
+        this.innerBefore = 'inner before value';
+      });
+
       hooks.beforeEach(function (assert) {
         assert.strictEqual(this.outer, 2);
+        assert.strictEqual(this.outerBefore, 'outer before value');
+        assert.strictEqual(this.innerBefore, 'inner before value');
```
raycohen, line 321: should we assert all of these things in `hooks.afterEach` as well?

src/module.js
```diff
@@ -22,9 +22,9 @@ function createModule (name, testEnvironment, modifiers) {
   const skip = (parentModule !== null && parentModule.skip) || modifiers.skip;
   const todo = (parentModule !== null && parentModule.todo) || modifiers.todo;
 
-  const env = {};
+  let env = {};
   if (parentModule) {
-    extend(env, parentModule.testEnvironment);
+    env = Object.create(parentModule.testEnvironment || {});
```
raycohen, line 27: We can't use `??` here because of IE support, right?

Krinkle: Newer syntax is generally fine. We use `...` rest/spread and ES6 classes for example as well. These are transpiled by Babel.

Where we're limited is calling to APIs such as new Array methods, new globals like Promise or Map, things like that.

src/test.js
```diff
@@ -158,7 +158,7 @@ Test.prototype = {
     return moduleStartChain.then(() => {
       config.current = this;
 
-      this.testEnvironment = extend({}, module.testEnvironment);
+      this.testEnvironment = Object.create(module.testEnvironment || {});
```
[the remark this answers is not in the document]
raycohen: This `test.before` function runs prior to the module `before` hook callback running. So with the `extend` approach, the environment is copied over prior to the `before` hook getting its chance to modify it.

src/test.js
```diff
@@ -251,17 +251,18 @@ Test.prototype = {
 
   queueHook (hook, hookName, hookOwner) {
     const callHook = () => {
-      const promise = hook.call(this.testEnvironment, this.assert);
+      let promise;
+      if (hookName === 'before' || hookName === 'after') {
+        promise = hook.call(this.module.testEnvironment, this.assert);
```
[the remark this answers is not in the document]
raycohen: the `Object.create` change gives the test's `testEnvironment` a long-lasting relationship with the module's `testEnvironment`. It's not a one-time copy anymore.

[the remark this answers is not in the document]
Krinkle: I see. This reminds me of the discussion at https://github.com/qunitjs/qunit/issues/869. Would the intended change still work with your updated tests passing, if we kept using `extend()` in this patch and instead made it so that `testEnvironment` gets re-created between `before()` / `test.hooks('before')` and `test.hooks('beforeEach')` / `test.run()`?

This is what preserveTestEnvironment does today basically.

We could look at re-ordering this more neatly, but moving the timing of `testStart` or leading `test.testEnvironment` undefined during that event seems behaviour-changing in other ways that I'm not sure we definitely want. Preserving the changes during `before()` onto the module, and creating the per-test enviornment after that, might be a good compromise. What do you think?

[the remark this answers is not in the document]
raycohen: I think I understand the path you're suggesting, but I'm not totally clear on what in practice about this path you're concerned about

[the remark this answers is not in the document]
Krinkle: The live inheritence via `Object.create` means that, we not only allow the one-time and known use case of a "before" hook modifying the environment between `test.before()` and `test.run()`, but also allow mutation at any later time by reference with hard-to-track retroactive changes that result in not all tests getting the same starting environment depending on when they run.

I'd prefer to guruantee that once a module begins, each test gets a fresh copy of the same starting environment.

While this is just one example to exploit it, I feel there are more unknown as well. The use cases we have is under our control and we can explicitly do it once for the point where we need to do it, rather than establish live inheritence. I'm open to re-considering this in QUnit 3.1, making sure that it will not prevent anyone from moving to QUnit 3.0, and giving us more time to think about and test it.

test/main/modules.js
Krinkle: Sure!

src/test.js
```diff
@@ -507,7 +501,7 @@ Test.prototype = {
         ...test.hooks('before'),
 
         function () {
-          test.preserveTestEnvironment();
+          test.testEnvironment = extend({}, test.module.testEnvironment, false, true);
```
raycohen, line 504: Maybe it's not worth using `extend` here as I don't want only own properties from the module's testEnvironment. I added a new boolean arg to switch it to all properties.

src/test.js
```diff
@@ -158,8 +158,6 @@ Test.prototype = {
     return moduleStartChain.then(() => {
       config.current = this;
 
-      this.testEnvironment = extend({}, module.testEnvironment);
```
raycohen, line 161: with the current approach nothing should access this until the `before` hooks have completed

src/test.js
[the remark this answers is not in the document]
raycohen: I've updated to only set `test.testEnvironment` to a one-time clone of the `module.testEnvironment` once the `before` hooks have completed

src/test.js
Krinkle: OK. In that case we'll need to keep in mind to note in the 3.0 Upgrade Guide that `QUnit.moduleStart()` callbacks can no longer access testEnvironment via [`QUnit.config.current`](https://qunitjs.com/api/config/current/), and recommend `QUnit.hooks.before` for that use case instead.

I'm not aware of this being common, so that's fine I think and could be worth doing. Perhaps more likely, is access to `QUnit.config.current` from `QUnit.testStart()`. It's ambigious whether that intends to modify the base context for all tests in the current module during the first test. Most likely not intended. But also most likely harmless since it'll re-run for each test. Making that no longer modify the base context could be seen as a bug fix and either way, a change I think is good to make.

But... I think the current version of the PR does more than that, it makes it undefined. That too could be mitigated by recommending use of `QUnit.hooks.beforeEach` instead in QUnit 3.0, but I'm less sure whether that's worth breaking. What do you think?

raycohen: so if we set `this.testEnvironment` to the module context here, conditional modifications to `QUnit.config.current.testEnvironment` would modify the module context and could "leak" awkwardly and impact following tests in unintended ways.

we could set `this.testEnvironment` to `extend({}, module.testEnvironment, false, true)` and clobber it when it gets set again later, which would allow reading it during `testStart` but not allow modifying it.

keeping it this way means it is `undefined` during `testStart` and can't be read or modified.

another option could be to make a sort of implied single-test module context here. so we'd do
```
this.testEnvironment = Object.create(module.testEnvironment || {});
```
here and use the resulting `this.testEnvironment` for the `extend` call in the test queue later on, so the test hooks still receive a flattened copy. Which I think would preserve modifications made during `testStart` _and_ only impact the current test? 

If the last options sounds worth doing then maybe we should add test coverage on that behavior and I can rebase with that strategy to make sure it still works how I'm imagining.

Krinkle:
_(continuing in main-thread, responding to [comment](https://github.com/qunitjs/qunit/pull/1762#discussion_r1646510053) about `this.testEnvironment = extend({}, module.testEnvironment);`)_

@raycohen Yeah, read-only could be a compromise here.

So what we have today:
* test.before() fires moduleStart + testStart event callbacks, which feel "outside" the test execution. Not given `assert` or `this`. But, can technically access it, albeit with the now-known strange and surprising behaviours in various cases, does broadly work fine in practice, but definitely not good enough, not up to our standard.
* we perform execution: before hooks, test, after hooks.
* test.finish() fires moduleDone + testEnd event callbacks.

I see a couple of options:

1. Keep clean separation of events and execution. Remove remnant `this.testEnvironment`.
  * `QUnit.config.current.testEnvironment` becomes undefined during moduleStart/testStart.
  * We'll later document as footnote in QUnit 3 guide to use `QUnit.hooks.before/beforeEach()` for this purpose instead.
3. Keep separation. Remove mutation and keep limited read-only compat.
  * give events a fresh throw-away copy of the module env (start) or last test (end), similar to what we do around each test.
  * Between moduleStart+testStart and hooks.before, we will create a new copy of the module env, this time mutable.
  * Then, between hooks.before and hooks.beforeEach/test.run/etc we create another throw-away copy, like today.
4. Interweave events and execution. This would offer clear and clean ordering with regards creating the module env, then copying to run the test, and restoring afterward. No lost writes. No time-travelling.
  * Today:
    * event moduleStart
    * **event testStart**
    * **hook before**
    * hook beforeEach
    * test run
    * hook afterEach
    * **hook after**
    * **event testEnd**
    * event moduleDone
  * Instead:
    * event moduleStart (testEnv=module env)
    * 🔺 **hook before** (testEnv=module env)
    * 🔻 **event testStart** (testEnv=test env)
    * hook beforeEach
    * test run
    * hook afterEach
    * 🔺 **event testEnd** (testEnv=test env)
    * 🔺 **hook after** (testEnv=module env)
    * event moduleDone

Summary:
1. **Pro:** Unchanged order. Intuitive outcome (no surprises/dead-ends). Change is easy to explain. Use hooks for env. **Con:** Breaks env in events.
2. **Pro:** Unchanged order. Unchanged support for read-only env in events. **Cons:** Break env writes in events. Change subtle/difficult to explain (only read-only is permitted). Writes are silently lost (surprising). Content of the env object either continues to be under-specified in edge cases and/or will undergo the same major change as for begin/after hooks. Continue time-travelling between "test" start and "module" hooks.
3. **Pro:** Unchanged support for event in events (read-write). Intuitive outcome for env lifetime and ordering (no lost writes, no time-travelling.). **Neutral:** Different execution order, arguably clearer in some ways. **Cons:** Breaks execution order. Encourages mixing of observing and runtime. Leaves odd entrypoint for writing to module env from `current.testEnv`, which if we improve that, we might as well go with (1) and recommend use of global hooks.

test/main/modules.js
```diff
@@ -41,30 +40,28 @@ QUnit.module('QUnit.module', function () {
       'child-test: before=PC beforeEach=PC',
       'child-afterEach: before=PC beforeEach=PC tester=1',
       'parent-afterEach: before=PC beforeEach=PC tester=1 afterEach=C',
-      'child-after: before=PC beforeEach=PC tester=1 afterEach=CP',
+      'child-after: before=PC',
       'parent-beforeEach: (empty)',
       'parent-test: beforeEach=P',
       'parent-afterEach: beforeEach=P tester=2',
-      'parent-after: beforeEach=P tester=2 afterEach=P'
+      'parent-after: (empty)'
```
raycohen, line 47: I'm looking into why this isn't `before=P`

src/test.js
```diff
@@ -251,17 +249,18 @@ Test.prototype = {
 
   queueHook (hook, hookName, hookOwner) {
     const callHook = () => {
-      const promise = hook.call(this.testEnvironment, this.assert);
+      let promise;
+      if (hookName === 'before' || hookName === 'after') {
+        promise = hook.call(this.module.testEnvironment, this.assert);
```
raycohen, line 254: there's still an issue here which is that the `before` and `after` hooks can be called with the wrong module's testEnvironment depending on the ordering of the tests. Instead of calling the hook with the _current test_'s module, it should be called with the module the hook belongs to.

test/main/modules.js
```diff
@@ -25,12 +25,9 @@ QUnit.module('QUnit.module', function () {
       'child-test: before=PC beforeEach=PC',
       'child-afterEach: before=PC beforeEach=PC tester=2',
       'parent-afterEach: before=PC beforeEach=PC tester=2 afterEach=C',
-      'child-after: before=PC beforeEach=PC tester=2 afterEach=CP',
-      'parent-after: before=PC beforeEach=PC tester=2 afterEach=CP after=C'
+      'child-after: before=PC',
+      'parent-after: before=P'
```
raycohen, line 29: the parent after hook can't see that `this.after` was set to `C` in the child's after hook, because the assignment becomes one of the child `module.testEnvironment`'s _own_ properties.

This isn't a problem with `before` because the parent's before hook always runs first, and the child _can_ see it, since it inherits from the parent environment.

I think this update to the test is technically correct, but it may highlight that there's additional things we want to assert

src/test.js
raycohen: In https://github.com/qunitjs/qunit/pull/1762/commits/759a391c40fbe8bdab391ba50c59ce76a5f9b912 I update to using `hookOwner` instead of `this.module` which resolves this problem.

raycohen:
>Neutral: Different execution order, arguably clearer in some ways.
I would say my initial impression is that it seems odd to include the before and after hooks inside the testStart/testEnd events, so the changed order you have seems clearer or less surprising to me.

It does seem like open 1 is the simplest to explain. Without being aware of any use cases for reading and writing there, I might vote for that. It seems like there's a refactor available that likely gets anyone using it for that a way forward?

src/test.js
Krinkle: This essentially offsets the introduction of `Object.create()` between module and parent module, which previously used extend but with inheritence that means you miss out on props. Is that right?

raycohen: Yes, that's right

Krinkle:
@raycohen Yeah, same here. By removing env, we remove importance of the other, and when all else is equal I'd rather not change the ordering there to avoid suble changes in external state in people's code. Sounds like option 1 is a good way forward then!

Krinkle:
I'm slowly comin around around to embracing `Object.create()`  inheritence.

I was pondering if its worth it and possible to avoid this between modules (in the same way we avoid it between module and test. In QUnit 2.x, all context props are owned by the `this` object, making them visible to `Object.keys(this)` and filtered for-loops, with consistent behaviour across module hooks, test hooks, and tests.

This patch I believe preserves that for tests, per-test hooks, and non-nested module hooks. But, for nested child modules, the "before" hook would now see a `this` that has prototypal inheritence to the parent context. I was focussing on how we set up the module hook's env, but completely forgot that we first create `env` during the `module()` function, which happens during the registration phase, not during execution. This starts out with props from the `options` parameter. Copying the parent props to this synchronously, like we do today, is exactly what causes the bug we're trying to solve. It's doing it too early. Inheritence naturally solves that by allowing continous later modification from underneath it.

I briefly considered if we could "park" these options and defer env creation until the test execution phase begins (e.g. at the "before" hook). However that would mean `this` in the module scope callback lacks parent properties set by the same mechanism, which is a breaking change. 

```js
QUnit.module('parent', { foo: 1 }, function () {
	this.foo++;
	QUnit.module('child', { bar: 1 }, function () {
		// sees this.foo=2 and this.bar=1
		var foo = this.foo;
		var bar = this.bar;
			assert.equal(foo, 2, 'foo');
			assert.equal(bar, 1, 'bar');
		});
	});
});
```

I'll cover this with a test, but it shows there's a need to create the env object earlier. The only way to create it correctly and use that same object throughout, is with inheritence. We can't realistically defer env creation, because the callback needs something usable right away. If we really wanted to avoid this behaviour change, we could go deeper and flatten it like today, and _also_ park it for later, and re-create it. But, I think that trades one set of surprises for another.

Given we're doing a semver-major, I'd say let's do this more cleanly as you have it, and document that anything hasOwn-filtering for loops can safely stop doing so. We already guruantee env to be a plain object by copying props. The downside is that convenience functions like `Object.keys()` can't be used. By comparison, `jQuery.each()` and `jQuery.map()` don't filter non-own properties.

I guess the remaining question then is: What level of consistency do we want to go for?

1. Current patch: Both parent and child props are owned, except during `child before hook` and `child after hook`. The parent props are flattened and owned in `beforeEach`, `test`, and `afterEach`.

  Note that if the property contains a primitive value and is "modified" by child before hook, then that read-write operation (e.g. +=) effectively copies it into the own object, thus unless our assertion checks for separate properties, this is hard to observe. I've fixed this in https://github.com/qunitjs/qunit/pull/1770.

2. Alternative: Always use inheritence for parent module. Each test env is created by using Object.create() from the parent module env, and then extend() to mixin the module's own env. This giving the appearance that within the current module, we always operate on own properties both. Both in before/after hooks at the module level, in test hooks, and in tests. This would mean that `Object.keys()` will consistently not see properties inherited from parent modules, in before/after hooks (like option 1), and beforeEach/test/afterEach alike.

3. Alternative: Use inheritence for everything, including between current module and current test. Ech test env is created by using Object.create() from the current module env. This would mean that properties created for the current module's tests by before hooks would become inherited properties during test hooks and tests.

4. Alternative: We treat the observable inheritence as a bug. Before the execution begins (at the "before" hook) we flatten the module env and disconnect from live inheritence. In this case, all code consistently sees own properties only, with the exception of the module scope callback. This has the other difference of making late writes from any badly written async parenet module hooks fail / ignored, same as today.

@raycohen What do you think?

Option 3 feels the purest, but also the largest change. Could a use case benefit from this? If not, I lean towards Option 2 (easy to explain: each module inherits from its parent, each test clones the current module env; the module env can be seen as a template for the test env), or Option 4 (keep inheritence internal like it is today).

Krinkle (approved):

raycohen:
I think I prefer 1 over 4 because having matching behavior for `Object.keys(this)` in the module callback and in the before hook seems desirable.

For 3 I like the implementation but I can't think of any usage-based argument for it. Why would user code care _which_ parent module had set up a given env value? If we go with flattening, a user could manually access that info if they needed, e.g

```
module('parent', function () {
  this.foo = 1;

  module('child', { parentEnv: this }, function () {
    this.hasOwnProperty('foo') // false
    this.parentEnv.hasOwnProperty('foo') // true

    module('deeper child', { parentEnv: this }, function () {
      this.parentEnv.parentEnv.hasOwnProperty('foo') // true
    });
  });
});
```

We could even (in the future, if we realized a compelling use for it) automatically set up a `parentEnv` property as a part of each module's own env. So I don't think we lose much by not doing inheritance at the test level as well.

Alternative 2 is interesting but if the goal is behavior the user expects I think Options 1 and 4 are probably better.

So, I'm also leaning towards Option 1

PR #1768, opened 2024-06-19
Test: Refactor module hook/context state testing

Krinkle:
This should make it significantly more robust and easier to see what's actually going on, primarily to help review https://github.com/qunitjs/qunit/pull/1762.

### DIFF
--- a/test/main/modules.js
+++ b/test/main/modules.js
@@ -1,178 +1,483 @@
 /* global setTimeout */
 QUnit.module('QUnit.module', function () {
-  QUnit.module('before/beforeEach/afterEach/after', {
-    before: function () {
-      this.lastHook = 'module-before';
-    },
-    beforeEach: function (assert) {
-      assert.strictEqual(this.lastHook, 'module-before',
-        "Module's beforeEach runs after before");
-      this.lastHook = 'module-beforeEach';
-    },
-    afterEach: function (assert) {
-      assert.strictEqual(this.lastHook, 'test-block',
-        "Module's afterEach runs after current test block");
-      this.lastHook = 'module-afterEach';
-    },
-    after: function (assert) {
-      assert.strictEqual(this.lastHook, 'module-afterEach',
-        "Module's afterEach runs before after");
-      this.lastHook = 'module-after';
+  function flat (obj) {
+    var str = [];
+    for (var key in obj) {
+      str.push(key + '=' + obj[key]);
     }
-  });
+    return str.join(' ') || '(empty)';
+  }
 
-  QUnit.test('hooks order', function (assert) {
-    assert.expect(4);
+  var actual = {};
+  var expected = {
+    'parent with children': [
+      // parent > child > one
+      // parent > child > two
+      'parent-before: (empty)',
+      'child-before: before=P',
+      'parent-beforeEach: before=PC',
+      'child-beforeEach: before=PC beforeEach=P',
+      'child-test: before=PC beforeEach=PC',
+      'child-afterEach: before=PC beforeEach=PC tester=1',
+      'parent-afterEach: before=PC beforeEach=PC tester=1 afterEach=C',
+      'parent-beforeEach: before=PC',
+      'child-beforeEach: before=PC beforeEach=P',
+      'child-test: before=PC beforeEach=PC',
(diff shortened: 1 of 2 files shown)


test/main/modules.js
```diff
@@ -1,178 +1,386 @@
...
+    for (var key in obj) {
+      str.push(key + '=' + obj[key]);
     }
-  });
+    return str.join(' ') || '(empty)';
+  }
+
+  var actual = {};
+  var expected = {
+    'parent with children': [
```
raycohen, line 13: this makes sense to me 👍

test/main/modules.js
```diff
@@ -1,178 +1,386 @@
...
+      'child-test: before=PC beforeEach=PC',
+      'child-afterEach: before=PC beforeEach=PC tester=1',
+      'parent-afterEach: before=PC beforeEach=PC tester=1 afterEach=C',
+      'parent-beforeEach: before=PC',
+      'child-beforeEach: before=PC beforeEach=P',
+      'child-test: before=PC beforeEach=PC',
+      'child-afterEach: before=PC beforeEach=PC tester=2',
+      'parent-afterEach: before=PC beforeEach=PC tester=2 afterEach=C',
+      'child-after: before=PC beforeEach=PC tester=2 afterEach=CP',
+      'parent-after: before=PC beforeEach=PC tester=2 afterEach=CP after=C'
```
raycohen, line 29: with the changes in #1762 the `after` hooks will lose access to everything set in hooks other than `before` and `after` so I expect these will become
```
'child-after: before=PC',
'parent-after: before=P'
```

Perhaps it would be worth showing that multiple `after` hooks on the same level can see the previous env changes? ie

```
'child-after1: before=PC',
'child-after2: before=PC after=CA1',
'parent-after1: before=P',
'parent-after2: before=P after=PA1'
```

raycohen (approved):

test/main/modules.js
```diff
@@ -282,109 +490,6 @@ QUnit.module('QUnit.module', function () {
...
-      });
-
-      QUnit.test('should run outer-after and after', function (assert) {
-        assert.expect(2);
-        this.lastHook = 'outer-test';
-      });
-    });
-  });
-
   QUnit.module('multiple hooks', function (hooks) {
```
Krinkle, line 493: FYI: This covers multiple hooks order, but not context indeed.

test/main/modules.js
Krinkle: Good idea! I'll add that.

Krinkle: Yeah, "afterEach" should tear down state from beforeEach or tests. "after" would only see state that started at "begin". I do note that this differs from how e.g. PHPUnit works where a single `$this` by-reference persists throughout a single module (TestClase class).

But, I think it makes sense for QUnit because we already don't expose treat the state of other test cases with "after" callback either. All tests get their env garbage-collected of without after hooks, only the last one, which indeed looks wrong today as there should not be anything special about the properties the last test creates vs those from any previous test.

RE: Adding multiple hooks of the same. Good idea! I'll add that.

test/main/modules.js
```diff
@@ -1,178 +1,483 @@
...
+      x.push('parent-beforeEach: ' + flat(this));
+      this.beforeEach = (this.beforeEach || '') + 'P';
+    });
+    hooks.afterEach(function () {
+      x.push('parent-afterEach: ' + flat(this));
+      this.afterEach = (this.afterEach || '') + 'P';
+    });
+    hooks.after(function () {
+      x.push('parent-after: ' + flat(this));
+      this.after = (this.after || '') + 'P';
```
raycohen, line 229: not sure this is a problem, but since we don't push onto x after the above line, nothing ever asserts that the `P` makes it onto the environment (except in the multiple-hooks case where the P2 shows up in the last after assertion)

Krinkle: I think this is a "tree falls in the forrest when no-one is around" type of situation. If there's no place to observe it within the API, there's not really a distinction between whether we do what we logically expect (pass the original `this` from before as-is) vs what would be "wrong" (pass a copy that we throw away later), since this is the last time the object is seen, after which we throw away the entire module.

If this isn't the last time, I stand corrected of course :)

PR #1787, opened 2024-07-25
Demos: Add Rollup and Webpack examples

Krinkle:
This is in preparation for adding an ESM distribution to QUnit. Before we do so, let's first observe how and what works today:

* import default: `import QUnit from 'qunit';`
* import named QUnit: `import { QUnit } from 'qunit';`
* import named methods: `import { module, test } from 'qunit';`
* require default: `require('qunit')`

In addition, verify that within a given bundler, a mixture of these across multple files, in indirect ways, always refers to the same object, both functionally (extensions to the object are visible to others), and by object identity (strict equality).

Specifically to satisfy Webpack, I've renamed the fixtures that use `require()` from `.js` to `.cjs` as [Webpack refuses](https://webpack.js.org/api/module-methods/) to compile `require()` calls otherwise in a directory with `{ type: 'module' }` in `package.json`. Rollup does not have this restriction.

Partially based on https://github.com/jquery/jquery/pull/5429 by @mgol.

Ref https://github.com/qunitjs/qunit/issues/1551.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /coverage/
 /demos/*/package-lock.json
 /demos/*/node_modules/
+/demos/*/tmp/
 /docs/.jekyll-cache/
 /docs/_site/
 /docs/Gemfile.lock
--- a/demos/bundlers/build.mjs
+++ b/demos/bundlers/build.mjs
@@ -1,0 +1,139 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import url from 'node:url';
+
+import { rollup } from 'rollup';
+import resolve from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+import webpack from 'webpack';
+
+const dirname = path.dirname(url.fileURLToPath(import.meta.url));
+const tmpDir = path.join(dirname, 'tmp');
+
+const inputs = [
+  `${dirname}/test/import-default.js`,
+  `${dirname}/test/import-named.js`,
+  `${dirname}/test/import-indirect.js`,
+  `${dirname}/test/require-default.cjs`,
+  `${dirname}/test/require-indirect.cjs`
+];
+
+const htmlTemplate = `<!DOCTYPE html>
+<html>
+<head>
+<title>{{title}}</title>
+<link rel="stylesheet" href="../node_modules/qunit/qunit/qunit.css">
+{{scriptTag}}
+</head>
+<body>
+<div id="qunit"></div>
+</body>
+</html>
+`;
+
+// Rollup configuration
+const rollupOutputs = [
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'es'
+  },
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'cjs'
+  },
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'iife'
+  },
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'umd',
+    name: 'UNUSED'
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -3,6 +3,8 @@
 /coverage/
 /demos/*/package-lock.json
 /demos/*/node_modules/
+/demos/*/tmp/
+/demos/bundlers/test-*.html
```
mgol, line 7: Couldn't this lie in `tmp` as well so that the previous line is enough?

demos/bundlers/build.mjs
```diff
@@ -0,0 +1,139 @@
...
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'iife'
+  },
+  {
+    dir: tmpDir,
+    entryFileNames: '[name].[format].js',
+    format: 'umd',
+    name: 'UNUSED'
```
mgol, line 55: What is this about?

demos/bundlers/build.mjs
```diff
@@ -0,0 +1,139 @@
...
+  // Clean up
+  fs.rmSync(tmpDir, { force: true, recursive: true });
+
+  const gRollup = buildRollup();
+  const gWebpack = buildWebpack();
+
+  for await (const fileName of gRollup) {
+    console.log('... built ' + fileName);
+
+    if (!fileName.endsWith('.cjs.js')) {
```
mgol, line 110: Can you explain this condition?

demos/bundlers/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "private": true,
+  "devDependencies": {
+    "@rollup/plugin-commonjs": "^26.0.1",
+    "@rollup/plugin-node-resolve": "^15.2.3",
+    "grunt": "1.6.1",
+    "grunt-contrib-connect": "^5.0.0",
+    "grunt-contrib-qunit": "10.1.1",
+    "qunit": "file:../..",
+    "rollup": "^4.18.0",
+    "webpack": "^5.92.0"
```
mgol, line 11: Is this a separate `package.json` so that these are not installed when developing QUnit at the top level?

mgol (approved):
I added a few questions but nothing blocking. An interesting approach with separating the bundlers test part.

demos/bundlers/build.mjs
Krinkle: The cjs outputs (not to confuse with cjs inputs) are not meant for browser context, as they depend on `module.exports` and would throw an uncaught error if you try to load it there. That reminds me, I wanted to add a Node.js test case as well. Right now this file is basically unused.

demos/bundlers/build.mjs
Krinkle: When reading a file as ESM, Rollup differentiates between a file that has no exports (i.e. app code), and a file that has exports. When reading a file as CJS, Rollup doesn't understand this, and so it insists on exporting. When using `umd`, one of the branches assigns a global to the overall `exports` which requires a name. Without the build will fail.

The exports is an empty object, but Rollup wants to put it somewhere, so gave it the name `window.` `UNUSED`.

.gitignore
Krinkle: Yep, I've done so already actually, but I forgot to remove the left-over line from gitignore. Fixed!

demos/bundlers/build.mjs
mgol: A comment in code could be useful here.

demos/bundlers/build.mjs
mgol: Can you add a comment?

demos/bundlers/build.mjs
Krinkle: Ack. Done in https://github.com/qunitjs/qunit/pull/1791.

PR #1794, opened 2024-08-07
CLI: Remove confusing `expected: undefined` from TAP for non-assert fail

Krinkle:
### pushFailure

This was setting `actual: null` and, left expected unset/undefined. In HtmlReporter this idom was recognised to mean "don't show values", based on hasOwn for `expected`. This worked because QUnit.log() passes the internal assert result object as-is.

In TAP output, this was not skipped because Test.js#logAssertion copies the object for the testEnd.errors array, and in doing so forges an `expected` property to exist no matter what, thus with an implied value of undefined. The hasOwn checks in TapReporter thus always evaluate to true.

This meant TAP output for pushFailure() calls not only showed redundant actual/expected entries, but actively created confusion by setting them to different values, drawing attention to a supposed difference that has no meaning

> actual: null
> expected: undefined

Fix by changing pushFailure to omit both actual and expected, and change the condition in both reporters to skip rendering of values based on both being strictly equal to `undefined`, instead of based on `hasOwn('expected')`.

Before:
```tap
TAP version 13
not ok 1 example
  ---
  message: Expected 2 assertions, but 1 were run
  severity: failed
  actual  : null
  expected: undefined
  stack: |
        at /test/example.test.js:1:7
```

After:
```tap
TAP version 13
not ok 1 example
  ---
  message: Expected 2 assertions, but 1 were run
  severity: failed
  stack: |
        at /test/example.test.js:1:7
```

### onUncaughtException / `QUnit.on('error')`

For uncaught errors, we already omitted both actual and undefined, the HtmlReporter thus already skipped them (for the reason above), but in TAP output they showed, for the same reason as above as:

> actual: undefined
> expected: undefined

Which, while not as actively confusing, is at least redundant. This is naturally fixed by the same change, which now omits this.

Before:

```tap
TAP version 13
Bail out! Error: boo example
  ---
  message: Error: boo example
  severity: failed
  actual  : undefined
  expected: undefined
  stack: |
        at /test.js:3:7
  ...
```

After:

```tap
TAP version 13
Bail out! Error: boo example
  ---
  message: Error: boo example
  severity: failed
  stack: |
        at /test.js:3:7
  ...
```

### DIFF
--- a/docs/api/assert/pushResult.md
+++ b/docs/api/assert/pushResult.md
@@ -18,8 +18,8 @@ Report the result of a custom assertion.
 | name | description |
 |------|-------------|
 | `data.result` (boolean) | Result of the assertion |
-| `data.actual` | Expression being tested |
-| `data.expected` | Known comparison value |
+| `data.actual` | Expression being tested (optional) |
+| `data.expected` | Known comparison value (optional) |
 | `data.message` (string or undefined) | Short description of the assertion |
 
 ## Examples
--- a/docs/resources/example-fail.html
+++ b/docs/resources/example-fail.html
@@ -14,7 +14,7 @@
   });
   QUnit.test('banana', function (assert) {
     assert.true(true, 'foo');
-    var sentence = 'This is actual.';
+    const sentence = 'This is actual.';
     assert.equal(sentence, 'This is expected.', 'example sentence');
     assert.true(true, 'bar');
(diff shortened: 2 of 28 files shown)


Krinkle:
Cross-ref https://github.com/qunitjs/qunit/pull/1789 (thematically related: polishing CLI/TAP output)

PR #1798, opened 2024-08-27
Build: Add native ESM distribution

Krinkle:
### Source

* Convert /src/core/qunit.js to ESM export.

  This file can be used directly by ESM in theory when debugging,   but in practice we'll let rollup bundle it into `qunit.module.js`, which is otherwise functionally the same as the source file.

* Add /src/core/qunit-commonjs.js as second entrypoint for rollup.

  Rollup would complain if attempting to export the ESM qunit.js source as "iife", as Rollup will detect the input file as having exports and insist on taking control over exporting it to a global variable. We don't want that as we control that ourselves instead.

  qunit-commonjs.js is a thin entrypoint that generates the `qunit.js` distribution file, which is the same as `qunit.module.js` except in "iife" format, with our (conditional) `module.exports` assignment at the end.

### Build

* `/qunit/qunit.js` is the same ES5/CJS compatible distribution it has always been, now sourced from /src/core/qunit-commonjs.js.

* `/qunit/qunit.module.js` is a complete standalone distribution in ESM format. Given that IE11 doesn't support type=module, I've removed IE11 from the Babel config for this file, which makes the file a bit smaller as it uses most ES6 features natively (275K vs 250K; uncompressed, unminified).

### Package

If we simply advertised the new `/qunit/qunit.module.js` file in `package.json#exports.module`, then we would break numerous Node.js and bundler use cases.

To test for these, previous commits already added `/demos/bunders/`, which is tested via `node bin/qunit.js demos/bundlers.js`, as part of `npm run test-demos` (and the "CI: Demos" GitHub job).

Specifically:

* A Node.js project that uses a mixture of CJS `require('qunit')` and ESM `import 'qunit'` previously shared a single logical QUnit instance since the module was just one file, and thus both used the same internal "require"-cache. This means the user's test files, or any test runners, plugins, etc, all import/require the same qunit module to define assertion plugins, add a hook listeners, or register tests. And thus everything executes correctly on the main QUnit instance.

  As-is this would break by having some of these attach stuff to a secondary instance that the main test runner (e.g. QUnit CLI) isn't running.

* A frontend project that uses a bundler or adapter (e.g. karma-qunit), and similarly involves mixed use among its test files and/or between it and any plugins, that sends the result to a browser. As-is this would end up embedding two separate copies of QUnit, unaware of each other.

To mitigate this, package.json contains two overrides:

* for `node`, tell ESM mode to import a Node.js-specific wrapper that re-uses and re-exports the CJS file.
* for `bundler`, tell CJS mode to import a bundler-specific wrapper that re-uses and re-exports the ESM file.

This is based on https://github.com/jquery/jquery/wiki/jQuery-4-exports-explainer as implemented at https://github.com/jquery/jquery/pull/5429, which does the same for jQuery.

Fixes https://github.com/qunitjs/qunit/issues/1551.
Fixes https://github.com/qunitjs/qunit/issues/1724.


mgol:
> A frontend project that uses a bundler or adapter (e.g. karma-qunit), and similarly involves mixed use among its test files and/or between it and any plugins, that sends the result to a browser. As-is this would end up embedding two separate copies of QUnit, unaware of each other.

This is a bit more convoluted. Bundlers, when the `module` condition is used without splitting it to `import` vs. `require` conditions, both `require` & `import` in source files will use the same file - bundlers can do it as they are preprocessors, they are not concerned with some operations being async vs. sync, so they are generally fine with `require('./esm-file.js')` & `import ... from 'commonjs-file.js'`.

However, the interop of `module.exports` treated as a module object vs. it exposing a single API and how it corresponds to a default export in ESM is problematic and having better control over it requires us to provide separate versions for `import` & `require`. As soon as we decide that, we need to use the wrapper approach to avoid code/data duplication.

package.json
```diff
@@ -37,6 +37,23 @@
...
+      "module": {
+        "import": "./qunit/esm/qunit.module.js",
+        "default": "./qunit/qunit-wrapper-bundler-require.js"
+      },
+      "import": "./qunit/esm/qunit.module.js",
+      "default": "./qunit/qunit.js"
+    },
+    "./qunit/qunit.css": {
+      "default": "./qunit/qunit.css"
+    }
```
mgol, line 55: If you have a single export, you can just provide the path to the entry point directly:
```suggestion
    "./qunit/qunit.css": "./qunit/qunit.css"
```

src/cli/require-qunit.js
```diff
@@ -1,3 +1,6 @@
+const path = require('path');
+const { Module } = require('module');
```
mgol, line 2: Any reason for not using strict mode in this & other build/CLI-related files?

src/core/qunit-wrapper-nodejs-module.js
```diff
@@ -0,0 +1,42 @@
+// In a single Node.js process, if different parts or dependencies
+// of a project mix ESM and CJS, avoid a split-brain state by making
+// sure both import and re-use the same instance via this wrapper.
+//
+// Node.js 18+ can import a CommonJS file from ESM.
```
mgol, line 5: Are you sure it's only 18+? From [12.x docs](https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_import_statements):
> An `import` statement can reference an ES module or a CommonJS module.

mgol (approved):
A few comments but nothing blocking

src/cli/require-qunit.js
Krinkle: No. I note that 3 of 5 files in `/src/cli/` have `'use strict';`. And `/bin` and `/build/tasks` have it too. I've fixed this separately in the main branch and rebased this PR. https://github.com/qunitjs/qunit/commit/5b7ab954c002cb9c47fcec095f486c8bf90e95e8

src/core/qunit-wrapper-nodejs-module.js
Krinkle: I only checked from 18 since that's our minimum, but fair enough, should have checked the docs for when it was introduced. Fixed.

matthewp:
This is amazing 🎇

PR #1801, opened 2024-09-29
CLI: Limit colors in TAP reporter to test names

Krinkle:
Don't apply colors to the "ok 1" and "not ok 1" prefix. This has the benefit of making the test names easier to visually extract for developers.

It also increases compatibility with the popular `tap-parser` package which doesn't parse TAP lines correctly otherwise, since blocks like "not ok" must start on a new line, and with the color code in front, they are technically either a continuation of the previous block, or an ignored "Anything" line between two blocks.

This hasn't been a problem in practice, since, if you pipe output from QUnit or TapReporter to a different process that uses tap-parser, colors are naturally disabled. But, if you get the console output from a browser process, the colors will be applied. I caught this while working on QTap (<https://github.com/qunitjs/qtap>).

https://github.com/tapjs/tapjs/tree/ccfea67fec/src/parser

https://testanything.org/tap-version-13-specification.html

| | Before | After
|--|--|--
| Pass |  | 
| Skip |  | 
| Failure |  |

### DIFF
--- a/src/core/reporters/TapReporter.js
+++ b/src/core/reporters/TapReporter.js
@@ -210,7 +210,7 @@ export default class TapReporter {
     if (!this.ended) {
       this.onRunStart();
       this.testCount = this.testCount + 1;
-      this.log(kleur.red(`not ok ${this.testCount} global failure`));
+      this.log(`not ok ${this.testCount} ${kleur.red('global failure')}`);
       this.logError(error);
     }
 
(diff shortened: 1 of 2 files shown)


src/core/reporters/TapReporter.js
```diff
@@ -247,9 +247,9 @@ export default class TapReporter {
 
     this.log(`1..${runEnd.testCounts.total}`);
     this.log(`# pass ${runEnd.testCounts.passed}`);
-    this.log(kleur.yellow(`# skip ${runEnd.testCounts.skipped}`));
-    this.log(kleur.cyan(`# todo ${runEnd.testCounts.todo}`));
-    this.log(kleur.red(`# fail ${runEnd.testCounts.failed}`));
+    this.log(`# ${kleur.yellow('skip ' + runEnd.testCounts.skipped)}`);
```
mgol, line 250: Why (here & below) use string concatenation instead of template literals?

mgol (approved):
One question / suggestion, LGTM overall. The arguments seem sensible and I have no problems with the visual changes.

src/core/reporters/TapReporter.js
Krinkle: I tend to default to concatenation still, applying template literals only when it makes the code more readable. 

Perhaps the difference is small enough that it doesn't matter, and that it is instead worthwhile to have only style in use. Personally, I haven't reached that point yet, especially when it's nested, I still favour concatenation:

```js
var str = 'foo bar ' + y.func( 'quux ' + z );

var str = `foo bar ${ y.func( `quux ${ z }` ) }`;
```

The concat example is immediately obvious to me as one string, followed by two more strings with a function applied. The template string example is notably more complex and ambigious until you read very carefully what starts and stops where, and what it embeds/calls, distinguishing between variables and string literals that look rather similar. Especially when producing output where special symbols are involved such as here in TAP, which of the braces and parenthesis are literal and which are JS template syntax? There's just a lot of clutter among them when it's nested and short.

I think the first form has less room for mistakes in writing, less ways to (mis)interpret it while reading, and quicker/easier to understand with confidence exactly what it does.

But, I'm definitely amplifying a small difference here for the purpose of answering your question "Why". I don't strongly oppose either style and would leave it as-is and unenforced based on what feels right.

Krinkle: Changed it. Mainly because I realize it was that way already, and keeping it that way minimises the diff :)

mgol (approved):

PR #1807, opened 2025-04-03
Docs: Tweak lifecycle diagram (add padding, remove open parenthesis)

Krinkle:
* remove unclosed parenthesis and comma

* add padding to improve dark mode

capture

Ref https://github.com/qunitjs/qunit/pull/1806.

\cc @FND


FND:
Looking good; that seems like a sensible solution - thank you!

PR #1815, opened 2025-11-05
Docs: Expand QUnit.config.testFilter docs

### LINKED ISSUE #1814: Add new config hook: `QUnit.config.testIf(() => {})`
I see that there exists `QUnit.test.if(name, condition, testCallback)`, but using this does not integrate well with flaky-test quarantine strategies.

For example, when running tests, we generate an ourput file (JUnit XML, or whatever), and then the quarantine system interprets that and does some analysis to determine if a test is flaky (over the course of multiple runs).

Next time we start a test suite, we have the list of known flaky tests ahead of time (as a json object available to us on page load).

I'd like to have something like:
```js
const myQuarantineList = []; // actually from file / api / whatever


QUnit.config.testIf((info) => {
  let { id, name } = info;

  let isQuarantined = myQuarantineList.find(x => {
    if (name === x) return true;

    return new RegExp(x).test(name);
  });

  return !isQuarantined;
});
```

gbudjeakp:
This PR implements `QUnit.config.testFilter`, a new configuration option that allows programmatic filtering of tests at runtime through a callback function.

**Context:**
Currently, `QUnit` provides filtering through CLI parameters (--filter, --module, etc.) and test-level methods `test.only()`, `test.skip()`, but there's no way to implement dynamic, programmatic filtering based on runtime conditions without modifying test code.

**This feature enables several important use cases:**

- Quarantining flaky tests in CI without modifying test code
- Parallel test execution by distributing tests across workers via hash-based sharding
- Runtime capability detection to skip tests for unavailable browser features (WebGL, IndexedDB, etc.)
- Dynamic test selection based on external data sources (APIs, configuration files)

Fixes https://github.com/qunitjs/qunit/issues/1814.

### DIFF
--- a/docs/api/config/testFilter.md
+++ b/docs/api/config/testFilter.md
@@ -32,6 +32,19 @@ The callback receives a `testInfo` object and must return a boolean:
 
 If the callback throws an error, the error is logged as a warning and the test is skipped.
 
+## Filter hierarchy
+
+The `testFilter` callback runs as part of a layered filtering system:
+
+1. **Test-level filters** run first: [`test.only()`](../QUnit/test.only.md), [`test.skip()`](../QUnit/test.skip.md), [`test.if()`](../QUnit/test.if.md)
+2. **Programmatic filter** runs next: `QUnit.config.testFilter`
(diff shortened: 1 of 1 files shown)


Krinkle (commented):
I'm landing the 2.x backport first, so the docs review is non-blocking. No rush :)

PR #1825, opened 2026-05-02
Assert: Return error value from rejects() and throws()

### LINKED ISSUE #1824: It would be convenient if `assert.rejects()` and `assert.throws()` returned the error value
It would be convenient if `assert.rejects()` and `assert.throws()` returned the error value.

When a more complex assertion is needed, I find the matcher callback to be a bit awkward, especially if the rest of the test code uses `await` and can mostly avoid nested callbacks. It would also make tests that check success and error cases more similar to each other.

Basically, I would like to be able to write tests like this:

```js
QUnit.test('getResultAsync (OK)', async (assert) => {
    let result = await getResultAsync('foo');
    assert.true(result instanceof MyResult)
    assert.strictEqual(result.value, 'FOO');
});

QUnit.test('getResultAsync (error, proposed API)', async (assert) => {
    let error = await assert.rejects(getResultAsync(''));
    assert.true(error instanceof MyError)
    assert.strictEqual(error.errorCode, 'empty-param');
});
```

Here's a runnable demo with some more examples, and the alternatives possible using the currently existing API.
[test.html](https://github.com/user-attachments/files/26979837/test.html)

This could be implemented by just adding `return actual;` at the end of the `throws` and `rejects` functions in [https://github.com/qunitjs/qunit/blob/main/src/core/assert.js](assert.js).

MatmaRex:
It permits a more natural way of asserting complex facts about the rejection value (thrown value) than the existing options of sticking assertions inside the `expectedMatcher` or writing your own try…catch blocks.

Fixes https://github.com/qunitjs/qunit/issues/1824.

### DIFF
--- a/docs/api/assert/rejects.md
+++ b/docs/api/assert/rejects.md
@@ -11,7 +11,8 @@ version_added: "2.5.0"
 ---
 
 `rejects( promise, message = "" )`<br>
-`rejects( promise, expectedMatcher, message = "" )`
+`rejects( promise, expectedMatcher, message = "" )`<br>
+`rejectionValue = await rejects( promise, message = "" )`
 
 Test if the provided promise rejects, and optionally compare the rejection value.
 
@@ -33,6 +34,10 @@ The `expectedMatcher` argument can be:
 * A RegExp that matches (or partially matches) `rejectionValue.toString()`.
 
 Note: in order to avoid confusion between the `message` and the `expectedMatcher`, the `expectedMatcher` **can not** be a string.
+
+`assert.rejects()` returns a promise which is resolved with the rejection value from the provided
+promise. This can be used instead of `expectedMatcher` to assert other facts about the expected
+failure.
 
 ## See also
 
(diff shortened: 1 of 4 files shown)


MatmaRex:
I saw those errors locally too:
```

/home/runner/work/qunit/qunit/src/core/dump.js
  225:23  error  Map is not supported in iOS Safari 7.0-7.1, IE 9  compat/compat

/home/runner/work/qunit/qunit/src/core/utilities.js
  26:22  error  Array.includes() is not supported in iOS Safari 7.0-7.1, IE 9  compat/compat
```

They don't seem to be my fault.

Krinkle:
Two CI issues:

1. A warning about `array.includes` from eslint-plugin-compat that was not previously found. This is techically correct, but not appplicable to us because we already check it beforehand. I'll fix this in the main branch with an ignore rule.
2. A warning about `Map` from an eslint-plugin-compat indirect update. This is incorrect and an upstream regression. I've reported this at https://github.com/amilajack/eslint-plugin-compat/issues/688, and will likewise temporarily work around it for now.

Krinkle (approved):

PR #1831, opened 2026-05-07
Core: Fix typo for td in qunit-assert-list

edg2s:


Krinkle:
Did you notice a bug or other rendering difference in the HTML diff after this change?

It seems that the data cell here is always as tall or taller than the heading cell to the left of it (which is always a one-word label). If true, that means there's no difference between `baseline` (default) and `top`. From a basic example, there is no difference between the rule on or off:

Screenshot 2026-05-07 at 23 49 38

Anyway, I'll land this since there is it a clear improvement to the code.

edg2s:
No, just the typo. Maybe it can be removed.

PR #1834, opened 2026-05-30
Dump: add Map and Set support to QUnit.dump.parse

### LINKED ISSUE #1812: QUnit.dump.parse returns an empty object for Maps
Version: 2.20.1
Test browser: Chrome 141.0.7390.55

Test cases:

```
> QUnit.dump.parse([1,3,4])
'[\n  1,\n  3,\n  4\n]'
```
^^ looks reasonable.

```
> QUnit.dump.parse({foo: "bar", baz: 2})
'{\n  "baz": 2,\n  "foo": "bar"\n}'
```
^^ likewise.

```
> QUnit.dump.parse(new Map([["a","b"], ["c","d"]]))
'{}'
```
^^ feels like a bug.

xxiaoxiong:
Fixes #1812

## Problem

`QUnit.dump.parse(new Map([["a","b"], ["c","d"]]))` currently returns `{}` instead of showing the Map contents.

Root cause: `typeOf()` doesn't recognize `Map`/`Set` objects, so they fall through to `typeof obj === "object"`. The `object` parser uses `for...in` which doesn't enumerate Map/Set entries.

## Fix

1. Added `Map` and `Set` detection in `typeOf()`
2. Added `map` parser that formats as `Map{key: value, ...}`
3. Added `set` parser that formats as `Set{val1, val2, ...}`

### DIFF
--- a/docs/api/extension/QUnit.equiv.md
+++ b/docs/api/extension/QUnit.equiv.md
@@ -1,0 +1,102 @@
+---
+layout: page-api
+title: QUnit.equiv()
+excerpt: A recursive comparison of JavaScript values.
+groups:
+  - extension
+redirect_from:
+  - "/extension/QUnit.equiv/"
+version_added: "1.0.0"
+---
+
+`QUnit.equiv( a, b, ... )`
+
+Compare two or more values for equality in a recursive, deep manner.
+
+| name | description |
+|------|-------------|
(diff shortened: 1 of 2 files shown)


xxiaoxiong:
I am unable to sign the CLA, so closing this PR. Thank you for the review.

PR #1836, opened 2026-08-24
Add viewport meta tag to browser testing example HTML

takluyver:
To enable viewing at native size on mobile browsers. This was added to the docs before (https://github.com/qunitjs/qunitjs.com/pull/110 ), but seems to have got lost in a subsequent change.

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -23,6 +23,7 @@ QUnit releases are standalone and require no runtime dependencies for use in the
 <!DOCTYPE html>
 <html>
 <meta charset="utf-8">
+<meta name="viewport" content="width=device-width">
 <title>QUnit</title>
 <link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.26.0.css">
 <body>


lbesecker195 (approved):
Restores the viewport meta tag that was previously in the docs example so the browser runner renders at native scale on mobile. LGTM.

Krinkle:
The linked PR (https://github.com/qunitjs/qunitjs.com/pull/110) added the meta viewport tag to `resources/example-index.html`, which is a demo embedded via iframe at https://qunitjs.com/browser/. The code of this iframe is not shown to users and is not an instruction. The purpose of this demo is to interactively show what an QUnit test page looks like. I don't remember why I later [removed](https://github.com/qunitjs/qunitjs.com/commit/30f316fdf0c57722b8393de70ae094ca639e015e#diff-f4de910ef3e94d0f1e4ed5be8de0dcdfc2025c80f0003d157b6c0e04f850d6ed) that, but I suspect I did so because the iframe demo has a fixed size.

That is something different from what this PR is doing, which is changing the code block on https://qunitjs.com/browser/, this code is shown to users and is an instruction to new users of QUnit, telling them how to create their own test page.

